PR #14, opened 2015-10-05
Add Template config property

reintroducing:
Small change to add the ability to define a template path in the config. I tested this locally and it worked so please do the same and let me know if any changes are necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
+.DS_Store
+node_modules
 example/styleguide/
 .publish/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,2 @@
-.DS_Store
-node_modules
 example/styleguide/
 .publish/
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-styleguidist",
   "description": "React components styleguide",
-  "version": "0.1.8",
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -1,2 +1,4 @@
+.DS_Store
+node_modules
```
sapegin: Things like `.DS_Store` (anything created by your OS, IDE, etc.) should be in your global .gitignore.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "react-styleguidist",
   "description": "React components styleguide",
-  "version": "0.1.8",
+  "version": "0.1.9",
```
sapegin: Please don’t change version number and changelog (at least the version).

sapegin:
Thanks! The feature itself is fine ;-)

.gitignore
reintroducing: Good thought, I removed it and set up a global (not sure why I never did it before).

package.json
reintroducing: No problem, I was wondering how you want to handle this, removing it.

reintroducing:
@sapegin Updated with the requested changes.

sapegin:
:+1:

PR #15, opened 2015-10-06
Add CodeMirror syntax theme support

reintroducing:
This is my first time using Webpack so not sure if this is the best way to do it but it does work.  Let me know if any changes need to be made.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -121,6 +121,12 @@ Type: `Number`, default: `3000`
 
 Dev server port.
 
+### syntaxTheme
+
+Type: `String`, default: `base16-light`
+
+The name of the [CodeMirror theme](http://codemirror.net/demo/theme.html) to use for syntax highlighting in examples.
+
 ### getExampleFilename
 
 Type: `Function`, default: finds `Readme.md` in the component folder
--- a/Readme.md
+++ b/Readme.md
@@ -121,7 +121,7 @@ Type: `Number`, default: `3000`
 
 Dev server port.
 
-### syntaxTheme
+### highlightTheme
 
 Type: `String`, default: `base16-light`
 
--- a/loaders/styleguide.loader.js
+++ b/loaders/styleguide.loader.js
@@ -43,6 +43,7 @@ module.exports.pitch = function() {
 	return [
 		'module.exports = {',
 		'	title: ' + JSON.stringify(config.title) + ',',
+		'	syntaxTheme: ' + JSON.stringify(config.syntaxTheme) + ',',
 		'	components: [' + components.join(',') + ']',
 		'};'
 	].join('\n');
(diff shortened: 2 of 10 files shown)


src/components/Editor/Editor.js
```diff
@@ -31,6 +34,16 @@ export default class Editor extends Component {
 		this._handleChange = debounce(this.handleChange.bind(this), UPDATE_DELAY);
 	}
 
+	componentWillMount() {
```
sapegin: I don’t like this part. Some thoughts: 

* Move default theme to the default Styleguide config.
* Do not mutate Editor.codemirrorOptions, extend it with theme name on render().
* Require CSS only if it was specified in this.props: `if (this.props.syntaxTheme) { cssRequire(…); }`.
* Don’t a need for _.isUndefined here.

Readme.md
```diff
@@ -121,6 +121,12 @@ Type: `Number`, default: `3000`
 
 Dev server port.
 
+### syntaxTheme
```
sapegin: For me “syntax” is more about the language rules than about the colors so maybe something like `highlightTheme` would be better.

sapegin:
Awesome, thanks!

loaders/styleguide.loader.js
```diff
@@ -43,6 +43,7 @@ module.exports.pitch = function() {
 	return [
 		'module.exports = {',
 		'	title: ' + JSON.stringify(config.title) + ',',
+		'	syntaxTheme: ' + JSON.stringify(config.syntaxTheme) + ',',
```
sapegin: Probably it’s better to export all needed options as a single object because we most probably will need to export more than two options.

reintroducing: Not 100% sure what you mean by this.

reintroducing:
@sapegin updated based on your feedback. Let me know if anything else is necessary.

loaders/styleguide.loader.js
sapegin: `config: JSON.stringify( some config fields )`.

But it’s not urgent.

src/components/Editor/Editor.js
```diff
@@ -39,9 +46,14 @@ export default class Editor extends Component {
 	}
 
 	render() {
+		let { highlightTheme } = this.props;
+		let options = Editor.codemirrorOptions;
+
+		options.theme = highlightTheme;
```
sapegin: Something like `let options = _.merge({}, Editor.codemirrorOptions, {theme: highlightTheme})` would be better because it won’t mutate default settings object.

reintroducing:
@sapegin Updated the Editor component with suggested changes.

sapegin:
Great, thanks!

PR #25, opened 2015-11-05
Added support for using require(...) in code examples.

mik01aj:
@sapegin, I need your help with testing this code. And I wait for any other comments. ;)

### DIFF
--- a/loaders/examples.loader.js
+++ b/loaders/examples.loader.js
@@ -4,7 +4,7 @@ var marked = require('marked');
 var evalPlaceholder = '<%{#eval#}%>';
 var codePlaceholder = '<%{#code#}%>';
 
-var requireAnythingRegex = /require\s*\((.+)\)/g;
+var requireAnythingRegex = /require\s*\(([^)]+)\)/g;
 var simpleStringRegex = /^"([^"]+)"$|^'([^']+)'$/;
 
 function readExamples(markdown) {
--- a/loaders/examples.loader.js
+++ b/loaders/examples.loader.js
@@ -1,12 +1,13 @@
-var fs = require('fs');
-var path = require('path');
+var _ = require('lodash');
 var marked = require('marked');
-var config = require('../src/utils/config');
 
-var defaultRenderer = marked.Renderer.prototype;
+var evalPlaceholder = '<%{#eval#}%>';
+var codePlaceholder = '<%{#code#}%>';
+
+var requireAnythingRegex = /require\s*\((.+)\)/g;
+var simpleStringRegex = /^"([^"]+)"$|^'([^']+)'$/;
 
 function readExamples(markdown) {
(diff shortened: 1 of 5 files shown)


mik01aj:
I thought more about it, and it should be actually possible to infer the right context from the example code. And for sure this would also be the cleanest way to do it.

sapegin:
That would be very nice because now it’s hard to understand.

mik01aj:
Well, I researched this a bit, but apparently webpack doesn't expose its internal behaviour of creating a require contect containing whatever the module needs. [The context API](https://webpack.github.io/docs/context.html) lets me create a context giving a regex, so I could look for `/require\s*\((.*)\)/` in the example code and make a regex that would match any of the found strings and nothing else, but this seems clunky. It will work, and I can provide some friendly error handling so it won't bite the users much. Maybe it's still better than adding the hardly understandable config option.

What do you think?

sapegin:
I think it’s much better ;-)

mik01aj:
Okay, see the new commit.

I'd add some tests for my helper functions but I don't see any test framework installed. Could you take care of this?

sapegin:
Cool! I’ll add Mocha soon.

mik01aj:
Check my new commit `facceef`! :)

sapegin:
Mocha is here!

mik01aj:
I thought about it again and I realized that I don't need the webpack's `context` at all. The `context` had also one drawback: requires of stuff from `../` didn't work. My new solution is simpler and cleaner, and requiring anything just works.

I still didn't add any tests, but I don't know what exactly could I test. Please look at the code and tell me if you feel like some part of it might break in the future.

sapegin:
Cool, thanks!

About tests:

1. Complex regexps could be tested. It probably means extracting extracting them to utils.
2. I thinks we can test that loader returns valid JavaScript (try to compile it with Babel?).

mik01aj:
> extracting them to utils

This is harder than I thought because the loader is in ES5 and utils are ES6.

sapegin:
Or you can just export these function from the loader and use them in tests.

mik01aj:
That's what I do:

```
module.exports = _.assign(exampleLoader, {
	requireAnythingRegex: requireAnythingRegex,
	simpleStringRegex: simpleStringRegex,
	readExamples: readExamples,
	findRequires: findRequires,
});
```

Btw, there's a simpler way to check JS (ES5) validity: http://www.ecma-international.org/ecma-262/5.1/#sec-15.3.2.1

sapegin:
Cool! And a simpler way is always better ;-)

mik01aj:
Done!

sapegin:
Cooooool!

mik01aj:
commit updated, ihmo it's good to merge now.

sapegin:
Thank you very much!

:shipit:

PR #26, opened 2015-11-05
Fix for #13

### LINKED ISSUE #13: Exception when having an enum in propTypes
(originally reported in https://github.com/sapegin/react-styleguidist/issues/9#issuecomment-145516722)

I can now run styleguidist, and it can display the props for one component, but when I add another one, it fails:

image

I think it's because of the enum in the other component's `propTypes`. Here are the `propTypes` of the relevant component:

```
    propTypes: {
        issue: React.PropTypes.object,
        width: React.PropTypes.oneOf([26, 40, 55, 70, 110, 155, 190, 265]),
        style: React.PropTypes.object,
    },
```

mik01aj:
This should fix #13.

### DIFF
--- a/src/components/Props/Props.js
+++ b/src/components/Props/Props.js
@@ -58,6 +58,9 @@ export default class Props extends Component {
 	}
 
 	renderEnum(prop) {
+		if (!Array.isArray(prop.type.value)) {
+			return <span>{prop.type.value}</span>;
+		}
 		let values = prop.type.value.map(({ value }) => (
 			<li className={s.listItem} key={value}>
 				<code>{this.unquote(value)}</code>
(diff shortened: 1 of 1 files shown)


src/components/Props/Props.js
```diff
@@ -58,6 +58,9 @@ export default class Props extends Component {
 	}
 
 	renderEnum(prop) {
+		if (prop.type.value.constructor !== Array) {
```
sapegin: I think Array.isArray is better than checking a constructor.

sapegin:
Cool, thanks!

sapegin:
Could you please fix this and I’ll merge it?

mik01aj:
should be ok now.

PR #29, opened 2015-11-10
Added "defined in file ..." information for the components.

mik01aj:
Looks like this:

image

The shown path is relative to `config.rootDir`.

### DIFF
--- a/loaders/styleguide.loader.js
+++ b/loaders/styleguide.loader.js
@@ -10,6 +10,7 @@ function processComponent(filepath) {
 	var hasExamples = !!fs.existsSync(examplesFile);
 	return '{' + [
 			'filepath: ' + JSON.stringify(filepath),
+			'relativePath: ' + JSON.stringify(path.relative(config.rootDir, filepath)),
 			'module: ' + requireIt(filepath),
 			'props: ' + requireIt('!!props!' + filepath),
 			'examples: ' + (hasExamples ? requireIt('examples!' + examplesFile) : null)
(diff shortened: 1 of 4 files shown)


sapegin:
Nice! Do we really need this “defined in”? I think just path is enough.

mik01aj:
Maybe not. I can remove it.

mik01aj:
Btw. I did some other style changes: bold headers and bigger margins between components.

sapegin:
Bigger margins: +1.
Bold headers: –1 ;-)

It’d be nice to make styleguide themable though.

mik01aj:
So, how do you like the updated commit?

mik01aj:
So, how do you like the updated commit?

sapegin:
Almost ;-) Why do you need margin-top instead of margin-bottom?

mik01aj:
to put some space before the first component. It doesn't do anything for subsequent components because of margin collapsing.

sapegin:
But it’s the same as margin-bottom on the header.

sapegin:
```
<header>
  <h2/>
  <div>description</div>
</header>
```

```
.header {
  margin-bottom: 1rem;
}
```

mik01aj:
I'm not sure we're talking about the same thing. Here's what I did:

image

sapegin:
Yes, about the same thing. The problem is: you introduced margin-top, which, I believe should (and can) be avoided.

mik01aj:
Got it! :) Check the new commit :)

sapegin:
Cool, thanks!

PR #60, opened 2015-12-09
Make css modules class names unique

sunesimonsen:
I have a problem where I use a css module style called `container` in my components, but that conflicts with the styleguide styles. Add a hash to make the class names unique.

### DIFF
--- a/src/make-webpack-config.js
+++ b/src/make-webpack-config.js
@@ -10,7 +10,7 @@ var config = require('../src/utils/config');
 
 module.exports = function(env) {
 	var isProd = env === 'production';
-	var cssLoader = 'css?modules&importLoaders=1&localIdentName=ReactStyleguidist-[name]__[local]!postcss';
+	var cssLoader = 'css?modules&importLoaders=1&localIdentName=ReactStyleguidist-[name]__[local]---[hash:base64:5]!postcss';
 
 	var codeMirrorPath = getInstalledPath('codemirror', true);
 	var reactTransformPath = getInstalledPath('babel-plugin-react-transform', true);


mik01aj:
These names were added because of https://github.com/sapegin/react-styleguidist/issues/32#issuecomment-157875674.

And I don't understand where you're having conflicts, because all the class names are already prefixeds with `ReactStyleguidist-`.

sunesimonsen:
Because my components uses css modules and are loaded by the postcss loader configured in react-styleguides

sunesimonsen:
This line adds the config.rootDir to the includes of the postcss loader: https://github.com/sapegin/react-styleguidist/blob/master/src/make-webpack-config.js#L183

Here you can see that the rootDir is parts of the includes: https://github.com/sapegin/react-styleguidist/blob/master/src/make-webpack-config.js#L18

Am I misunderstanding something here, or should `rootDir` point to the directory where your components live?

sunesimonsen:
@mik01aj I see, it is actually a feature that you can override the styles, so they should not be unique. But then I would like the postcss loader to keep its hands of my components and I can add my own loader.

I might just confused about how the `rootDir` should be used, because right now it prefixes my components with `ReactStyleguidist-`.

Sorry for noise if I'm just wrong.

mik01aj:
Ah, so *that's* the bug. Your components shouldn't be prefixed. I suppose it's a problem with the webpack config.

sunesimonsen:
@mik01aj I basically want to control the css loading for my own components.

sunesimonsen:
I'll close this PR and make an issue for that instead.

sapegin:
CSS loader was enabled for your components folder by mistake :-) I’ll fix that in the next release.

sunesimonsen:
Thanks that sounds great.

sunesimonsen:
@sapegin I'll just make an issue for that, so it isn't forgotten.

sunesimonsen:
@sapegin https://github.com/sapegin/react-styleguidist/issues/62

PR #66, opened 2015-12-16
Issue 59 fix

mik01aj:
#59

PR #67, opened 2015-12-16
Layout fix

mik01aj:
Layout was broken when there was some wide component. This little change fixes this.


sapegin:
Look like you’ve committed it to the same branch as #66.

mik01aj:
Not really. It was a branch based on that one and I thought the other commits will disappear when you merge the first one.
image

mik01aj:
But now I rebased it and there should be just one.

PR #78, opened 2016-01-28
Support flow annotations for prop types

phpnode:
Support for flow type annotations is coming to react-docgen via https://github.com/reactjs/react-docgen/pull/48 (note: PR is not yet merged but works locally and looks ready)

With this PR, the styleguide will use the flow annotation if it is present, and fall back to the `propType` if it isn't.

The syntax looks like this:

```js
type Props = {
  /* The title of the box */
  title: string;
}

class Box extends React.Component {
  props: Props;
  render (): React.Element {
    return <h1>{this.props.title}</h1>;
  }
}
```

### DIFF
--- a/src/rsg-components/Props/Props.js
+++ b/src/rsg-components/Props/Props.js
@@ -25,7 +25,7 @@ export default class Props extends Component {
 			rows.push(
 				<tr key={name}>
 					<td className={s.cell}><Code className={s.name}>{name}</Code></td>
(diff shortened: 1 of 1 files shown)


sapegin:
Looks like it’s safe to merge now, isn’t it?

phpnode:
@sapegin yep, it's safe

sapegin:
Thanks!

danez:
But be aware that the returned types differ between propTypes and flow due to the different options one has in flow.
see https://github.com/researchgate/react-docgen/blob/add-flow-type-support/README.md#types

PR #81, opened 2016-02-07
Feature/babel6 simple

oliverturner:
A reduced change set rebased off master


src/make-webpack-config.js
```diff
@@ -12,10 +12,9 @@ function getPackagePath(packageName) {
 
 module.exports = function(env) {
 	var isProd = env === 'production';
-	var cssLoader = 'css?modules&importLoaders=1&localIdentName=ReactStyleguidist-[name]__[local]!postcss';
+	var cssLoader = 'css?module&importLoaders=1&localIdentName=ReactStyleguidist-[name]__[local]!postcss';
```
sapegin: It should be [`modules`](https://github.com/webpack/css-loader#css-modules).

src/rsg-components/Preview/Preview.js
```diff
@@ -37,7 +37,10 @@ export default class Preview extends Component {
 	}
 
 	compileCode(code) {
-		return babel.transform(code, {stage: 0}).code;
+		return transform(code, {
+			presets: ['es2015', 'react', 'stage-0'],
+			ignore: [/node_modules/]
```
sapegin: Do we need to ignore `node_modules` here? We transpile only the code inside the textarea.

sapegin:
That’s great, thank you very much!

The only thing I’d like to change is Lodash imports: I prefer `import func from 'lodash/func'` style.

.nvmrc
```diff
@@ -0,0 +1 @@
+v5
```
sapegin: Is it required by Babel 6 or is it good practice or something very useful? ;-)

src/make-webpack-config.js
oliverturner: My bad... #willfix

src/rsg-components/Preview/Preview.js
oliverturner: Force of habit :)

.nvmrc
oliverturner: Not required by Babel 6, but incredibly useful to people who need to switch between versions of Node: on some projects I need to use the Node LTS version (4.2.x) to stay in sync with an environment, on others I want to be switch to v5. 

Using https://github.com/wbyoung/avn, simply by cd'ing into the directory I get switched to the project-specific version. I can certainly remove, but it's a nice feature for some.

oliverturner:
Re the preference for lodash import syntax, I'd suggest that it's more in line with the direction that lodash itself is going to reference the module directly?

sapegin:
Modules, not packages. And now it’s very simple (in 3.x we had to write `lodash/groupname/func`, now it’s just `lodash/func`. I thinks it’s easier to maintain (you have to install just one `lodash` package) and only the required modules will be bundled by Webpack (so it’s as efficient as using multiple packages).

oliverturner:
Removed all references to lodash.x packages and replaced `_.pluck` with `_.map` per discussion at https://github.com/underdash/underdash/issues/4#issuecomment-110026962

PR #87, opened 2016-02-21
Custom react components

oliverturner:
PoC of an easier way for users to customise the layout of the styleguide: each component provides a stateless "renderer" that can be switched out as required using the `alias` hooks in the webpack section of `styleguide.config.js`.

Currently implemented for `ReactComponent` and `Layout` only: can be extended as desired.


example/lib/styleguide/components/ReactComponent/ReactComponent.css
```diff
@@ -0,0 +1,87 @@
+.root {
+  composes: font from "../../styles/common.css";
```
sapegin: Spaces? ;-) Do yo use EditorConfig?

oliverturner: oops...

sapegin:
Looks like we need another example to show this customizations. Otherwise people will this that all this files you’ve added to example are required.

I like the implementation and most of my comment below are about code style.

src/rsg-components/Layout/Layout.js
```diff
@@ -28,28 +26,21 @@ export default class Layout extends Component {
 		}
 	}
 
+	renderToC(components) {
+		return <TableOfContents components={components}/>;
+	}
+
 	render() {
-		let { config, components } = this.props;
+		const { config, components } = this.props;
```
sapegin: I think pull request is not a good place to change project’s code style. If you think that `const` is better than `var` we can discuss it and maybe change for the whole project.

example/styleguide.config.js
```diff
@@ -1,4 +1,4 @@
-var path = require('path');
+const path = require('path');
```
sapegin: This file should be ES6 because Webpack loaders require it. And as far as I know Node 0.12 don’t support let/const.

src/rsg-components/Layout/Layout.css
```diff
@@ -1,53 +1,53 @@
 .root {
-	composes: base base-bg from "../../styles/colors.css";
+  composes: base base-bg from "../../styles/colors.css";
```
sapegin: Ooops.

src/rsg-components/Layout/Layout.js
```diff
@@ -1,18 +1,16 @@
 import { Component, PropTypes } from 'react';
-import Sticky from 'react-sticky';
 import Components from 'rsg-components/Components';
 import TableOfContents from 'rsg-components/TableOfContents';
 
 import s from './Layout.css';
 
-export default class Layout extends Component {
+const Layout = (Renderer) => class extends Component {
```
sapegin: Why do you move export to the end of file?

src/rsg-components/Layout/Layout.js
```diff
@@ -28,28 +26,21 @@ export default class Layout extends Component {
 		}
 	}
 
+	renderToC(components) {
```
sapegin: renderToC — renderTableOfContents?

src/rsg-components/Layout/Layout.js
oliverturner: > I think pull request is not a good place to change project’s code style.

Agreed. One of those force-of-habit things, not intending to change style without consensus.

> If you think that const is better than var we can discuss it and maybe change for the whole project.

Personally I do prefer it, and indeed only ever use `let` to indicate that a variable can be redeclared.

example/styleguide.config.js
oliverturner: `const` seems to be supported in Node 0.12: see http://stackoverflow.com/questions/28388885/ecmascript-6-features-available-in-node-js-0-12#answer-28389178

src/rsg-components/Layout/Layout.css
oliverturner: :stuck_out_tongue:

next commit changes it back...

example/styleguide.config.js
sapegin: Cool ;-)

src/rsg-components/Layout/Layout.js
sapegin: We can change it later. I don’t have strong opinion about `let` vs `const` but `const` seems very popular style now.

oliverturner:
> Looks like we need another example to show this customizations. Otherwise people will this that all this files you’ve added to example are required.

Good point. What about moving the current `/example/lib/styleguide/`directory to `example/styleguide-customised` to indicate that it's not meant to be a part of the app source or eve necessary?

sapegin:
Yep, move all your example to a new folder and revert basic example from master. And probably move them to `examples/basic` and `examples/customised` folders.

sapegin:
I also think that it’d be good to have a complex example in a separate repository. Something more realistic: maybe an existing project like React Bootstrap.

oliverturner:
Well that would at least be the ultimate test of how effective this tool is: if it can work against a library that size, including getting the UX right - performance, making components easy to find, etc. - then it can work on almost any codebase

sapegin:
Yep!

oliverturner:
Really not sure what the conflict would be: merges fine against master for me :(

sapegin:
Most probably some of my latest changes. I’ll merge it manually after 2.0 release.

oliverturner:
Yeah... think I fell out of sync. Apologies if I messed anything up!

sapegin:
You can update the docs and squash everything ;-)

oliverturner:
Hm... Probably should apply the same pattern (splitting out stateless renderers) to other components, otherwise it'll confuse users.

sapegin:
Would be cool!

oliverturner:
Gonna have to ask for your help squashing these commits :(

sapegin:
I’ll merge it manually later ;-)

oliverturner:
Thanks :muscle:

sapegin:
Squashed and merged: 0ff2b2f93dffe30046d8c20c819bf9aeb28b209e.

sapegin:
Any plans to finish the feature (add renderers to other components, update docs)?

oliverturner:
Absolutely: sorry for the delay, been prepping for my first tech talk (yesterday: went ok; need practice!)

oliverturner:
Thanks for wrangling the rat's nest of commits!

sapegin:
Cool! What was the talk? Slides, video? :-)

oliverturner:
It was a primer for how tooling works: how package managers, task runners, transpilers & module bundlers work together

oliverturner:
For new devs it looks impenetrable!

sapegin:
That’s true, and not only for new sometimes ;-)

oliverturner:
I'll link to slide & video presently 😀

mik01aj:
Can't wait! :)

PR #89, opened 2016-02-22
Updated the README.

mik01aj:
Most important change is the more streamlined installation instruction. I also changed formatting in a few places, added a few badges, links, and stuff.

See it with formatting: https://github.com/sapegin/react-styleguidist/tree/readme-update

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -397,4 +397,4 @@ Everyone is welcome to contribute. Please take a moment to review the [contribut
 
 [Artem Sapegin](http://sapegin.me) and [contributors](https://github.com/sapegin/react-styleguidist/graphs/contributors).
 
-MIT Licenses, see the included [License.md](License.md) file.
+MIT License, see the included [License.md](License.md) file.
--- a/Readme.md
+++ b/Readme.md
@@ -1,8 +1,8 @@
 # React Styleguidist
 
-[![Build Status](https://travis-ci.org/sapegin/react-styleguidist.svg)](https://travis-ci.org/sapegin/react-styleguidist) [![Dependency Status](https://david-dm.org/sapegin/react-styleguidist.svg)](https://david-dm.org/sapegin/react-styleguidist) [![npm](https://img.shields.io/npm/v/react-styleguidist.svg)](https://www.npmjs.com/package/react-styleguidist) [![npm](https://img.shields.io/npm/dm/react-styleguidist.svg)]()
+[![Build Status](https://travis-ci.org/sapegin/react-styleguidist.svg)](https://travis-ci.org/sapegin/react-styleguidist) [![Dependency Status](https://david-dm.org/sapegin/react-styleguidist.svg)](https://david-dm.org/sapegin/react-styleguidist) [![npm](https://img.shields.io/npm/v/react-styleguidist.svg)](https://www.npmjs.com/package/react-styleguidist)
 
-React Styleguidist is a React components style guide generator. It lists component `propTypes` and shows live, editable usage examples based on Markdown files. You can use it as both in a hot reloaded (style guide) dev server during development and a static site generator (which makes a great deliverable). Check out [**the demo styleguide**](http://sapegin.github.io/react-styleguidist/).
+React Styleguidist is a React components style guide generator. It lists component `propTypes` and shows live, editable usage examples based on Markdown files. You can use it to generate a static HTML page to share and publish or as a workbench for developing new components using hot reloaded dev server. Check out [**the demo styleguide**](http://sapegin.github.io/react-styleguidist/).
 
 Based on Webpack, webpack-dev-server and Babel.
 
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -1,124 +1,127 @@
 # React Styleguidist
 
-[![Build Status](https://travis-ci.org/sapegin/react-styleguidist.svg)](https://travis-ci.org/sapegin/react-styleguidist)
+[![Build Status](https://travis-ci.org/sapegin/react-styleguidist.svg)](https://travis-ci.org/sapegin/react-styleguidist) [![Dependency Status](https://david-dm.org/sapegin/react-styleguidist.svg)](https://david-dm.org/sapegin/react-styleguidist) [![npm](https://img.shields.io/npm/v/react-styleguidist.svg)](https://www.npmjs.com/package/react-styleguidist) [![npm](https://img.shields.io/npm/dm/react-styleguidist.svg)]()
```
sapegin: Not sure about this badges:

1. I regularly update dependencies. Do people really care about that or looks at this badge?
2. 1K is quite small number. Nothing to be proud of yet ;-)

Readme.md
```diff
@@ -1,124 +1,127 @@
...
 
-React components style guide generator with a hot reloaded (style guide) dev server.
+React Styleguidist is a React components style guide generator. It lists component `propTypes` and shows live, editable usage examples based on Markdown files. You can use it as both in a hot reloaded (style guide) dev server during development and a static site generator (which makes a great deliverable). Check out [**the demo styleguide**](http://sapegin.github.io/react-styleguidist/).
```
sapegin: > You can use it as both in a hot reloaded (style guide) dev server during development and a static site generator (which makes a great deliverable). 

Maybe: You can use it as a static site generator to publish your style guide or as a workbench for developing new components using hot reloaded dev server.

Readme.md
```diff
@@ -1,124 +1,127 @@
...
-    return webpackConfig;
-  }
-};
-```
+2. Add a **`styleguide.config.js`** file into your project’s root folder. A simplest possible config looks like this:
 
-**Note**: don’t forget `include` option for your Webpack loaders, otherwise they will interfere with Styleguidist’s loaders.
+   ```javascript
+   module.exports = {
+     title: 'My Great Style Guide',
```
sapegin: `title` is also optional :-)

Readme.md
mik01aj: I think dependency status is useful, it's an indicator that the project is maintained.

Readme.md
```diff
@@ -1,124 +1,127 @@
...
+     components: './lib/components/**/*.js',
+     // Put other configuration options here...
+   };
+  ```
 
-When you run dev-server `NODE_ENV` is set to `development` so if you use [React Transform](https://github.com/gaearon/react-transform-hmr) hot module replacement it will be enabled for your components. When you build style guide `NODE_ENV` is set to `production`.
+  All the configuration options are documented in the [Configuration section](#configuration) below. But you don't need all of them to start. Just read on.
```
sapegin: > But you don't need all of them to start. Just read on.

Can be safely removed ;-)

Readme.md
```diff
@@ -1,124 +1,127 @@
...
-      ]
-    }
-  }
-}
-```
+   Put the `updateWebpackConfig` function in your config:
 
-See the *Configuration* section below for the list of available options.
+   ```javascript
+   updateWebpackConfig: function(webpackConfig, env) {
```
sapegin: It’s better to have `module.exports = {` and all that so people have no doubts where to put this code.

Readme.md
mik01aj: I think that it helps to keep focused. I could reword it to:

> (you might need them later)

sapegin: I don’t think so, there are just words that have no value but that the user has to read ;-)

Readme.md
mik01aj: I'd leave it :)

Readme.md
mik01aj: I can change this, but... I wanted to also mention that you can easily make a package with the styleguide and email it to your manager, or client, and it just works.

Readme.md
mik01aj: that's why I added a comment above...

Readme.md
```diff
@@ -1,124 +1,127 @@
...
-Add these scripts to your `package.json`:
+   **Note**: don’t forget `include` option for your Webpack loaders, as otherwise they will interfere with Styleguidist’s loaders.
 
-```json
-{
-  // ...
-  "scripts": {
-    "styleguide-server": "styleguidist server",
-    "styleguide-build": "styleguidist build"
+4. Configure [React Transform](https://github.com/gaearon/react-transform-hmr) HMR (hot module replacement). This is optional, but highly recommended.
```
sapegin: I don’t like this section. I think many people already have it but it’s really hard to understand that if you have it it should just works. And we don’t have to use the HRM abbreviation, “hot module replacement” is enough ;-)

I’d merge  it as is and try to improve later.

Readme.md
```diff
@@ -257,32 +255,21 @@ You can change some settings in the `styleguide.config.js` file in your project
...
-  getExampleFilename: function(componentpath) {
-    return componentpath.replace(/\.js$/, '.examples.md');
-  },
-};
-```
 
 ## CLI commands and options
 
-These commands supposed to be placed in `package.json` `scripts` (see Installation section above). If you want to run them from command line do it like this: `./node_modules/.bin/styleguidist`.
+These commands supposed to be placed in `package.json` `scripts` (see Installation section above). If you want to run them directly, use `./node_modules/.bin/styleguidist` or `$(npm bin)/styleguidist`.
```
sapegin: We’re not teaching npm here, on way to run binary is enough ;-)

Readme.md
sapegin: I don’t understand what “deliverable” means anyway ;-)

Readme.md
mik01aj: It's like leading the user by hand: "let's go here... see this... and that... and now we continue to there...". It's putting these actions in context.

Readme.md
mik01aj: ok

Readme.md
sapegin: That’s my opinion: if a phrase or sentence doesn’t add anything to the text it can be removed.

The list already creates a context: a ordered set of operations. Then we provide links or notes with additional information like a link to all available options.

I think it’d better to move this paragraph to the end of list because at this time user has no idea which options he’d need and we describe the most important later anyway.

Readme.md
mik01aj: hm, that's a good point. How about:

> You can use it to generate a static HTML page to share and publish or as a workbench for developing new components using hot reloaded dev server.

Readme.md
mik01aj: okay, agreed! I'll move it to the end.

Readme.md
mik01aj: On the other hand, it's an implicit way to say "well, actually babel is required in your project". That's why I say it's optional.

sapegin: I’ll think about that ;-)

mik01aj: btw: HMR appears later too and also the config contains "hmre". I think it makes sense to explain this and keep using the acronym.

sapegin:
If we can avoid acronyms we should ;-)

oliverturner:
Any thoughts on making the introductory README slimmer? As in, a brief overview and quick set-up, but break out customisation, advanced config, how things work behind the scenes, etc. into linked wiki pages.

Really like the changes here, but we're still heading into wall-of-text territory. This is understandable given the flexibility of the app, but it may be more appealing to new users if it looks easy to get started. Just a thought in advance of offering to help out...

sapegin:
@oliverturner I like the idea but I don’t like wiki ;-) Maybe `docs` folder with markdown files would be better.

mik01aj:
I think the FAQ could be a separate file. Everything else is imho fine in the README.

oliverturner:
@mik01aj agreed: not suggesting that we strip the README of useful top-level stuff, only that we distribute some of the weight.

@sapegin not fixated on the wiki: any pattern - e.g. docs + markdown - works for me :)

PR #91, opened 2016-03-02
Handle unknown proptypes

### LINKED ISSUE #77: Missing propTypes for getDefaultProps results in tricky exception
I was getting
```
TypeError: Cannot read property 'name' of undefined
Props.renderType
/.../node_modules/react-styleguidist/src/rsg-components/Props/Props.js

```

And could not figure out why. Apparently reason was that one of values that was returned from `getDefaultProps` was not defined in `propTypes` in one of my components.

Maybe just some warning instead of 'unknown' exception. Alternatively exception with better explanation would also be good.

Very cool tool btw :+1:  .

karlbright:
Fixes #77.

image

This ensures that any prop that is assigned via `getDefaultProps` but missing from `propTypes` will be shown as an unknown prop type.

Although this is bad practice in React, I think it's probably best to display the style guide rather than throw and error for this, and ensure that the user can find out the explanation behind unknown proptypes in the styleguide readme.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -382,6 +382,23 @@ export default class Wrapper extends Component {
 2. Press the ![Debugger](http://wow.sapegin.me/image/2n2z0b0l320m/debugger.png) button in your browser’s developer tools.
 3. Press the ![Continue](http://wow.sapegin.me/image/2d2z1Y2o1z1m/continue.png) button and the debugger will stop exception a the next exception.
 
+### Why does the style guide list one of my prop types as unknown?
+
+This occurs when you are assigning props via getDefaultProps that are not listed within the components propTypes.
+
+For example, the color prop here is assigned via getDefaultProps but missing from the propTypes, therefore the styleguide is unable to display the correct propType.
+
+```javascript
+Button.propTypes = {
+	children: PropTypes.string.isRequired,
+	size: PropTypes.oneOf(['small', 'normal', 'large']),
+};
+
+Button.defaultProps = {
+	color: '#333',
+	size: 'normal'
+};
+```
 ## Similar projects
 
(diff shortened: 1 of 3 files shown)


mik01aj:
LGTM. I like the idea of displaying `unknown`.

karlbright:
@mik01aj Good call. I was meant to go back and fix those up and completely forgot :+1:

src/rsg-components/Props/Props.js
```diff
@@ -11,6 +11,8 @@ export let Code = ({ className = '', children }) => {
 	return <code className={sMarkdown.code + ' ' + className}>{children}</code>;
 };
 
+export let UnknownPropType = () => <span>unknown</span>;
```
sapegin: Do we need to export it?

Readme.md
```diff
@@ -382,6 +382,23 @@ export default class Wrapper extends Component {
 2. Press the ![Debugger](http://wow.sapegin.me/image/2n2z0b0l320m/debugger.png) button in your browser’s developer tools.
 3. Press the ![Continue](http://wow.sapegin.me/image/2d2z1Y2o1z1m/continue.png) button and the debugger will stop exception a the next exception.
 
+### Why does the style guide list one of my prop types as unknown?
```
sapegin: `unknown`? ;-)

sapegin:
Just two comments from me, otherwise I really like this. Thanks @karlbright!

src/rsg-components/Props/Props.js
karlbright: It is actually no longer required, have removed this now.

karlbright:
All sorted.

Issue #68, opened 2015-12-29
We need to change our flexbox layout to something else (poor rendering performance)


mik01aj:
The single reason is: performance.

![image](https://cloud.githubusercontent.com/assets/3027415/12033557/a292869e-ae24-11e5-9f97-e746cec59805.png)

In this recorded timeline I'm playing with a simple progress bar component. And it seems that anything that *might* become bigger than page width would cause a full page layout recalculation and re-render (I mean browser rerender, not react). (Rendering is purple in these graphs)

Attaching the nav always via `position: fixed` and having a `margin-right` on `Layout` should fix this.

sapegin:
Could you try it please?

mik01aj:
![image](https://cloud.githubusercontent.com/assets/3027415/12034011/1d32beae-ae2b-11e5-9199-4d9e7770604f.png)

Disabling this single rule makes animations smooth as silk.

mik01aj:
Closed in #92.

PR #92, opened 2016-03-02
Ensure components list is scrollable

karlbright:
**This PR makes some layout and style changes to the default style guide output**

styleguidist-scrolling

My existing codebase has a large number of components, running styleguidist on this results in a list of components that is not scrollable do the "sticky" functionality that currently exists.

This PR makes changes to the layout to ensure that the component list is always scrollable. The component list has been moved to the left of the screen due to the fact that there is a max-width on the styleguide(which i think is important) and implementing scrolling functionality with that given layout felt odd. It wasn't clear that you could scroll the components list, this approach goes for something a little more "traditional" as far as documentation styling goes.

image

### DIFF
--- a/examples/customised/styleguide/components/Layout/Layout.css
+++ b/examples/customised/styleguide/components/Layout/Layout.css
@@ -19,12 +19,6 @@
 	padding-left: 15px;
 }
 
-.sidebarIsSticky {
-	position: fixed;
-	top: 15px;
-	padding-left: 15px;
-}
-
 .heading {
 	composes: reset font from "../../styles/common.css";
 	margin: 0 0 30px 0;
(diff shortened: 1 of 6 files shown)


mik01aj:
Does it also fix #68?

karlbright:
@mik01aj Damn, i was JUST about to add that comment :)

mik01aj:
LGTM, I like the simplification.

sapegin:
Cool, thanks! The only problem is now it’s shifted to the left side of the window which looks terrible on a big screen:

sapegin:
How about [something like this](http://styleguide.cfapps.io/react_components_autocomplete.html)?



Sidebar is on the left but the main area is centered. I also like the style guide title inside the sidebar.

karlbright:
I'm all for the title in the sidebar :+1:

karlbright:
image

sapegin:
Nice!

PR #102, opened 2016-03-10
Explicitly specify when a prop is required

CompuIves:
Maybe this is just a nitpick from my side, but I think it would be better to explicitly specify it when a prop is required. This way a developer does not overlook your prop requirement that easy.

I'm curious what you think about this change :smile:.

### DIFF
--- a/src/rsg-components/Props/Props.css
+++ b/src/rsg-components/Props/Props.css
@@ -44,9 +44,6 @@
 	padding-left: 15px;
 }
 
-.optional {
-	composes: light from "../../styles/colors.css";
-}
 .required {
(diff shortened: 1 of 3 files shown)


CompuIves:
Sure!
with explicit
versus
without explicit

sapegin:
Maybe?

CompuIves:
Yes, this is more clear. I'll update the pull request!

CompuIves:
Updated:
screenshot-localhost 3000 2016-03-11 15-51-57

sapegin:
Cool, thanks!

PR #104, opened 2016-03-14
added support for import statements in code examples

ctoscano:
Added support for import statements so that code examples and component definitions can keep the same code style.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -126,6 +126,11 @@ const mockData = require('./mocks');
 <Message content={mockData.hello}/>
 ```
 
+Or you can `import` external files into your examples:
+
+  import {female} from "dog-names";
+  <RandomButton variants={female}/>
+
 As an utility, also the [lodash](https://lodash.com/) library is available globally as `_`.
 
 Each example has its own state that you can access at the `state` variable and change with the `setState` function. Default state is `{}`.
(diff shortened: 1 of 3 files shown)


sapegin:
Very cool, thanks!

Could you please update the readme too: https://github.com/sapegin/react-styleguidist#writing-code-examples

mik01aj:
Waaait... I'm not sure we actually need to match `import` statements. Babel should transpile `import` to `require`.

ctoscano:
@sapegin: I added the example to Readme.md#writing-code-exmaples. 

@mik01aj: While I agree with you, that would cost a broader re-write and re-evaluating an unknown number of previously made design decisions. If you know the code well enough to point out where the change should be made I can take a closer look, if you think it is worth it.

mik01aj:
Ah, I see the loader doesn't transpile the code, it's transpiled in runtime. So this would mean adding additional transpilation step to the loader, only to parse the requires. So maybe you're right, it's just practical to handle imports this way. But, could you please add unit tests to make sure that it works reasonably well for these cases:

```
import { square, diag } from 'lib';
import * as lib from 'lib';
import myFunc from 'myFunc';
import MyClass from 'MyClass';
import _, { each } from 'underscore';
import { default as foo } from 'lib';
import foo from 'lib';
import * as lib from 'lib';
import { inc, counter } from 'lib';
import theDefault, { named1, named2 } from 'src/mylib';
import theDefault from 'src/mylib';
import { named1, named2 } from 'src/mylib';
import { named1 as myNamed1, named2 } from 'src/mylib';
import * as mylib from 'src/mylib';
import 'src/mylib';
```

(Source: http://www.2ality.com/2014/09/es6-modules-final.html)

sapegin:
+1 for unit tests.

sapegin:
Closing it for now. Feel free to reopen with tests.

PR #108, opened 2016-03-28
Support for creating Sections

paulj:
This PR introduces the ability to split components into sections. Sections are configured in the `styleguide.config.js` file as a peer to the main `components` property. Sections can be nested if desired, since they are processed recursively. Sections can also include a reference to a `content` examples file that is used an introduction to the section.

Support for non-sectioned components is retained, so this PR should have no effect on existing guides that don't configure sections.

### DIFF
--- a/loaders/styleguide.loader.js
+++ b/loaders/styleguide.loader.js
@@ -31,16 +31,15 @@ function requireIt(filepath) {
 	return 'require(' + JSON.stringify(filepath) + ')';
 }
 
-module.exports = function() {};
-module.exports.pitch = function() {
-	this.cacheable && this.cacheable();
+function processComponentsSource(components, config) {
+	if (!components) return null;
 
 	var componentSources;
-	if (typeof config.components === 'function') {
-		componentSources = config.components();
+	if (typeof components === 'function') {
+		componentSources = components();
 	}
 	else {
-		componentSources = glob.sync(path.resolve(config.configDir, config.components));
+		componentSources = glob.sync(path.resolve(config.configDir, components));
 	}
 
 	if (config.verbose) {
@@ -54,7 +53,29 @@ module.exports.pitch = function() {
 		componentSources = _.filter(componentSources, hasExamples);
 	}
 
-	var components = componentSources.map(processComponent);
+	return '[' + componentSources.map(processComponent).join(',') + ']';
+}
+
+function processSection(section, config) {
+	return '{' + [
+		'name: ' + JSON.stringify(section.name),
+		'content: ' + (section.content ? requireIt('examples!' + section.content) : null),
+		'components: ' + processComponentsSource(section.components, config),
+		'sections: ' + processSectionsList(section.sections, config)
+	].join(',') + '}'
+}
+
+function processSectionsList(sections, config) {
+	if (!sections) return null;
+
+	return '[' + 
+		sections.map(function(section) { return processSection(section, config); }).join(',') +
+	']';
+}
+
+module.exports = function() {};
+module.exports.pitch = function() {
+	this.cacheable && this.cacheable();
 
 	var simplifiedConfig = _.pick(config, [
 		'title',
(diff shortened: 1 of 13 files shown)


loaders/styleguide.loader.js
```diff
@@ -54,7 +53,29 @@ module.exports.pitch = function() {
 		componentSources = _.filter(componentSources, hasExamples);
 	}
 
-	var components = componentSources.map(processComponent);
+	return '[' + componentSources.map(processComponent).join(',') + ']';
```
sapegin: JSON.stringify?

src/rsg-components/Components/Components.js
```diff
@@ -1,20 +1,26 @@
 import { Component, PropTypes } from 'react';
 import ReactComponent from 'rsg-components/ReactComponent';
 import Renderer from 'rsg-components/ReactComponent/Renderer';
+import Section from 'rsg-components/Section';
+import SRenderer from 'rsg-components/Section/Renderer';
```
sapegin: SectionRenderer?

src/rsg-components/Layout/Layout.js
```diff
@@ -4,16 +4,19 @@ import TableOfContents from 'rsg-components/TableOfContents';
 
 import s from './Layout.css';
 
+import _ from 'lodash';
```
sapegin: Please use this form of imports for Lodash: `import isEmpty from 'lodash/isEmpty';`.

sapegin:
That’s very cool! Thank you very much. Just a few questions:

1. Can we also use HTML files as a `content`? 

2. Is it renders everything on a single page? No routing?

src/rsg-components/Components/Components.js
paulj: Calling it `SectionRenderer` led to a name duplication later when I mixed the renderer and section together. I thought for consistency, it was better to rename this on import.

paulj:
No worries. Do you have any thoughts on how you're wire in using HTML files for content? I'm just forwarding them straight into the examples loader, so we're inheriting whatever capabilities it has there.

Everything is on a single page for now. I didn't want to make this PR too huge!

src/rsg-components/Components/Components.js
sapegin: I think `Section/Renderer` should be used only inside `Section/Section`, like in `Components/Components` for example.

sapegin:
> No worries. Do you have any thoughts on how you're wire in using HTML files for content? I'm just forwarding them straight into the examples loader, so we're inheriting whatever capabilities it has there.

We could figure it out later. But it could work as is for most of HTML files I think.

> I didn't want to make this PR too huge!

Sure ;-)

src/rsg-components/Components/Components.js
paulj: So you're thinking that as a peer to `<Components>`, there is a `<Sections>`? I'd assumed they were fairly closely tied, and to try to keep compatible with non-sectioned HTML, I was needing to make `<Components>` add in child sections. Because the import `Renderer` was already taken by the component renderer, I was left a bit stuck for names,

sapegin: I think it’s make code more consistent: you call a component by it’s name and it creates a Renderer inside which user may override in their style guide config.

loaders/styleguide.loader.js
paulj: That would break the inline `require` statements, and presumably end up quoting the JSON object returned by `processComponent`?

sapegin: Yeah, you’re right.

src/rsg-components/Components/Components.js
paulj: Ok. I've added a `Sections` component that mirrors the style of the `Components` component.

paulj:
I've merged master into this branch to remove the conflicts. Was there anything else you can see that needs changing at the moment?

sapegin:
Thank you very much!

PR #109, opened 2016-03-28
Example Doclets

paulj:
This PR introduces the ability to associate additional example files using doclet tags in the description.

For example, a component defined like the below will have the examples in `extra.examples.md` associated with it:

  ```js
  /**
   * Component is described here.
   *
   * @example ./extra.examples.md
   */
  export default class SomeComponent extends React.Component {
    // ...
  }

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -115,6 +115,22 @@ Styleguidist generates documentation from 2 sources:
   Any [Markdown](http://daringfireball.net/projects/markdown/) is **allowed** _here_.
   ```
 
+* **Loading examples using doclet tags**
+
+  Additional example files can be specifically associated with components using doclet (`@example`) syntax.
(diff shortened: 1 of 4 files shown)


sapegin:
Thanks!

Can we reuse someting from [this file](https://github.com/reactjs/react-docgen/blob/44777a9040b7bde9498ee8cf0a527a4c747a6955/src/utils/docblock.js)?

paulj:
@sapegin we actually do use the functionality from that file. The doclets are parsed with `props.doclets = reactDocs.utils.docblock.getDoclets(props.description);`, which ultimately lands in that file. I had to reproduce the regexp though, since that file doesn't provide any mechanism for stripping out the doclets from the description once they are parsed.

sapegin:
Sorry, didn’t notice it.

Issue #111, opened 2016-03-30
Union elements share same key


marklundin:
React is raising the `Encountered two children with the same key, '.$instanceOf'` warning when attempting to render a component with multiple `PropType.instanceOf`s.  It seems that the component keys aren't properly assigned.

Here's a test component that causes the error 

```
import { PropTypes } from 'react'

let EmptyComponent = _ => <div>empty</div>

EmptyComponent.propTypes = {
    someValue:PropTypes.oneOfType([
        PropTypes.instanceOf( Int8Array ),
        PropTypes.instanceOf( Uint8Array )
    ]).isRequired
}

export default EmptyComponent
```
```
warning.js?7f53:45 Warning: flattenChildren(...): Encountered two children with the same key, `.$instanceOf`. Child keys must be unique; when two children share a key, only the first child will be used.
```

I've had a look through the code and think it could be [this](https://github.com/sapegin/react-styleguidist/blob/master/src/rsg-components/Props/Props.js#L120) but can't be sure.

marklundin:
Further info on this. Whilst the above throws the warning, the modified code below does not.

```
import { PropTypes } from 'react'

let EmptyComponent = _ => <div>empty</div>

let oneOfType = [
    PropTypes.instanceOf( Int8Array ),
    PropTypes.instanceOf( Uint8Array )
]

EmptyComponent.propTypes = {
    someValue:PropTypes.oneOfType( oneOfType )
}

export default EmptyComponent
```

sapegin:
Yeah, looks like we shouldn’t use name alone as a key. Could you please submit a pull request with a fix?

marklundin:
opened a PR

PR #112, opened 2016-03-30
Update Props.js

marklundin:
`renderUnion` method now uses the array index to create a component key. Fixes issue #111

### DIFF
--- a/src/rsg-components/Props/Props.js
+++ b/src/rsg-components/Props/Props.js
@@ -116,8 +116,8 @@ export default class Props extends Component {
 		if (!Array.isArray(getType(prop).value)) {
 			return <span>{getType(prop).value}</span>;
 		}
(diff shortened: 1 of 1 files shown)


sapegin:
Using indexes as keys are not good thing but there’s no other choice here.

mik01aj:
A unit test would be appreciated.

sapegin:
Yeah, we should write more tests ;-)

mik01aj:
Well, it's like this in every project. I mean that anyone else can see `key={value.name + i}` and think "wtf, name+number? I will simplify it". And we can easily have a test case for this specific scenario. Or at least a comment in the code.

sapegin:
Better both. But I agree at least comment is necessary.

marklundin:
Feel free to comment

Issue #37, opened 2015-11-14
No suitable component definition found


panayi:
![image](https://cloud.githubusercontent.com/assets/707005/11164476/43f71f3e-8afa-11e5-80b7-fd164fd0c439.png)

I know this is not an issue specifically with react-styleguidist, for example: https://github.com/reactjs/react-docgen/issues/29

This happens in a lot of cases: when extending/composing components, or when using Redux and the [ `connect()` wrapper](http://redux.js.org/docs/basics/UsageWithReact.html). For example, the following component is not recognised:

``` javascript
class Theme extends Component {
  // ...
  render() {
    // ...
  }
}

function mapStateToProps(state) {
  const variables = selectVariables(state.theme.name)
  return {
    variables: selectVariables(state.theme.name),
    styles: selectStyles(state.theme.name, variables)
  }
}

function mapDispatchToProps(dispatch) {
  return bindActionCreators(ThemeActions, dispatch)
}

export default connect(mapStateToProps, mapDispatchToProps)(Theme)
```

When using wrappers, we could use two files: `_Theme.jsx` (export "raw" component and use it for styleguide), and `Theme` (that uses wrappers etc.). However it would be great if it just work.

Any ideas on what needs to be done?

sapegin:
> Any ideas on what needs to be done?

That’s what I wanted to ask you :-| No ideas right now.

mik01aj:
Maybe the [update of react-docgen](https://github.com/sapegin/react-styleguidist/pull/46) will fix this.

If not, I believe this bug should be reported to react-docgen, as it is thrown from https://github.com/reactjs/react-docgen/blob/c94b7689484be5fb5b573d24d4d7c04833d03fd8/src/parse.js

panayi:
I have worked a bit on this, and react-docgen will search for a "suitable" component definition in any of the exports. So,

Before:
```javascript
// this throws the above error
module.exports = someWrappedComponent
```

After:
```javascript
// react-docgen will find a component in module.exports.otherKey and use that.
module.exports = {
  default: someWrappedComponent,
  otherKey: suitableComponent
}
```

However when doing the latter `react-styleguidist` fails since it expects that a single default component is exported, see https://github.com/sapegin/react-styleguidist/blob/master/loaders/styleguide.loader.js#L26. 

I think `react-styleguidist` should use whatever component `react-docgen` resolves to.

sapegin:
[...]

PR #114, opened 2016-03-31
added support for named export

pvasek:
This is attempt to fix issue #37. I chose the simple method described in the issue. Sorry, could not figure out the where I should put the resolver. It seems there is a lot of magic going on :) 

I could not find any unit tests for this code. Could you point me out where I can find them?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -298,6 +298,11 @@ You can change some settings in the `styleguide.config.js` file in your projectâ
   };
   ```
 
+* **`namedExport`**<br>
+  Type: `bool`, optional<br>
+  If set the named export is used instead of the default export for resolving components in examples. 
+  The name of the file is used as name of the export. 
+
 * **`resolver`**<br>
   Type: `Function`, optional<br>
   Function that allows you to override the mechanism used to identify classes/components to analyze. Default
--- a/loaders/styleguide.loader.js
+++ b/loaders/styleguide.loader.js
@@ -5,6 +5,10 @@ var prettyjson = require('prettyjson');
 var _ = require('lodash');
 var config = require('../src/utils/config');
 
+function getNamedExport(filepath) {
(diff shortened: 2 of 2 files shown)


mik01aj:
The test for examples loader is in [`examples.loader.spec.js`](https://github.com/sapegin/react-styleguidist/blob/master/test/examples.loader.spec.js).

I'd say :-1: for adding a new config option. Ideally it should "just work" without any config (I think it's easy to detect that the exported component is not the default export).

I also wonder how does this relate to https://github.com/sapegin/react-styleguidist/pull/106 (I haven't tested that one). I mean, if you don't want to `export default` something, then it's probably because you have something more in this file, so isn't this case already served by that pull request?

sapegin:
I agree that a new option is not good.

pvasek:
OK, no problem, I will try a different approach.

I think #106 solved the first part of the problem, to get props information for these components.

Could you point me out where is a good place to do that work. I am still quite confused how it works :) Thanks

mik01aj:
See this PR: https://github.com/sapegin/react-styleguidist/pull/104. If you could write unit tests for it, you could use the ES6 `import` in your example code. Do you need anything more?

pvasek:
OK, I will check it. Thank you!

pvasek:
Turned outs, it was actually related to #104. Because in that case I can use named export and default export together. I just didn't realize that it hasn't been released yet. Sorry to bother you and thank you once again.

sapegin:
Does it mean that we can close #37 and this PR?

pvasek:
I have already closed this PR. PR #106 solves completely if you have the same component as named and default as well. 

The #37 is probably more connected to the fact that the example component is always the default exported one. In that case the #104 could probably be a solution or at least workaround.

PR #119, opened 2016-04-11
Ensure lack of sections does not break styleguide and update examples

karlbright:
This fixes issues introduced by the merge of #108 in to master.

Both @paulj and @sapegin may have some ideas on the changes to `index.js` here, but this ensures that sections not being provided will not break the styleguide. There were issues with running examples as a result of this merge.

This also adds a basic example using sections for testing this new feature.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,7 +33,15 @@ function processSections(ss) {
 	});
 }
 
-components = processComponents(components);
-sections = processSections(sections);
+// Components are required unless sections are provided
+if (sections) {
+	sections = processSections(sections)
+	components = []
+} else {
+	components = processComponents(components)
+	sections = []
+}
+components = components ? processComponents(components) : [];
+sections = sections ? processSections(sections) : [];
 
 ReactDOM.render(<StyleGuide config={config} components={components} sections={sections} />, document.getElementById('app'));
--- a/src/index.js
+++ b/src/index.js
@@ -35,13 +35,12 @@ function processSections(ss) {
 
 // Components are required unless sections are provided
 if (sections) {
-	sections = processSections(sections)
-	components = []
-} else {
-	components = processComponents(components)
-	sections = []
+	sections = processSections(sections);
+	components = [];
 }
-components = components ? processComponents(components) : [];
-sections = sections ? processSections(sections) : [];
+else {
+	components = processComponents(components);
+	sections = [];
+}
 
 ReactDOM.render(<StyleGuide config={config} components={components} sections={sections} />, document.getElementById('app'));
--- a/src/index.js
+++ b/src/index.js
@@ -33,14 +33,7 @@ function processSections(ss) {
 	});
 }
 
-// Components are required unless sections are provided
-if (sections) {
-	sections = processSections(sections);
-	components = [];
-}
-else {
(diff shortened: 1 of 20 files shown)


src/index.js
```diff
@@ -33,7 +33,15 @@ function processSections(ss) {
...
+// Components are required unless sections are provided
+if (sections) {
+	sections = processSections(sections)
+	components = []
+} else {
+	components = processComponents(components)
+	sections = []
+}
+components = components ? processComponents(components) : [];
+sections = sections ? processSections(sections) : [];
```
sapegin: So it will run twice?

Maybe:

```javascript
if (sections) {
	sections = processSections(sections);
	components = components ? processComponents(components) : [];
 }
else {
 	components = processComponents(components);
 	sections = sections ? processSections(sections) : [];
 }
```

karlbright: Whoops. I completely forgot to take line 44 and 45 out :)

sapegin:
Please also check Travis log.

karlbright:
@sapegin The other option here is to move these to the `process*` functions, which means that this base code should remain quite simple?

src/index.js
```diff
@@ -33,7 +33,14 @@ function processSections(ss) {
...
-sections = processSections(sections);
+// Components are required unless sections are provided
+if (sections) {
+	sections = processSections(sections);
+	components = [];
+}
+else {
+	components = processComponents(components);
+	sections = [];
+}
```
paulj: A trivial fix to this would be just:

    components = processComponents(components);
    sections = processSections(sections || []);

The intention was that you could use sections and components in parallel (if for some reason you wanted that).

sapegin:
But how would you check for required sections or components? (Not sure it’s necessary.)

And in you current implementation you can use only sections or components, not both. (No sure it’s critical or bad.)

src/index.js
sapegin: That’s the best version I think.

karlbright:
@sapegin Yes, It wasn't clear whether or not they could both be supplied or if components were ignored when sections was provided, so I expected you or @paulj would have some information on this front.

src/index.js
karlbright: It wasn't clear whether they could both be provided or not, if they can then this works fine. I went for the more verbose ifelse here just to make it clear what was happening. I'd be more than happy to clean it up like this though :+1:

karlbright:
So to be clear here, my understanding is this:

- *Components* is not required, if the *Sections* property of the config is provided
- If no *Sections* property is provided, then *Components* is expected/required to exist
- Both *Components* and *Sections* can be provided and the styleguide takes these both in to account without issue

That last point is something I am not clear on. I assume that they can both be provided, however if a section contains a component that is provided within the components configuration property, then that component will be listed twice within the style guide. Is this correct @sapegin and @paulj?

paulj:
@karlbright Yes. If the same files ended up being listed twice (either in components and sections, or even just in two different sections), then they would end up being included twice.

karlbright:
@paulj Cool, figured as much. Updated to cleaner syntax, much happier.

sapegin:
Thanks @karlbright @paulj!

PR #120, opened 2016-04-11
Add basic styling for section headings

karlbright:
Takes some styling from the existing ReactComponent headings and ensures that sections do not use Times New Roman and look out of place.

image

### DIFF
--- a/src/rsg-components/Section/Section.css
+++ b/src/rsg-components/Section/Section.css
@@ -4,13 +4,18 @@
 }
 
 .header {
-
+	composes: font from "../../styles/common.css";
+	margin-bottom: 20px;
+	font-weight: bold;
 }
 
 .heading {
-
+	position: relative;
+	margin: 0 0 6px 0;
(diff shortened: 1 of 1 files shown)


src/rsg-components/Section/Section.css
```diff
@@ -4,13 +4,18 @@
...
 .header {
-
+	composes: font from "../../styles/common.css";
+	margin-bottom: 20px;
+	font-weight: bold;
 }
 
 .heading {
-
+	position: relative;
```
sapegin: Do we really need position:relative here?

karlbright: Probably not, this was taken directly from `ReactComponent` heading styling.

sapegin: That’s because of `<a className={s.anchor} href={'#' + name}></a>` but it doesn’t work now because of overflow:auto in `Layout` component. But I think we can remove them because now we have a table of contents with links.

Issue #98, opened 2016-03-03
Add routing to Styleguidist


zammer:
For the project I'm working on we implemented simple routing (not including react-router because we didn't need the more advanced functionality). 

We have a homescreen then each component has it's own page.
Is this something you may interested in?
If so would you want to use react-router?
Also should there be an option to have a one page or multi page app? (Not thought how to implement this)

sapegin:
Not sure about one page per component but routing is definitely a good thing to have because we want to add components grouping (by tag or function or whatever) and ability to add random Markdown and HTML documents to style guide.

Also check this issues: #31, #12.

sapegin:
There is very basic routing implementation and ability to open a component in a separate page now. Feel free to open a separate issue or submit a pull request if you want something more specific.

mik01aj:
@sapegin we have filtering (#123), but can we actually have an URL that would jump straight to the single component view? Maybe I'm missing something.

sapegin:
@mik01aj I’ve added links yesterday but the UI is not the best. But better than nothing because I always forget this `#!/Button`.

sapegin:
![](http://wow.sapegin.me/1e3t2b3U0u2v/Screen%20Recording%202016-06-24%20at%2010.19%20AM.gif)

eduardoinnorway:
What is the status here? I would like to have one page per group of components ...

pnepe6:
We would like to have a site next to our documentation, I've been trying to implement react-router-redux but I can't get wih this configuration another path served than `/`. Every created route return 404. any idea

PR #123, opened 2016-04-12
Single Component View

karlbright:
**This is part of an ongoing project to look at speeding up webpack, as well as speeding up the development within styleguide driven development**

This has been mentioned in #98 and is something my team has struggled with as we look to use this library for styleguide driven development.

Currently, when using the styleguide and hot reload functionality to develop new components, or modify existing components, there is an issue with scroll position due to loading of code previews and rendered components between page reloads.

Although this could be solved with some extra functionality around page load and ensuring the page is scrolled to the correct position, our team decided that having a simple solution for viewing a single component on the page would be more beneficial. This PR includes the work responsible for implementing this functionality.

As noted, this is a WIP Pull Request and some of the implementation could be handled a little nicer. But I made a conscious effort to avoid something like react-router as it would definitely feel like overkill for such a simple task as this.

I've pushed this out for my team to use and will be looking at feedback from them to see if this helps, but i'm keen to get feedback from others here to see if this fixes up any issues you have with the existing styleguide.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "babel-runtime": "6.5.0",
     "babel-standalone": "6.4.4",
     "chalk": "1.1.1",
-    "classnames": "2.2.3",
+    "classnames": "^2.2.3",
     "codemirror": "5.11.0",
     "css-loader": "0.23.1",
     "entities": "~1.1.1",
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import filter from 'lodash/filter';
 import ReactDOM from 'react-dom';
 import { setComponentsNames, globalizeComponents, promoteInlineExamples, flattenChildren } from './utils/utils';
 import StyleGuide from 'rsg-components/StyleGuide';
@@ -36,4 +37,13 @@ function processSections(ss) {
 components = processComponents(components);
 sections = processSections(sections || []);
 
-ReactDOM.render(<StyleGuide config={config} components={components} sections={sections} />, document.getElementById('app'));
+if (window.location.hash.substr(0, 3) === '#!/') {
+	const targetComponentName = window.location.hash.substr(3);
+	const filteredComponents = filter(components, function (component) {
+		return component.name === targetComponentName
+	})
+	ReactDOM.render(<StyleGuide config={config} components={filteredComponents} sections={[]} />, document.getElementById('app'));
+} else {
+	ReactDOM.render(<StyleGuide config={config} components={components} sections={sections} />, document.getElementById('app'));
+}
+
--- a/src/index.js
+++ b/src/index.js
@@ -40,6 +40,7 @@ components = processComponents(components);
 sections = processSections(sections || []);
 
 
+let hasRenderedFullStyleguide = false;
 function renderStyleguide() {
 	if (window.location.hash.substr(0, 3) === '#!/') {
 		const targetComponentName = window.location.hash.substr(3);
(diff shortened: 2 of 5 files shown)


karlbright:
single-comp

mik01aj:
Looks great. I also reviewed the code and it looks OK. But I would wait with merging until the feature is finished (most importantly, it should be easily discoverable with some image button.

sapegin:
Thank you very much! This is really great and useful feature.

The only question I have is configuration:

1. Should it be optional or how @mik01aj suggests with an icon (probably shown on hover)?
2. Switch to disable default behavior with all components on the same page? So links in the sidebar will switch different single-component pages.

karlbright:
I didn't add the button here just as I wasn't sure how we wanted to tackle that, as well as not being sure that this is how we wanted to tackle the feature. I thought it best to get this "okayed" and merged with a seperate PR for linking from components. The idea was to provide a button/icon to jump to a single component preview, much like you said @mik01aj 

The other issue was definitely around how to handle it with the sidebar like you mentioned @sapegin. At this stage I think it's best to remove the sidebar on single component pages.

sapegin:
Yep, for development mode. But there is a use case when you want to have each component at a separate page in a style guide build.

mik01aj:
I would rather stop here and think more about a good, consistent, intuitive solution.

For example, we have the search box. It filters the menu on the right. But maybe instead of filtering it could just highlight the relevant links, and filter the component docs on the right? Then the text from the search box could be added to the URL hash, and thus the functionality will be the same - with even less changes.

karlbright:
Yes, I'm +1 with @mik01aj on this. This PR is definitely just a look at providing the functionality for viewing a component by itself, which is helpful when building out a component. I'd rather look at how the styleguide can be organised to better provide such functionality in a seperate issue than look at building on this PR.

The only thing I'd maybe look at adding here is the button from a ReactComponent.

sapegin:
+1 for reducing the scope of this PR.

mik01aj:
PS. A quick mockup of my idea:

screenshot 2016-04-12 10 30 56

karlbright:
@mik01aj I like that idea but I don't think it scales very well. At the moment i'm using Styleguidist with 150~ components and this would mean that if I search for "Button" I might have 10~ results scattered throughout the list. Changing the ordering on these wouldn't work either as we have to take sections into consideration.

karlbright:
Thinking about it though, this could definitely work without graying out other options, and just reducing the list to that search input text. So in this case, a search for "Button" would result in the sidebar showing "Button" and "RandomButton", perhaps with a "show all components" link at the bottom of the list to clear it.

This doesn't tackle the issue of showing one component however, but there are some ideas here to play with. I'll mess with it for a quick spike to see what we could do.

mik01aj:
How about an "exact match" checkbox under the search field?

sapegin:
I also already have scrolling so highlighting will not work well. And in your case `RandomButton` should be also highlighted which leads to a problem: how can I show just `Button`?

So I thinks both are useful:

1. Dynamic filtering: filter lists, filter components, update URL.
2. Link to show just one component.

> perhaps with a "show all components" link at the bottom of the list to clear it.

Or just a clear button inside a search field.

> How about an "exact match" checkbox under the search field?

For me it’s more like trying to fix bad UI with more UI ;-)

mik01aj:
>>perhaps with a "show all components" link at the bottom of the list to clear it.
>Or just a clear button inside a search field.
Clear button is not enough. It's too small. There should be a clear message that there are, say, 40 components that are currently hidden because of the search. It makes sense when you come back to an open styleguide windo after some time, or when someone sent you a link pointing to a filtered list. You'll save the users moments of frustration: "where is my component?".

karlbright:
I'm having a look at this now with some quick ideas. Will open a new PR and ping.

sapegin:
Is it ready for merge as a base for other PRs?

karlbright:
@sapegin Yes, I think this is fine for the time being with understanding that it may change.

PR #124, opened 2016-04-12
[SPIKE] Filter component single view

karlbright:
**DO NOT MERGE THIS! THIS IS A SPIKE, TESTING SOME POSSIBLE FUNCTIONALITY. TESTS MAY BREAK. THE IMPLEMENTATION WILL BE HACKED TOGETHER**

filter-focus-styleguide

This is a look at how best to handle "focused components" within the styleguide. A focused component allows the user to view one component alone, this is useful when developing components using styleguide driven development.

The way this implementation works is by filtering the components using the search, then double clicking a component to "focus" on that item. This then removes the other filtered componets from the preview, and fades them out as "inactive" in the sidebar. Any change to the filter input, single clicking another item in the sidebar, or double clicking on the focus component, will clear this mode.

This is part of some ongoing work in #123 

cc @sapegin @mik01aj

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "glob": "6.0.4",
     "highlight.js": "9.1.0",
     "html-webpack-plugin": "2.8.1",
+    "immutable": "^3.7.6",
     "json-loader": "0.5.4",
     "lodash": "4.2.1",
     "markdown-it": "5.1.0",
--- a/src/rsg-components/Components/Components.js
+++ b/src/rsg-components/Components/Components.js
@@ -1,6 +1,6 @@
 import { Component, PropTypes } from 'react';
 import ReactComponent from 'rsg-components/ReactComponent';
-import Renderer from 'rsg-components/ReactComponent/Renderer';
+import ReactComponentRenderer from 'rsg-components/ReactComponent/Renderer';
 import Sections from 'rsg-components/Sections';
 
 export default class Components extends Component {
@@ -12,10 +12,16 @@ export default class Components extends Component {
 
 	renderComponents() {
 		const { highlightTheme, components } = this.props;
-		const ComponentRenderer = ReactComponent(Renderer);
 
 		return components.map((component) => {
-			return (<ComponentRenderer key={component.name} highlightTheme={highlightTheme} component={component} />);
+			return (
+				<ReactComponent
+					renderer={ReactComponentRenderer}
+					key={component.name}
+					highlightTheme={highlightTheme}
+					component={component}
+				/>
+			);
 		});
 	}
 
--- a/src/rsg-components/Layout/Layout.js
+++ b/src/rsg-components/Layout/Layout.js
@@ -1,18 +1,20 @@
 import { Component, PropTypes } from 'react';
 import Components from 'rsg-components/Components';
 import TableOfContents from 'rsg-components/TableOfContents';
+import Renderer from './Renderer';
 
 import s from './Layout.css';
 
 import isEmpty from 'lodash/isEmpty';
 
-const Layout = (Renderer) => class extends Component {
+class Layout extends Component {
 	static propTypes = {
 		config: PropTypes.object.isRequired,
 		components: PropTypes.array.isRequired,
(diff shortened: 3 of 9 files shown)


karlbright:
A few things to note on this as it stands:

- I'm not entirely sold on the double click functionality, but considering that this is some functionality that will not always be used, it could work.
- Currently the search input/filter term is not held anywhere, ideally this will be added to the url as a search/filter query (`/styleguide?filter=button`)
- Focusing an item/component does not change the URL, ideally in a cleaner implementation of this, it would do that, much like #123 does.

mik01aj:
I don't like the way you manage state. Each component can be filtered, or grayed out, and it's not clear how to transition between these states. **All state should be visible.**

The functionality is served, but it's not consistent nor intuitive.

I would actually start by designing the application state - in my idea, I wanted to make the search term the only state we need to handle. Seemingly it's not enough and we also need a flag to force an exact match to show exactly one component. But having a search string for filtering and component selection for isolating one component is too much. And note that these 2 variables can easily get out of sync, causing confusion.

karlbright:
This is a spike. The code behind it is definitely hacked together and it's
not how I'd want to handle it at all. I wasn't going to spend time on this,
it's primarily here to look at the interface and functionality, not the
implementation.

mik01aj:
Btw, maybe in the "isolated mode" we could hide the search field and show a button to go back to "normal mode" again.

karlbright:
@sapegin Keen to get your opinion on how this "focused" mode could be handled.

sapegin:
I need to check out the branch and play with it a little. There are quite a lot of changes.

sapegin:
Sorry for the delay, I was at [React Amsterdam](http://react.amsterdam/) this weekend ;-)

1. Do we really need Immutable for just one feature?
2. Instead of double click I’d prefer an icon (probably on hover).
3. Can just not render hidden components instead of setting `display: none`?

More on the UI. I think filtering and isolated mode are separate things with separate use cases and we should not mix them together. Filtering works fine, we just (probably) need to update the URL.

Isolated mode should be clear. For example we could hide a search field and show a link like “Show all 352 components” like @mik01aj suggests. The same link would be useful for filter too.

karlbright:
Again, this is a spike on a feature. The important thing here is the result,
not the implementation. It's to look at and mess with ideas around the
"focused mode".

  

\- I used Immutable out of habit and it was easy, I didn't need to but it
doesn't matter.

\- Display: none is a hack, we'd ideally not render the components, although I
still believe its better to use render none for a few reasons off the top of
my head.

  

I want to make sure people realise that this implementation isn't final, it's
all one big hack to show how it COULD work for the user.

On Apr 18 2016, at 10:57 pm, Artem Sapegin &lt;notifications@github.com&gt;
wrote:  

> Sorry for the delay, I was at [React Amsterdam](http://react.amsterdam/)
this weekend ;-)

>

>   1. Do we really need Immutable for just one feature?

>   2. Instead of double click I’d prefer an icon (probably on hover).

>   3. Can just not render hidden components instead of setting `display:
none`?

>

>

> More on the UI. I think filtering and isolated mode are separate things with
separate use cases and we should not mix them together. Filtering works fine,
we just (probably) need to update the URL.

>

> Isolated mode should be clear. For example we could hide a search field and
show a link like “Show all 352 components” like
[@mik01aj](https://github.com/mik01aj) suggests. The same link would be useful
for filter too.

>

> —  
You are receiving this because you authored the thread.  
Reply to this email directly or [view it on GitHub](https://github.com/sapegin
/react-styleguidist/pull/124#issuecomment-211415743)

mik01aj:
@karlbright it's not clear for me if you're waiting now for our feedback, or if we wait for new stuff from you :)

And if you wait for feedback, I would very much appreciate some GIF with a demo :)

karlbright:
Closing this for the time being. We need to think about how best to handle the interface for this and not sure if this the best way. I think there is definitely better ways to handle this, which may mean bigger changes, but it's a better path.

sapegin:
Thank you very much anyway!

Issue #51, opened 2015-11-18
Changing state re-renders the component in a way that breaks animation


chrisdrackett:
First off I want to say I'm pretty excited about this project. I've been working on my own styleguide based on much of the same tech, so I'm excited to potentially work along side others.

One thing I'm currently doing is letting example files be either markdown or full on react components. This way if I need to store state or similar to show examples of how a component works I can. I'm still digging into your code, so maybe this is currently possible, but if not it might be a nice thing to have if there is a simple way to provide it.

sapegin:
So you want to be able to use (kind of) `Readme.jsx` instead of `Readme.md` for some components? It’s not possible but you can use `require()` in examples so you can share code between examples, etc.

But it would be nice to be able to store particular examples in files.

sapegin:
Now examples [can have a state](https://github.com/sapegin/react-styleguidist#usage-examples-and-further-documentation). Will it solve your issue?

mik01aj:
@chrisdrackett any feedback? I'd like to close this ticket.

chrisdrackett:
I'll work on this today so you guys can close the ticket. Thanks for checking in, and sorry for the delay!

chrisdrackett:
ok, so I hooked this up on one of my components today. this is a switch component that animates between two states. For some reason in the styleguide the component jumps between its two states without animation. Its almost like its being re-rendered from scratch on state change?

sapegin:
That’s how it works now.

mik01aj:
@sapegin, could you explain this further? I see that [the example styleguide modal](https://github.com/sapegin/react-styleguidist/tree/master/example/lib/components/Modal) uses `state` and `setState`, but the [examples loader](https://github.com/sapegin/react-styleguidist/blob/master/loaders/examples.loader.js) passes only `setState` to it.

---

Ah, I got it, it's done in [the `Preview` component](https://github.com/sapegin/react-styleguidist/blob/master/src/rsg-components/Preview/Preview.js)... but I don't understand why you reload the component each time instead of creating a new react class for the example and then using the native setState and stuff. I mean making a component class like this:

```
let PreviewDemo = React.createClass({

    render: function () {
        try {
            return executeCode(this.props.code, this.state, this.setState.bind(this));
        } catch (err) {
            return makeReadBox(err);
        }
    }
    
})
```

...and then rendering this one in the nested `ReactDOM.render()`.

sapegin:
[...]

PR #134, opened 2016-04-26
setState doesn't trigger rerender

saschatimme:
This pull request should solve the issues in #51. 

The previus method was basically, that we keep the state in memory and at every call of setState we render everything again. The problem is that this breaks animations as well as other "dom-dependent"actions.

This would have the following breaking API changes:

Instead of:
```js
'count' in state || setState({ count : 1})
```

it would just be:

```js
setState({ count : 1})
```

Another breaking change would be, that instead of:

```js
const mockData = require('./mocks');
<Message content={mockData.hello}/>
```

you would have to call require directly in the component:

```js
<Message content={require('./mocks').hello}/>
```

The reason is that I split the example code at the (single) react component. Everything before will be executed in the ```componentWillMount``` and the rest in the ```render```method of a react component (calling setState in a render method is not possible).

Would love to hear some feedback.

P.S.: The examples seem to be broken because of the "dog-names" packages, which runs into an error.


src/rsg-components/Preview/Preview.js
```diff
@@ -57,15 +52,39 @@ export default class Preview extends Component {
...
-			`;
-			let compiledCode = this.compileCode(code);
-			let component = this.props.evalInContext(compiledCode, this.setComponentState.bind(this));
+			let compiledCode = this.compileCode(this.props.code);
+
+			// the code contains the setup of the state and the react component to render.
+			// we split the setup of the state and the react component, since calling setState
+			// in render is not allowed.
+
+			const splitIndex = compiledCode.indexOf('React.createElement');
```
sapegin: Will it work with ES6 classes and functional components?

sapegin:
Thank you very much, this is very cool!

```js
'count' in state || setState({ count : 1})
```

It was a hack so I won’t miss it ;-)

```html
<Message content={require('./mocks').hello}/>
```

It’s fine when you need it only once (I think in most of the cases). Any way to reuse the required module?

mik01aj:
I don't like this implicit separation of code into `componentWillMount` and `render`. And, to be honest, I don't have an idea how to make it explicit in a clean way. I just think that this feature will be confusing. Right now the semantic of a code example is simple: it executes like a regular function. When I define a variable and can't access it one line below, then something is broken.

sapegin:
@mik01aj I agree but it fixes more things than breaks.

mik01aj:
For me it will break many examples, I use mocks pretty often. For example: 

    var mocks = require('./mocks');
    <div>
        <Avatar user={ mocks.exampleUser } />
        <Avatar user={ mocks.exampleUserWithoutAvatar } />
        <Avatar user={ mocks.exampleDeletedUser } />
        <Avatar user={ mocks.exampleUserWithInvalidAvatarUrl } />
    </div>

sapegin:
Maybe we could “hoist” `setState` calls — move only them to the `componentWillMount`? It’s more complicated and fragile but might work for more use cases.

mik01aj:
Let's think again.

We're talking about example usages of a component. If, in order to use a component, you need some special initialization in `componentWillMount`, then your component is not really reusable imho. `getInitialState` should be enough.

So maybe all we need is just some way to set the **initial** state, and then everything else could be done in `render`?

So I would propose this:

```
intialState = {x:1}
<button onClick={ () => setState({x: state.x + 1}) }>{ state.x }</button>
```

We could then run this once to get the initial state, and then just put this in `render`.

sapegin:
Actually I like this idea very much ;-)

src/rsg-components/Preview/Preview.js
saschatimme: In my examples does it work. The React.createElement () is just the result after the JSX transformation. So ```<Button />``` becomes ```React.createElement(button,...)```

saschatimme:
We should probably also consider the case that somebody requires a module and uses this to initiate the  state and his component as well.

For this we need to make some assumptions / heuristics about the example codes.

I will give it a try in the next days :)

mik01aj:
@saschatimme I think it will also work this way. That's what I think:

```
var DemoComponent = React.createClass({
    getInitialState: function () {
         var initialState = {};
         eval(EXAMPLE_CODE);
         return initialState;
    },
    render: function () {
         return eval(EXAMPLE_CODE);
    },
});
```

saschatimme:
I got it running with the idea of @mik01aj, thanks!

I also added a error if the old syntax is used, since the setState would now cause a stack overflow.

Maybe you could could give this version a try with your examples and let me know if I overlooked something :)

sapegin:
I need to try it one day.

tizmagik:
Any update on this?

sapegin:
@tizmagik Sorry didn’t have time to check it.

mik01aj:
I tested it. It seems to fix the animations, but on the other hand, it breaks examples like this:

```
var ButtonToy = React.createClass({
    render: function () {
        return <Button>Click me!</Button>;
    },
});
<ButtonToy /> 
```

I'm OK with this (I mean, I can update my examples so they don't create components), but this is a breaking change.

src/rsg-components/Preview/Preview.js
```diff
@@ -57,15 +52,65 @@ export default class Preview extends Component {
...
+					};
+					initial({}, setStateError, initialStateCB);
+					this.state = state;
+				}
+
+				render() {
+					if (this.state.error) {
+						return <pre className={s.playgroundError}>{this.state.error}</pre>;
+					}
+					const setState = partialState => this.setState(partialState);
```
mik01aj: `setState` in React is defined as `void setState(function|object nextState, [function callback])`; you missed the 2nd argument.

sapegin:
Then we can merge it and combine with hiding code examples by default in a single major release.

src/rsg-components/Preview/Preview.js
```diff
@@ -57,15 +52,65 @@ export default class Preview extends Component {
...
 		try {
-			code = `
-				const state = Object.freeze(${JSON.stringify(this.componentState)});
-				${code}
+			let compiledCode = this.compileCode(this.props.code);
+
+			// the code contains the setup of the state and the react component to render.
+			// we split the setup of the state and the react component;
+
+			const splitIndex = compiledCode.indexOf('React.createElement');
```
mik01aj: This is a bad idea. And after the discussion in the PR, I thought we don't need any splitting anymore.

sapegin: Yeah, I also thought that we’ve found a better solution.

src/rsg-components/Preview/Preview.js
```diff
@@ -57,15 +52,65 @@ export default class Preview extends Component {
...
+			const initial = this.props.evalInContext(initCode);
+
+			// 1) setup initialState so that we don't get an error;
+			// 2) use require data or make other setup for the example component;
+			// 3) return the example component
+			const renderCode = `
+				var initialState = {};
+				${compiledCode.substring(0, splitIndex)}
+				return ${compiledCode.substring(splitIndex)};
 			`;
```
saschatimme: In order to pass the setState Function and the state into our component, we have to use `new Function` and cannot work with `eval` (this happens ins `this.props.evalInContext`). The current implementation uses the fact, that `eval` returns the result of the last statement, but for a with `new Function` generated function this is not the case. Hence to get a proper render Function with a return value, we have to insert somehow a return statement for the last component. This is the only reason `splitIndex` exists.

mik01aj: Why not just add more arguments to `evalInContext` insetad?

saschatimme: what do you mean? I don't see what extra argument could solve the problem.

mik01aj: In `examples.loader`, there was this line: 

    'function(code, setState) {' +

The `setState` callback was passed to the component and it worked fine. If you needed to pass something more, you could just add arguments to this function. Why did you change it to use `new Function()`? Without that change, things would get simpler here.

sapegin:
Would be very nice merge this PR. @saschatimme are you going to finish it?

saschatimme:
Sorry I was quite busy the last two weeks but I will try to find a solution that doesn't involve any code splitting this weekend.

sapegin:
@saschatimme Cooool!

saschatimme:
@sapegin I removed the code splitting and the counter examples from mik01aj are now working

src/rsg-components/Preview/Preview.js
```diff
@@ -57,15 +52,58 @@ export default class Preview extends Component {
 		}
 
 		try {
-			code = `
-				const state = Object.freeze(${JSON.stringify(this.componentState)});
-				${code}
+			let compiledCode = this.compileCode(this.props.code);
```
tizmagik: `const`?

sapegin:
@saschatimme Could you please update your branch and fix conflicts? I’ll merge it after that.

sapegin:
There are a few ESLint warning now.

saschatimme:
@sapegin sorry for that, it should be fixed now.

sapegin:
Thanks! :shipit:

PR #142, opened 2016-05-09
More accurate filename in case of error when parsing component

MoOx:
This avoid "Error when parsing index.js" and offer a better filename like "Error when parsing web_modules/SidebarItem/index.js".


sapegin:
Thanks!

PR #151, opened 2016-06-14
Allow markdown ```example to render interactive example.

uipoet:
Markdown fenced code blocks without a syntax prevent text editors from enabling syntax features within the code block.

This change allows fenced code blocks with `jsx` syntax to render the same as without a syntax specified.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -101,18 +101,23 @@ Styleguidist generates documentation from three sources:
 
       <Button size="large">Push Me</Button>
 
-  One more:
+  One more with generic code fence:
 
   \`\`\`
   <Button size="large">Push Me</Button>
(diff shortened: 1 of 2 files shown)


mik01aj:
If you change this, then you should update the docs too. See: https://github.com/sapegin/react-styleguidist/search?utf8=%E2%9C%93&q=fenced

sapegin:
Then how would we make a JSX code example that shouldn’t be rendered as a playground?

uipoet:
@mik01aj Thanks, I will update relevant docs.

@sapegin How about a unique fence type that won't conflict? We can configure our editors to treat any fence type as an alias to a valid one. I suggest `example` or `playground`.

sapegin:
`example` is fine, thanks!

PR #157, opened 2016-07-22
Improve table of contents usability

smashercosmo:
Added possibility to open component in isolated view from the sidebar menu

Demo

### DIFF
--- a/src/rsg-components/TableOfContents/TableOfContents.css
+++ b/src/rsg-components/TableOfContents/TableOfContents.css
@@ -38,3 +38,14 @@
 	border-color: #1978c8;
 	outline: 0;
 }
+
+.isolatedLink {
+	composes: link from "../../styles/colors.css";
+	margin-left: 5px;
+	display: none;
+	position: absolute;
+}
+
+.item:hover .isolatedLink {
+	display: inline;
+}
--- a/src/rsg-components/TableOfContents/TableOfContents.js
+++ b/src/rsg-components/TableOfContents/TableOfContents.js
@@ -22,6 +22,7 @@ class TableOfContents extends Component {
 				{(components || []).map(({ name }) => (
 					<li className={s.item} key={name}>
 						<a className={s.link} href={'#' + name}>{name}</a>
+						<a className={s.isolatedLink} href={'#!/' + name}>⇢</a>
 					</li>
 				))}
 				{(sections || []).map(({ name, components: subComponents, sections: subSections }) => (


mik01aj:
doesn't the arrow create an empty line with longer component names?

smashercosmo:
Sorry, can't really understand the question)

mik01aj:
I mean that when the component has a long name, like `CampaignsOverviewPageLoader` or so, the arrow might not be visible, or it might even wrap to the next line. And as you use `display: none` for toggling the arrow, it will make the stuff below jump one line down/up on mouse enter/leave.

smashercosmo:
Yes, the arrow will jump to the next line, but it won't make component names jump, because it has position: absolute. Although, that's also not very beautiful. Any ideas how to improve it?

smashercosmo:
We can place link to isolated view on the left side of component name. But in that case I think the icon should be different. Something like [external link icon](http://fontawesome.io/icon/external-link/).

mik01aj:
You could put `white-space: nowrap` on the `<li>`. I also wonder why did you need `position: absolute`...

smashercosmo:
If you put ```white-space: nowrap``` then in case of long component name arrow becomes completely invisible. ```position: absolute``` removes external link from the flow, so it doesn't affect vertical alignment of component name and external link. If you remove ```position: absolute``` you'll see how lines jump.

sapegin:
`display:none` is not the best way to show things anyway. `opacity` with animation like in the examples would be better. I’d just make them `display:inline` instead of `position:absolute`.

mik01aj:
@smashercosmo I don't think it's a big problem if they become invisible.

And honestly, I don't see a big improvement coming from this PR. Imagine, the user comes for a first time, clicks, some innocent-looking arrow and the navigation disappears. The current behaviour is at least consistent and it's easier to go back.

sapegin:
Yeah, I’m also not sure whether it makes UI better of worse ;-)

smashercosmo:
No problem) Feel free to close. Just thought it's useful, when user (if he has a lot of components) don't have to scroll down to the end of the page to be able to open component in isolated view.

sapegin:
No need to scroll, just one extra click ;-)

smashercosmo:
Yep)) You're right. Then again: feel free to close) Anyway, thx for discussion.

smashercosmo:
@sapegin by the way: simple ```display: inline``` will affect line height.

Issue #133, opened 2016-04-25
Emulate responsive behaviour


XOP:
In the very basic implementation there is no common ability to test components against various breakpoints in `@media`-queries - by resizing the browser window or using DevTools. 
Generally, if we hide sidebar and adjust paddings of the components wrapper there is a good approximation for what we are targeting.
Do you have plans to implement the feature in some different way (iframes etc.) or should one use the above-mentioned approach to achieve this? Maybe by injecting custom CSS...

the examples of how it could look like:
![image](https://cloud.githubusercontent.com/assets/1815878/14782983/057780b4-0aed-11e6-9fce-8944f2c693b4.png)
![image](https://cloud.githubusercontent.com/assets/1815878/14782989/0cf2f4ae-0aed-11e6-8b68-fdeaf6573d96.png)

or simply open a new page of a component without wrappers?

sapegin:
Looks useful for me. I’ll be happy to merge a pull request with this feature.

XOP:
I'll take a look into this, but can't promise anything :(
I'm thinking about incorporating react-styleguidist for the project needs, but apparently there's no chance to 'just use it' without this feature and perhaps something else

tizmagik:
> or simply open a new page of a component without wrappers?

I'm a fan of that approach. The responsive Chrome DevTools and in Safari are hard to beat, so I think we just simply need a clean and easy way to isolate a component without the styleguidist chrome.

tizmagik:
I'm with you @XOP , we need this as well. Let me know where you end up or if you need any help.

I think something as simple as a `view isolated` link next to each component name that then goes to a blank page with just that example rendered and no other padding or chrome on the page would be helpful. Then use the browser's back button to go back to the full list of components. That may be a good v1 of this feature.

operatino:
Here's the example implementation of the solution you're discussing:

https://sourcejs.com/specs/example-bootstrap-bundle/carousel/

+ isoleted view:

https://sourcejs.com/specs/example-bootstrap-bundle/carousel/?clarify=true&sections=1
https://sourcejs.com/specs/example-bootstrap-bundle/carousel/?clarify=true&tpl=clear&sections=1

operatino:
This kind of "separate" component view page helps a lot, but solutions with placing the examples in iframes also makes sense, if on one page you would like to show different screen states. We're currently discussing similar plugins for SourceJS.

XOP:
@tizmagik Hey man, thanks for support!
[...]

PR #160, opened 2016-07-27
WIP: Emulate responsive behavior

JulienPradet:
Hi!

This PR is related to #133. 

It's still a WIP. But is this implementation good enough for passing down the configuration ?

Next, I'll have to do:
- a bit of design
- add some scaling behavior if the width is too big compared to the container

Cheers


smashercosmo:
I would rather prefer resizable iframe solution

smashercosmo:
Changing container width is pointless, if component has styles, that depend on the window width

JulienPradet:
That's how it would look like with an iframe.

However, I lose the CSS that's given with  webpack loaders.
Any idea how to manage this case ?

mik01aj:
Imho this should be configurable because:

* some projects have responsive components that change layout depending on container size (using flexbox and sizes in percent) - in this case embedding directly just works and doesn't break styles
* some projects use mostly media queries, and thus require a iframe to preview their responsive css. In this case they need to configure the styles themselves.

I don't see a way to reasonably handle both cases at once.

sapegin:
And also a way to disable this, so an option like:

```
responsivePreview: false | 'iframe' | 'container'
```

smashercosmo:
mmm... can't understand why iframe solution can't handle both cases

mik01aj:
@smashercosmo It can, it just needs another configuration for styles. And currently afaik there's no way to correctly add the styles to the iframe, as the user styles (css) are added to webpack config `entry` (along with the user js and stuff).

smashercosmo:
@mik01aj Well, we can add a button to toggle 'responsive mode' per component. After button is clicked current component's markup and styles are just copied to an iframe. The drawback is of course that user can just test static markup.

mik01aj:
@smashercosmo What do you mean by "styles are just copied to an iframe"? The relevant styles can be in `<head>`, or in external stylesheet, anywhere.

smashercosmo:
@mik01aj I thought, that webpack puts all styles to <style> tags to handle hot reloading.

smashercosmo:
@mik01aj So we need to copy all ```<style>``` tags and all ```<link rel="stylesheet">``` tags. Inline styles will be copied with the markup.

smashercosmo:
@mik01aj so the idea is that we just make a static snapshot of a component in its current state

mik01aj:
This seems hacky. What if you create the iframe before some styles are loaded? Also, the idea of copying hundreds of CSS rules to hundreds of iframes (one per example, and I currently have 120 examples in my SG) doesn't appeal to me.

smashercosmo:
Surely it's hacky) What about loading separate component's page in an iframe? Using the real url?

mik01aj:
Well, actually when I need to test responsiveness of components, I either:

* go to isolated view and resize my window
* add multiple examples, each one wrapped with some `<div style={ {width: 250, border: '2px dashed orange'} }>`. Of course this doesn't work with media queries; but on the other hand, you can see various layouts of your component at once.

smashercosmo:
I've just made a proof of concept of loading real component's url in an iframe. That works pretty well, except of course the fact that you can see your component in it's initial state. What about combined approach: load component's url and then, when page is loaded replace html in an iframe with component's markup in it's current state? Then if you change the state, you just need to replace markup.

mik01aj:
so you load another full-blown instance of styleguidist in the iframe? This will not scale. (But maybe I'm missing something.).

smashercosmo:
Yep, there are of course performance issues with this approach. But this iframe should be rendered only on demand, if user wants to test particular component.

smashercosmo:
But, yes, I'm not objecting: this looks like an overhead. I'm just trying to think of every possible approach to implement this feature. (There is a big chance, that eventually we'll come conclusion, that we don't need one :))

tizmagik:
Can someone help me understand the benefits of an iframe solution vs just using the isolated view without any styleguidist chrome (e.g. sans any borders, margin, padding, etc)? I tend to think the latter approach is more useful because it means you can use the excellent Google Chrome responsive dev tools to simulate device sizes and so on.

JulienPradet:
@tizmagik The main benefit is that it allows you to see several components at once, I guess.

tizmagik:
@JulienPradet but imagine if there were a way to hide the styleguidist chrome while viewing all components, then you could use the responsive dev tools on several components at once. 

I guess my concern is that the iframe solution is not only adding code complexity, but just won't ever be as robust as what's possible with Chrome or Safari's built-in responsive dev tools.

sapegin:
@tizmagik Actually I like this idea very much! At least as a kind of responsive preview MVP.

smashercosmo:
@tizmagik yep, iframe solution is pretty ugly and hacky. no need to reinvent the wheel. there are much better tools to test responsiveness, so the only thing that we need to do is to let them to do their job))

XOP:
@tizmagik Ideally we should be able to see both, having responsive solution for the several views and a dedicated one for some specific things - mobile debugging that comes to mind first.

I also support the idea of hiding styleguidist interface, pretty useful (doing that in the proprietary lib right now)

tizmagik:
@XOP agreed. I think a button or config option to "toggle-styleguidist-chrome" would be extremely useful, both in the isolated view and the full view.

JulienPradet:
Ok, so I'll try to implement this sometime around Sunday.

sapegin:
@JulienPradet Awesome, thanks!

mik01aj:
@tizmagik and others, let's make it clear:

1. We already have 2 views. One shows everything with menu on the left (let's call it **overview**), other shows one component, including all its examples, but without the menu (let's name it **component view**). I understand the new view (let's name it **example view**) should show just one specific example of some component. Correct?
2. should the new example view have its **own URL**? Imho it would be convenient to be able to point the mobile browser directly to the example, or to open it in a new window/tab. Furthermore, in the future we could maybe optimize styleguidist so that it doesn't need to load all the components; it would load just as much is needed to run these examples (I mean, that's something for later - it can be difficult because right now, all components are global).

sapegin:
2. That’s how I see it: with a separate URL. And we could move CodeMirror, babel-standalone, etc. to a separate bundle and do not load it there.

smashercosmo:
I think, that even in this example view you should be able to change component's state and edit markup. Because how then you suppose to test responsiveness for different input data?

sapegin:
@smashercosmo For that use case we can make the sidebar collapsible. Otherwise I don’t see any difference with the current views.

mik01aj:
@sapegin good point about these libraries. So, the URL would be like `#!/MyComponent/example2`?

(I still wait for others' feedback)

smashercosmo:
@sapegin Yes, I was also thinking about the approach.

mik01aj:
@smashercosmo if you want to keep the code editor, why would the new example view be better than the component view? I mean, what's the use case, where you want one example with editor but don't want other examples of the same component (normally there aren't many; for me it's around 80 components with an average of perhaps 1.3 example per component).

smashercosmo:
@mik01aj Well, in most cases it will be ok. There could be situations, when you have many examples or not so many but component itself has a big height, so when you're changing window's width, upper  examples will push down lower examples significantly and you'll have to scroll up and down after every resize.

mik01aj:
Kinda yes, but 1. I don't think it happens so often that we need to put work into it and 2. you still end up having a page with babel and codemirror, so it's hard to test page on mobile. So I would still say no to code editor in example view.

smashercosmo:
Wondering, if it is possible to edit component's example in the list view or isolated view and see changes reflected in the example view in the separate tab :) Well, possible of course, but how complex is it...

mik01aj:
The question is, do you really need this? If you want to test your component for short/long data, make many examples. That's what examples are for. Editing the source code of examples is cool but if we do it too much, it's manual testing. With multiple examples, you at least can see everything at once when you work on the styles. I still say no to code editor.

tizmagik:
Shall we close this one @mik01aj @sapegin ?

sapegin:
Probably it’s too complicated, at least for now. But what we really need?

1. No chrome mode?
2. Collapsible sidebar?
3. Anything else?

sapegin:
Thanks for your work @JulienPradet! I‘m closing this for now.

PR #163, opened 2016-08-08
Support webpack 2

### LINKED ISSUE #146: Webpack 2 support?
Could you add support for Webpack 2?

aaronjensen:
This should support webpack 2 out of the box.

Fixes #146

### DIFF
--- a/src/make-webpack-config.js
+++ b/src/make-webpack-config.js
@@ -38,14 +38,15 @@ module.exports = function(env) {
 			filename: 'build/bundle.js',
 		},
 		resolve: {
-			root: __dirname,
 			extensions: ['', '.js', '.jsx'],
-			modulesDirectories: [
+			modules: [
+				__dirname,
 				path.resolve(__dirname, '../node_modules'),
 				'node_modules',
 			],
 			alias: {
-				codemirror: codeMirrorPath,
+				'codemirror': codeMirrorPath,
+				'rsg-components': path.resolve(__dirname, 'rsg-components'),
 			},
 		},
 		resolveLoader: {
@@ -141,7 +142,7 @@ module.exports = function(env) {
 			],
 			debug: true,
 			cache: true,
-			devtool: 'eval-source-map',
+			devtool: 'eval',
 			stats: {
 				colors: true,
 				reasons: true,
--- a/src/make-webpack-config.js
+++ b/src/make-webpack-config.js
@@ -50,11 +50,17 @@ module.exports = function(env) {
 			},
 		},
 		resolveLoader: {
+			modules: [
+				path.resolve(__dirname, '../loaders'),
+				path.resolve(__dirname, '../node_modules'),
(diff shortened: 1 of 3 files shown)


aaronjensen:
This doesn't quite work yet, there seems to be a problem w/ the loader. Looking into it.

aaronjensen:
Ok, this should be good to go.

sapegin:
Thanks! Does it still work in Webpack 1?

aaronjensen:
Yep as far as I can tell. It'd be good if you gave it a test as well.

src/make-webpack-config.js
```diff
@@ -38,22 +38,29 @@ module.exports = function(env) {
...
-			modulesDirectories: [
+			modules: [
+				__dirname,
 				path.resolve(__dirname, '../node_modules'),
 				'node_modules',
 			],
 			alias: {
-				codemirror: codeMirrorPath,
+				'codemirror': codeMirrorPath,
+				'rsg-components': path.resolve(__dirname, 'rsg-components'),
```
sapegin: This line breaks component replacement with aliases like:

```js
webpackConfig.resolve.alias['rsg-components/Wrapper'] = path.resolve(__dirname, 'lib/styleguide/Wrapper');
```

src/make-webpack-config.js
```diff
@@ -38,22 +38,29 @@ module.exports = function(env) {
 			filename: 'build/bundle.js',
 		},
 		resolve: {
-			root: __dirname,
```
sapegin: I had to revert this line to make my style guide work.

src/make-webpack-config.js
aaronjensen: Can you use `NormalModuleReplacementPlugin` for this instead?

Is this why you're not resolving `rsg-components` relatively? That was the biggest problem I had upgrading to webpack2, I'm not sure how to do that without using this alias, the new `modules` doesn't seem to actually work the same as `root` did.

aaronjensen:
@sapegin thank you for testing. I believe I've resolved both of the issues you pointed out, could you try again please?

src/make-webpack-config.js
```diff
@@ -38,9 +38,16 @@ module.exports = function(env) {
...
 			extensions: ['', '.js', '.jsx'],
-			modulesDirectories: [
+			// Webpack 1
+			root: __dirname,
+			moduleDirectories: [
+				path.resolve(__dirname, '../node_modules'),
```
aaronjensen: For what it's worth, I think that this actually belongs in `root` on webpack1. See: https://github.com/webpack/webpack/issues/472#issuecomment-55706013

sapegin:
Tested again and now it just works with Webpack 1. Merging and thanks again!

sapegin:
And this new squash and merge feature is very nice: I don’t have to ask contributors to do it manually.

PR #174, opened 2016-08-30
Expose webpack stats config on webpack dev middleware

andresgutgon:
## Why?
I want to be able to manage what is the output of dev server. I think passing `webpack.stats` options to dev middleware is a good way. What do you think?

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -9,23 +9,23 @@ config.initialize(); // we need to initialize config before requiring anything e
 var makeWebpackConfig = require('./make-webpack-config');
 
 module.exports = function server(callback) {
-	var app = express();
-       var webpackConfig = makeWebpackConfig('development');
-       var stats = webpackConfig.stats || {};
-	var compiler = webpack(webpackConfig);
+  var app = express();
+  var webpackConfig = makeWebpackConfig('development');
+  var stats = webpackConfig.stats || {};
+  var compiler = webpack(webpackConfig);
 
-	app.use(require('webpack-dev-middleware')(compiler, {
-		noInfo: true,
-               stats: stats,
-	}));
+  app.use(require('webpack-dev-middleware')(compiler, {
+    noInfo: true,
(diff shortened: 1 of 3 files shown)


sapegin:
Thanks! Would be nice to document it with some use cases.

sapegin:
Now there are even more lint errors.

andresgutgon:
I use spaces in my editor :trollface:

sapegin:
Use EditorConfig and run `npm t` before commit. It’s easy.

andresgutgon:
Yes! 🍏

src/server.js
```diff
@@ -10,10 +10,13 @@ var makeWebpackConfig = require('./make-webpack-config');
...
 module.exports = function server(callback) {
 	var app = express();
-	var compiler = webpack(makeWebpackConfig('development'));
+	var webpackConfig = makeWebpackConfig('development');
+	var stats = webpackConfig.stats || {};
+	var compiler = webpack(webpackConfig);
 
 	app.use(require('webpack-dev-middleware')(compiler, {
 		noInfo: true,
+		stats,
```
andresgutgon: I'm not sure if this is right?  I just installed the package from github and I'm getting this error:

```
~/my-code/react-component-ui-styleguide npm run react-styleguide-server

> styleguidist server

/Users/rbd/code/ftb/node_modules/react-styleguidist/src/server.js:19
		stats,
		     ^
SyntaxError: Unexpected token ,
```

This code is using ES6?

Should we change this to `stats: stats`? I did this becasue eslint was complaining about it

sapegin: What Node version are you using? Shorthand properties should be supported in Node 4.x: http://node.green/ and I’m going to rise the minimum Node version to 4 anyway soon.

andresgutgon: was that, I'm using `0.12`. Just today we switched to `4.5.x` :) 🎉 . Thanks

PR #177, opened 2016-08-31
Reproduce undefined error on documented method

andresgutgon:
PR just to reproduce this issue https://github.com/sapegin/react-styleguidist/issues/176


sapegin:
I have different error because of non-existent `Button.sizes`. When I remove it it works fine.

What Node version are you using? Don’t know what else can be different.

andresgutgon:
Yep, sorry looks like is my node version :/
With `0.12` is failing but it's not with `4.5.0`.

Issue #162, opened 2016-08-08
Include in project build


aaronjensen:
Is it possible to include styleguidist in a existing webpack build rather than relying on its own dev server?

I ask because:

1. Things do not seem to work out of the box with webpack 2
2. Needing to maintain two webpack configs is harder than one
3. Many projects already have hmr set up so making that "easy" probably doesn't help too many people

I am imagining a way to include a component either on a standalone page or a route handler that is loaded with the appropriate styleguidist loader. I'm not familiar enough with the setup of styleguidist yet to know exactly what would be involved, but I can imagine it could get to a point of similar ease today without the complexity of having it built via webpack (I'm imagining the package published would be prebuilt like a typical package)

Thoughts?

mik01aj:
👍  from me!

I also thought about this, see https://github.com/sapegin/react-styleguidist/issues/158#issuecomment-234877031

aaronjensen:
@mik01aj it's similar to that, but I'm actually imagining prebuilding styleguidist an exposing a component and webpack loader. It would be similar to https://github.com/blueberryapps/react-bluekit but instead of a watching server to generate a manifest file, it'd be handled by a webpack loader.

oskolsky:
Output syleguide/index.html, build from Webpack 2

`Html Webpack Plugin:
<pre>
  TypeError: Cannot read property '0' of undefined
  
  - ExternalModuleFactoryPlugin.js:18 ExternalModuleFactoryPlugin.<anonymous>
    [wgcp-ingamebrowser]/[webpack]/lib/ExternalModuleFactoryPlugin.js:18:38
  
  - NormalModuleFactory.js:159 
    [wgcp-ingamebrowser]/[react-styleguidist]/[webpack]/lib/NormalModuleFactory.js:159:3
  
  - Tapable.js:75 NormalModuleFactory.applyPluginsAsyncWaterfall
    [wgcp-ingamebrowser]/[react-styleguidist]/[tapable]/lib/Tapable.js:75:69
  
  - NormalModuleFactory.js:144 NormalModuleFactory.create
    [wgcp-ingamebrowser]/[react-styleguidist]/[webpack]/lib/NormalModuleFactory.js:144:8
  
  - Compilation.js:356 Compilation.process [as _addModuleChain]
    [wgcp-ingamebrowser]/[react-styleguidist]/[webpack]/lib/Compilation.js:356:16
  
  - Compilation.js:424 Compilation.process [as addEntry]
[...]

PR #186, opened 2016-09-22
[WIP] Run as component

aaronjensen:
This is a super early proof of concept of what I was asking for in #162 

The idea is that styleguidist would publish a loader and a component. The loader could be added to the user's webpack config (or not) and then they would use `require.context` to get the examples and pass them to the `Styleguide` component.

Sections could be supported easily as well by passing the `Styleguide` component a sections object similar to the one used today but with loaded examples instead of globs.

It could look something like:

``` js
const examples = path => {
  const context = require.context(`styleguidist!.${path}`, true, /\/Readme\.md$/)
  return context.keys().map(context)
}

const sections = {
  'Atoms': { components: examples('./atoms') },
  'Molecules': { components: examples('./molecules') },
}

<Styleguide title="My Styleguide" sections={sections} />
```

`Styleguide` would be the extensibility point as well, if you want to render each example differently you could pass in a `renderExample` function.

This fully supports hot reloading including adding files and removing them (try adding a new `Readme.md` in the example after running `npm start`)

We would have to provide guidance for the user so that they do not inadvertently include the styleguide in the javascript served to users in production; There are lots of ways to do that.

I just wanted to get this up there for early thoughts.


sapegin:
But you don’t use any Styleguidist code here. Of course it’s easy now ;-)

aaronjensen:
Ha. I wasn't saying it'd be easy.

sapegin:
But I like the idea :-)

sapegin:
It actually may be possible to do after all refactorings we did for 5.0. I’m not saying I think that it’s very useful but anyway ;-)

 I‘m closing this for now.

Issue #193, opened 2016-10-11
4.0.x broke hot module reloading


aaronjensen:
On 3.1.1 hot module reloading for stateless functional components worked. On 4.0.0 and 4.0.1 it no longer does for our project.

sapegin:
I can reproduce it but have no idea why. All HMR configuration should be external.

aaronjensen:
@sapegin did you change any components from class components to stateless functional components? stateless functional components cannot reload on their own, they must be wrapped at some point by a class component

sapegin:
I did! Now I’m trying to change them back to classes and it doesn’t help ;-(

tizmagik:
Hm, I'm not sure if that's it because the root component is a proper class component, so it shouldn't be an issue. I can repro this locally too, not sure what happened. It did reload once but subsequent reloads don't happen, very odd.

sapegin:
I can reproduce this too in the example. Will try to find the exact commit that breaks it.

sapegin:
Here is the bad commit: f713a3ee2f0b7da2e78138445b9a015daa10270a. Unfortunately the biggest but not the one where I converted classes to functions.

sapegin:
The only solution I could find is this but I’m not sure it’s better than yours. It’s hack too.

```diff
@ Examples.js:15 @
const Examples = ({ examples }) => {
                         <Playground
                             code={example.content}
                             evalInContext={example.evalInContext}
-                             key={index}
+                             key={Math.random()}
                         />
                     );
                 case 'markdown':
```

aaronjensen:
that seems worse, it'll recreate everything anytime anything is rerendered won't it?

sapegin:
Yeah, I think you’re right. How about 3cf04d13ffdd36c60a2a02129b20bb81bf2db69a? Experimenting with your branch.

sapegin:
Actually I like this approach more: no need to configure anything and performance shouldn’t be a big issue because you’d want to develop component in isolated mode anyway.

aaronjensen:
3cf04d1 looks fine. Which approach are you saying you like more? #200?

sapegin:
@aaronjensen Yeap, #200 and 3cf04d13ffdd36c60a2a02129b20bb81bf2db69a ;-)

aaronjensen:
Awesome. I'll close this and my pull out. Thanks!

sapegin:
Thank you too, the solution is 90% yours! :shipit:

PR #200, opened 2016-10-13
Fix hmr

aaronjensen:
@sapegin this is a brute-force fix for #193 

It'll blow away any state in any components if the reload happens this way, but at least it works.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -16,17 +16,29 @@ import './styles.css';
 global.React = React;
 global._ = _;
 
-if (module.hot) {
-	module.hot.accept();
+// Load style guide
+let config;
+let components;
+let sections;
+let hasRenderedFullStyleguide;
+let key = 0;
+
+function loadStyleguide() {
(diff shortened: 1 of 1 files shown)


sapegin:
I’d try to find where I broke it ;-) 

But I was thinking about something like that because it doesn’t require any set up from the user. How fast it is in comparison to the true HMR?

aaronjensen:
Well, I don't know where it broke before as I don't know exactly how it was working before. I'm guessing by "true HMR" you mean `react-transform-hmr` as it's all hmr (including what is done in this branch). The main problem w/ this branch is that it rerenders the entire page. Ideally we'd only rerender each component as needed.

I'm not totally sure what you'd need in order for that to work but I would guess that each styleguide'd component would need a loader-generated wrapping class component that `require`d and rendered the styleguide'd component. Then, when the styleguide'd component changed, it'd walk back to the class component and be reloaded via `react-transform-hmr`'s transformations.

PR #203, opened 2016-10-22
Show whitespace character for default values

### LINKED ISSUE #196: Show whitespace in default prop
Now if prop’s default value contains only whitespace it renders invisible. I’m not sure about the best solution but at least we can quote whitespace-only values or render spaces as something like `␣`.

image 2016-10-12 at 10 31 09 am

tizmagik:
Closes #196 

image

### DIFF
--- a/src/rsg-components/Props/util.js
+++ b/src/rsg-components/Props/util.js
@@ -1,5 +1,3 @@
-import trim from 'lodash/trim';
-
 /**
  * Remove quotes around given string.
  *
@@ -7,7 +5,7 @@ import trim from 'lodash/trim';
  * @returns {string}
  */
 export function unquote(string) {
-	return trim(string, '"\'');
+	return string.replace(/^\'|^\"|\"$|\'$/g, '');
 }
 
 /**
@@ -18,4 +16,14 @@ export function unquote(string) {
  */
 export function getType(prop) {
 	return prop.flowType || prop.type;
+}
+
+/**
+ * Show starting and ending whitespace around given string.
+ *
+ * @param {string} string
+ * @returns {string}
+ */
+export function showSpaces(string) {
+	return string.replace(/^\s|\s$/g, '␣');
 }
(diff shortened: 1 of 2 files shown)


src/rsg-components/Props/util.js
```diff
@@ -1,13 +1,11 @@
...
-
 /**
  * Remove quotes around given string.
  *
  * @param {string} string
  * @returns {string}
  */
 export function unquote(string) {
-	return trim(string, '"\'');
+	return string.replace(/^\'|^\"|\"$|\'$/g, '');
```
sapegin: Good catch ;-) But I’d even replace it with this module, it also unescape quotes inside the sting: https://github.com/dominictarr/quote-unquote/blob/master/index.js#L14-L20

src/rsg-components/Props/util.js
```diff
@@ -19,3 +17,13 @@ export function unquote(string) {
...
 }
+
+/**
+ * Show starting and ending whitespace around given string.
+ *
+ * @param {string} string
+ * @returns {string}
+ */
+export function showSpaces(string) {
+	return string.replace(/^\s|\s$/g, '␣');
```
sapegin: Maybe it’d be more useful to show different kind of whitespace: space, new lines, etc.

tizmagik: I think it makes sense only to show those if they are at the beginning or end of a line, so I can add support for replacing `\n` with `⏎`

For example, I wouldn't want to see this:

```
text
here
```

Represented as:

```
text⏎here
```

Also, I imagine a string that starts/ends with a new-line would be an extremely rare case, but I can still add it if you think it makes sense. Let me know!

src/rsg-components/Props/util.js
tizmagik: Hm, we only "unquote" so that we don't show the quotes surrounding the stringified values parsed from the defaultProp definitions right? E.g. I wouldn't necessarily want the quotes _within_ my defaultProps touched at all.

Also, the function itself looks broken:

```js
var unquote = function (s) {
  var quote = s[0]
  var single = quote === "'"
  return s.substring(1, s.length - 1)
      .replace(/\\\\/g, '\\')
      .replace(single ? /\\'/g : /\\"/g, quote)
}

unquote("here's my quote"); // "ere's my quot"
```

It seems to eat the first and last char.

sapegin:
:shipit:

PR #204, opened 2016-10-22
Code-split on CodeMirror (<Editor />)

### LINKED ISSUE #198: Do not load CodeMirror until first code editor opened
Try to lazy load as much as possible until the user actually opens a code editor (if it’s closed by default).

CodeMirror is the first and the biggest candidate for code splitting but maybe we can extract something else.

Related issue: #136.

tizmagik:
Closes #198


.gitignore
```diff
@@ -3,3 +3,4 @@ examples/**/styleguide/index.html
 .publish/
 node_modules/
 Changelog.md
+*.build/
```
sapegin: What’s this for?

sapegin:
:shipit:

PR #205, opened 2016-10-24
Adds react-docgen-displayname-handler

### LINKED ISSUE #202: Exception thrown with HOC-Components
We've fallen in love with styleguidist at work and plan on using it in our current react-based project. We've hit one showstopper in combination with [HOC-Components](https://medium.com/@franleplant/react-higher-order-components-in-depth-cf9032ee6c3e#.dgafgyv9j). We get the following exception when working with styleguidist:

```
ReferenceError: MyComponent is not defined
```

Reproducible test case:

```js
// MyComponent.js
import React, { Component } from 'react';

const Enhancer = Composed => class MyHOC extends Component {
  render() {
    return <Composed {...this.props} />;
  }
};

const MyComponent = () => {
  return <h1>Hello</h1>;
};

export default Enhancer(MyComponent);
```

MyComponent.md
```md
React component example:

  <MyComponent />
```

tizmagik:
Fixes #202 

This adds `react-docgen-displayname-handler` to the default `handlers[]` config. I wasn't sure whether to include it in the default config or forcibly add it where it's used in `props.loader.js`. I opted to include it into the configuration to make it more extensible in userland. Feedback welcome!

This also adds a `<WrappedButton />` example which shows off HoC support.


docs/Configuration.md
```diff
@@ -163,9 +163,11 @@ You can change settings in the `styleguide.config.js` file in your project’s r
   ```
 
 * **`handlers`**<br>
-  Type: `Array of Function`, optional<br>
+  Type: `Array of Function`, optional, default: [[react-docgen-displayname-handler](https://github.com/nerdlabs/react-docgen-displayname-handler)]<br>
   Array of functions used to process the discovered components and generate documentation objects. Default behaviours include discovering component documentation blocks, prop types, and defaults. If setting this property, it is best to build from the default `react-docgen` handler list, such as in the example below. See the [react-docgen handler documentation](https://github.com/reactjs/react-docgen#handlers) for more information about handlers.
 
+  > NOTE: By default, `react-docgen-displayname-handler` is included which provides better support for Higher Order Components. If you'd like to add additional handlers, be sure to _add_ to the Array rather than replace entirely.
```
tizmagik: Now that I think about it, this isn't actually possible because we don't provide a "callback" per-say, we just take their `config.handlers` value. 

Hm... any thoughts on how we should handle this?

sapegin: Not sure I even understand this part: “be sure to _add_ to the Array rather than replace entirely”. You mean to copy default list and add your own stuff?

sapegin:
And I agree that making it default handler is a good idea. More flexibility for us, if we ever decide to add more default handlers, and for the users, who could customize handlers as they want.

docs/Configuration.md
tizmagik: > copy default list and add your own stuff
Right, but I think there's no real way to copy the default list, right? I mean at runtime at least, sure you can look at the source and reproduce the list but that's an implementation detail that the user would have to worry about/figure out.

What do you suggest here?

sapegin: I’d suggest to give an example of overriding handlers array that will include `react-docgen-displayname-handler` so people could just copy it and change whatever they want.

tizmagik:
Not sure why the build is failing, looks unrelated to code changes. 🤔

sapegin:
Have no idea. Looks like some issues with npm.

sapegin:
Works fine with Webpack 1.13.3, please merge master into your branch.

tizmagik:
@sapegin build seems to be passing now :shipit:

sapegin:
Cool, thanks!

chrisdrackett:
I put this code in my config and I'm now getting a  `SyntaxError: Unexpected token )`

PR #225, opened 2016-11-14
Enable isolation of a single fenced block example

cef62:
Hi @sapegin 
first of all thanks for this project! I really love it and we use it in several projects.
In many of our library projects is the only available compiled output  and we often need to isolate a single example from a readme file.

This PR add the ability to isolate a single code example to simplify the development process.
Would you like to add this feature to the official repo? 

I'd love to have your feedback!
Cheers

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import _ from 'lodash';
+import isFinite from 'lodash/isFinite';
 import ReactDOM from 'react-dom';
 import {
 	getComponentNameFromHash,
@@ -28,7 +29,15 @@ function renderStyleguide() {
 	let sidebar = true;
 	let singleExample = false;
 
-	const { targetComponentName, targetComponentIndex } = getComponentNameFromHash();
+	// parse url hash to check if the components list must be filtered
+	const {
+		// name of the filtered component to show isolated
+		targetComponentName,
+		// index of the fenced block example of the filtered component isolate
+		targetComponentIndex,
+	} = getComponentNameFromHash();
+
+	// filter the requested component id required
 	if (targetComponentName) {
 		components = [
 			...filterComponentsByExactName(components, targetComponentName),
@@ -37,7 +46,8 @@ function renderStyleguide() {
 		sections = [];
 		sidebar = false;
 
-		if (components.length === 1 && _.isFinite(targetComponentIndex)) {
+		// if a single component is filtered and a fenced block index is specified hide the other examples
+		if (components.length === 1 && isFinite(targetComponentIndex)) {
 			components[0] = filterComponentExamples(components[0], targetComponentIndex);
 			singleExample = true;
 		}
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ import ReactDOM from 'react-dom';
 import {
 	getComponentNameFromHash,
 	filterComponentsByExactName,
+	filterComponentExamples,
 	filterComponentsInSectionsByExactName,
 	processComponents,
 	processSections,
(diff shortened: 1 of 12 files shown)


sapegin:
Hi @cef62! That looks very interesting, thank you very much for the PR!

I’ll review the code a bit later.

Using indexes looks fragile (what if you add or move an example?) but I have no better idea how to implement it.

cef62:
Thanks @sapegin, let me knows you want me to change something!

> Using indexes looks fragile (what if you add or move an example?) but I have no better idea how to implement it.

I agree with you but I didn't want to touch too much your current code structure and all texts and fenced blocks are stored in a single array with just numeric indexes. Anyway I don't think should be a real problem because the index will change only if you add other texts or fenced blocks to the readme file and that shouldn't happen while you are debugging a single example. 
Probably the only problematic scenario is if you bookmark a single example and then the readme file changes. 

I think I can get a more robust solution but that would involve the user somehow to explicitly always specify an ID for the fenced block; personally, as user, I'd prefer avoiding it but it's a viable option.

sapegin:
I think indexes are fine unless someone find better solution that doesn’t affect developer experience.

src/index.js
```diff
@@ -25,15 +26,21 @@ function renderStyleguide() {
...
+	const { targetComponentName, targetComponentIndex } = getComponentNameFromHash();
 	if (targetComponentName) {
 		components = [
 			...filterComponentsByExactName(components, targetComponentName),
 			...filterComponentsInSectionsByExactName(sections, targetComponentName),
 		];
 		sections = [];
 		sidebar = false;
+
+		if (components.length === 1 && _.isFinite(targetComponentIndex)) {
```
sapegin: Please use cherry-picking for Lodash for consistency: `import isFinite from 'lodash/isFinite'`.

sapegin: Looks like we need some comments here to explain all the things we’re doing: isolated component, isolated example, etc.

src/rsg-components/Playground/Playground.css
```diff
@@ -14,7 +14,7 @@
 	composes: font from "../../styles/common.css";
 	composes: border link from "../../styles/colors.css";
 	position: absolute;
-	right: -1px;
+	right: 138px;
```
sapegin: This is not cool ;-)

src/index.js
```diff
@@ -43,6 +50,7 @@ function renderStyleguide() {
 			components={components}
 			sections={sections}
 			sidebar={sidebar}
+			singleExample={singleExample}
```
sapegin: I thinks it would be better to use context here, like for `config`. Probably make a new object in config for `sidebar` and `singleExample`.

cef62:
> We need to do something with the UI. Right now it’s inconsistent and a bit too loud for such a feature (But I we can merge it and think more later.)

Yep the UI is not that good... my intention was to understand if you'd like the feature before thinking on the look & feel. 

> Otherwise it’s great ;-)
Thanks :)

I've addressed your comments, let me know if you want further tweaks.

cef62:
I was thinking to move the link to isolate a single example to the upper-right of the rendered fenced block. Something similar the link you already have in place in the upper-right corner of every rendered component.

What do you think? Would you like I add it to the PR?

sapegin:
I was thinking about that too. And show it on hover. At least it would be more consistent with component isolation UI. So please add it to the PR and I’ll merge it.

cef62:
Great! I'll add it to the PR.

cef62:
I've made the discussed changes, let me know if it's ok for you!

sapegin:
Cool, that’s much better!

cef62:
Thanks! I'm happy you liked the addition 👍

PR #229, opened 2016-11-15
Preview container to overflow the floating components

varya:
If a component has `float: left` (or another float) in styles, current `.ReactStyleguidist-Playground__preview` container does not give it enough space to be rendered. With `overflow: auto` I can ensure that the container's height is enough.


sapegin:
Cool, thanks!

PR #233, opened 2016-11-17
Fix prop warning for markdown text

### LINKED ISSUE #232: `name` prop was undefined warning with Markdown text
The `<Examples>` Component throws a PropType error when the markdown contains headings or other markdown specific components. In that case the name property is `undefined`.

The name property is only required when rendering actual user components. See: https://github.com/sapegin/react-styleguidist/blob/master/src/rsg-components/Examples/Examples.js#L9

marvinhagemeister:
Fixes #232


sapegin:
Awesome, thanks!

PR #236, opened 2016-11-17
Support hot reloading for new example files

aaronjensen:


aaronjensen:
@sapegin I finally got around to getting this working. This uses webpack's context feature to essentially express your intent that you want to require something from a particular directory, so rather than having to check in loader code we can check in the actual webpack compiled code.

This makes hot reloading work for new example files and I'm hoping it'll fix our usage with [hard-source-webpack-plugin](https://github.com/mzgoddard/hard-source-webpack-plugin)

aaronjensen:
Reopened at #237

PR #237, opened 2016-11-17
Support hot reloading for new example files

aaronjensen:
I finally got around to getting this working. This uses webpack's context feature to essentially express your intent that you want to require something from a particular directory, so rather than having to check in loader code we can check in the actual webpack compiled code.

This makes hot reloading work for new example files and I'm hoping it'll fix our usage with hard-source-webpack-plugin

### DIFF
--- a/loaders/utils/js.js
+++ b/loaders/utils/js.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const map = require('lodash/map');
+const path = require('path');
+const escapeRegExp = require('lodash/escapeRegExp');
 
 /**
  * Convert plain JavaScript objects and arrays to text representation: WITHOUT any escaping.
@@ -31,7 +33,22 @@ function requireIt(name) {
 	return 'require(' + JSON.stringify(name) + ')';
 }
 
+function requireMaybe(name, alternate) {
+	const dir = JSON.stringify(path.dirname(name));
+	const file = JSON.stringify(`./${path.basename(name)}`);
+	const fileRegexp = `/\\/${escapeRegExp(path.basename(name))}$/`;
+	const requireAlternate = alternate ? requireIt(alternate) : JSON.stringify(null);
+
+	return [
+		'(function () {',
+		`var req = require.context(${dir}, false, ${fileRegexp});`,
+		`return req.keys().indexOf(${file}) >= 0 ? req(${file}) : ${requireAlternate}`,
+		'})()',
+	].join('\n');
+}
+
 module.exports = {
 	toCode,
 	requireIt,
+	requireMaybe,
 };
--- a/src/index.js
+++ b/src/index.js
@@ -24,8 +24,8 @@ let codeKey = 0;
 function renderStyleguide() {
 	const styleguide = require('styleguide!index.js');
 
-	let components = processComponents(styleguide.components);
-	let sections = processSections(styleguide.sections || []);
+	let components = processComponents(styleguide.components, styleguide.config);
+	let sections = processSections(styleguide.sections || [], styleguide.config);
 	let sidebar = true;
 	let singleExample = false;
 
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -43,18 +43,25 @@ export function flattenChildren(components) {
 	});
 }
 
-export function processComponents(components) {
+function removeComponentsWithoutExample(components) {
+	return components && components.filter(component => component.examples);
+}
+
+export function processComponents(components, config) {
 	components = flattenChildren(components);
 	components = promoteInlineExamples(components);
 	components = setComponentsNames(components);
+	if (config.skipComponentsWithoutExample) {
+		components = removeComponentsWithoutExample(components);
+	}
 	globalizeComponents(components);
 	return components;
 }
 
-export function processSections(sections) {
+export function processSections(sections, config) {
 	return sections.map(section => {
-		section.components = processComponents(section.components || []);
-		section.sections = processSections(section.sections || []);
+		section.components = processComponents(section.components || [], config);
+		section.sections = processSections(section.sections || [], config);
 		return section;
 	});
 }
--- a/loaders/styleguide.loader.js
+++ b/loaders/styleguide.loader.js
@@ -1,12 +1,12 @@
 'use strict';
 
-const fs = require('fs');
 const path = require('path');
 const glob = require('glob');
 const prettyjson = require('prettyjson');
 const pick = require('lodash/pick');
 const utils = require('./utils/js');
 const requireIt = utils.requireIt;
+const requireMaybe = utils.requireMaybe;
 const toCode = utils.toCode;
 
 /* eslint-disable no-console */
@@ -53,15 +53,11 @@ function getNameFallback(filepath) {
  * @returns {string}
  */
 function getExamples(examplesFile, nameFallback, defaultExample) {
-	if (fs.existsSync(examplesFile)) {
-		return requireIt('examples!' + examplesFile);
-	}
+	const alternate = defaultExample
+		? 'examples?componentName=' + nameFallback + '!' + defaultExample
+		: null;
 
-	if (defaultExample) {
-		return requireIt('examples?componentName=' + nameFallback + '!' + defaultExample);
-	}
-
-	return null;
+	return requireMaybe('examples!' + examplesFile, alternate);
 }
 
 /**
@@ -90,10 +86,6 @@ function processComponentsSource(components, config) {
 		console.log('Loading components:');
 		console.log(prettyjson.render(componentFiles));
 		console.log();
-	}
-
-	if (config.skipComponentsWithoutExample) {
-		componentFiles = componentFiles.filter(filepath => fs.existsSync(config.getExampleFilename(filepath)));
 	}
 
 	return toCode(componentFiles.map(filepath => processComponent(filepath, config)));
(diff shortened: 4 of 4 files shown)


aaronjensen:
This actually doesn't fully work yet. It lets you remove examples and readd them, but you can't yet add an entirely new example. I'll keep working on it.

aaronjensen:
Ok, this is working now. It has one side-effect that I'm not a fan of, but am willing to trade for now until I can fix the side-effect. That is that because the skipping of components without examples when `skipComponentsWithoutExample` is `true` happens at runtime, the props loader will be loaded for every component--even ones w/o examples. I have some ideas how I might fix it, but it's a little complicated.

For now though, this is a huge improvement IMO. Thoughts?

aaronjensen:
Ugh, this is more complicated than I thought. I only tested adding an example for an existing component. This does not work for adding an entirely new component and an example for it.

I believe that in order to handle that I will need to change the configuration.

Rather than allowing globs to find components, we'll need to take the same parameters that `require.context` takes--path, include subdirectories and a regexp.

It'd look something like this:

```js
  sections: [
    {
      name: 'General',
      sections: [
        {
          name: 'Particles',
          path: './client/app/components/particles',
          includeSubdirectories: true,
          regExp: /\.js$/,
        },
        {
          name: 'Atoms',
          path: './client/app/components/atoms',
          includeSubdirectories: true,
          regExp: /\.js$/,
        },
      ],
    },
  ],
```

sapegin:
That’s very cool, thanks!

1. I don’t want to make config more complicated and I think it now worth it in this case. Maybe we can convert globs automatically. And don’t forget that `components` can be a function.

(I‘ll be back soon ;-)

loaders/utils/js.js
```diff
@@ -31,7 +33,22 @@ function requireIt(name) {
 	return 'require(' + JSON.stringify(name) + ')';
 }
 
+function requireMaybe(name, alternate) {
+	const dir = JSON.stringify(path.dirname(name));
+	const file = JSON.stringify(`./${path.basename(name)}`);
+	const fileRegexp = `/\\/${escapeRegExp(path.basename(name))}$/`;
+	const requireAlternate = alternate ? requireIt(alternate) : JSON.stringify(null);
+
+	return [
```
sapegin: [One multiline template literal](https://github.com/sapegin/react-styleguidist/blob/master/loaders/examples.loader.js#L57-L75) would be more readable.

src/index.js
```diff
@@ -24,8 +24,8 @@ let codeKey = 0;
 function renderStyleguide() {
 	const styleguide = require('styleguide!index.js');
 
-	let components = processComponents(styleguide.components);
-	let sections = processSections(styleguide.sections || []);
+	let components = processComponents(styleguide.components, styleguide.config);
```
sapegin: Probably better to pass `skipComponentsWithoutExample` explicitly.

src/utils/utils.js
```diff
@@ -43,18 +43,25 @@ export function flattenChildren(components) {
 	});
 }
 
-export function processComponents(components) {
```
sapegin: Could you please add some JSDocs while you’re here? ;-)

sapegin:
I’m back ;-)

2. Could you compare bundle size just in case? `require.context` might include files you don’t want.

src/index.js
aaronjensen: I considered this, but it's less flexible (it may not be the only config option we want to pass). Plus, if this were called from anywhere else w/ a hard-coded boolean it's less easy to know what the arg is:

```js
processComponents(components, false)
// vs
processComponents(components, { skipComponentsWithoutExample: false })
```

Some people resort to comments:

```js
processComponents(components, /* skipComponentsWithoutExample */ false)
```

Of course, this isn't public API, so it's probably moot and I'm happy to switch it to just the flag if you still want it, I just wanted to share my reasoning 😄

sapegin: Maybe you’re right. This part of Styleguidist is a bit messy anyway ;-)

aaronjensen:
> Could you compare bundle size just in case? require.context might include files you don’t want.

On my project, before:

```
$ ll priv/static/styleguide/build
total 5328
-rw-r--r--  1 aaronjensen  staff   276678 Nov 18 07:54 0.js
-rw-r--r--  1 aaronjensen  staff  2445578 Nov 18 07:54 bundle.js
```

after:

```
$ ll priv/static/styleguide/build
total 6472
-rw-r--r--  1 aaronjensen  staff   281126 Nov 18 07:53 0.js
-rw-r--r--  1 aaronjensen  staff  3029668 Nov 18 07:53 bundle.js
```

first--whoa why is that bundle so big? I need to dig into that to figure out if I've got something wrong in my config.

second--I'm guessing that it's now larger because I'm requiring every component rather than discriminating against components that do not have examples. The `require.context` I'm using should be limited to a single file, so I don't think that's making things bigger other than any superficial boilerplate that require.context requires.

aaronjensen:
> I don’t want to make config more complicated and I think it now worth it in this case.



Unfortunately, this is the only way I can think of to make this work. I'd suggest not merging this in until I'm able to fix the scenario of adding an entirely new component. Otherwise, the only thing this pull request adds is the ability to add an example to a component that was already there when the server started, which isn't super useful and it comes at the cost of some extra bundle weight due to it pulling in components that may never get rendered.

I've reached out to webpack folk on gitter to see if there's a way to use `addDependency` with a glob, but my guess is that that is what `require.context` is for.

> Maybe we can convert globs automatically. 

Maybe, but it'd take a decent amount of work and I'd worry about hitting all of the edge cases.

> And don’t forget that components can be a function.

Do people use this functionality? I'm not convinced of its utility and the more options we have the less flexible we can be with changes to the code and the more complicated the code is. Theoretically we could support this as well--we'd just require those manually and not use `require.context` for them. If anyone is doing anything file system globbing on their own or anything like that they're actually going to be breaking hot loading as well. It's not a very "pit of success" feature imo. I could see an argument for a filter function that could filter the result of a context if we really needed to, but even that could be done w/ a regex in many cases. I guess I don't know the real use case of this, so I'm probably missing something.

sapegin:
> I'd suggest not merging this in until I'm able to fix the scenario of adding an entirely new component. 

OK and then we’ll see what we can do.

> Maybe, but it'd take a decent amount of work and I'd worry about hitting all of the edge cases.

At least [we can try](https://npms.io/search?q=glob-to-regexp) and see if that works.

> Do people use this functionality?

It’s the only way to ignore components and it was mentioned in docs many times until I found simpler way to match `Component.js` but not `index.js` in a component folder.

aaronjensen:
> It’s the only way to ignore components and it was mentioned in docs many times until I found simpler way to match Component.js but not index.js in a component folder.

If the goal is only to ignore components, what about a filter function? You could also ignore components with a regexp and negative lookaheads. This would be the best way to do it as they wouldn't get included in the bundle at all in that case.

aaronjensen:
> At least we can try and see if that works.

There's ~~a~~ five packages for everything. I'm skeptical about it, I'd personally rather have direct access to the api and not have to worry about any impedance mismatches.

We could simplify the config by defaulting the regexp to match all js files or maybe all js files that start with a capital letter. We could decide on a default for using subdirectories (probably `true`). We could also let them override the defaults at the root config for all sections.

That way, the only thing normal users have to set is the path, so the config actually gets a little easier than it is today--you don't have to worry about globs, just set a path. Then we allow for regexp modifying for people that have specific use cases.

sapegin:
That’s a lot of assumptions ;-) Not everybody starts file names from capital letter, may use `Component/index.js`. Some people use different extensions (.jsx, .es6), etc., etc. Glob is s simple way to satisfy everybody in my opinion. I don’t really want to change it again.

aaronjensen:
> Not everybody starts file names from capital letter, may use Component/index.js. Some people use different extensions (.jsx, .es6), etc., etc.

Sure, that's what the regex would be for. Personally, I have an easier time writing complex regexes than complex globs, but that's just me. 

That said, let's say the glob-to-regex works well. Then we still have to split that up into path, includesubdirs and regex. We could probably do that by looking for the first `*` or `[` and splitting there, but we'd have to respect escaping and any other glob mechanics like `{`, `?` maybe others, I don't have the glob spec memorized. In other words, we'd have to actually parse the glob and do that split and determine based on the presence of `**` that they want to include subdirs. There's not a direct one-to-one and there are going to be a lot of edge cases. It'd like be a source of annoying/confusing bugs for people. All of this so that people can write globs instead of regexes. 

I can understand not wanting to change the configuration api, but the current configuration api just has too big of an impedance mismatch with webpack.

Rather than try and tackle that problem, would you be open to me working on splitting the resolver out in such a way that I could plug in my own resolver and configure it in a more webpack way? This would mean that the current way would still work, you could still use globs, you just wouldn't get hot reloading for new files--that would be reserved for a new resolver (which I could release as a package).

sapegin:
Yeah, I agree that converting globs to regexes is not the best idea ;-)

> All of this so that people can write globs instead of regexes.

I‘d say to be able to not restart the styleguide when they add new files.

> would you be open to me working on splitting the resolver out in such a way that I could plug in my own resolver and configure it in a more webpack way

Absolutely! And thanks for your time and interest! :shipit:

aaronjensen:
> I‘d say to be able to not restart the styleguide when they add new files.

Well, the most straight forward way to do this would be to change the config options, so the only reason to convert globs to regexes would be to not have to restart AND to maintain the current configuration schema. Noble goal, for sure, but I wouldn't consider it worth it. 

I'll go ahead and work on the alternative resolver and maybe some day you'll want to merge it in 😄 . Thanks for your feedback.

PR #242, opened 2016-11-18
Add support for contextDependencies

aaronjensen:
After much digging I happened across another solution to the hot reloading issue. This one seems to work pretty well (it still doesn't support HardSourceWebpackPlugin, but I'd say that's their bug).

This is similar to what sass-loader added: https://github.com/jtangelder/sass-loader/pull/6


sapegin:
Nice! I‘d even set default value for that based on `components` (would solve problem for most of the users) but I can merge it as is.

aaronjensen:
How would you recommend getting an absolute path from a glob?

sapegin:
Check the library I’ve linked in the comment. So for `lib/components/**/*.js` it would be exactly what you give as an example in docs. It works with a list of files, not with glob itself, so it’s not fragile magic which is very good ;-)

yarn.lock
```diff
@@ -1,5 +1,7 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 # yarn lockfile v1
+
```
aaronjensen: ℹ️  There are so many changes in this file because I added common-dir and yarn'd with 0.17.4. I'm guessing this was originally generated with 0.16.*. I don't know what they changed in yarn, but it made significant changes in my project's yarn.lock as well when we upgraded.

aaronjensen:
Ah, I didn't see that link the first time I read the comment. There really is a package for everything 😄  Thanks for finding that.

I updated the pull. I combined all section components + top level components to find the common dir. Let me know if there's anything you'd like me to change.

yarn.lock
sapegin: Don’t worry, it’s better than shrinkwrap anyway ;-)

sapegin:
Looks very good now! Merged ;-)

PR #243, opened 2016-11-19
Add extra HTML markup to make previews easier to find with CSS selectors

### LINKED ISSUE #238: Add ids to component containers
I would like to ask you to add unique ids to component containers, something like `<div id={ name + '-container' }>`. It would help me have better CSS selectors in my visual testing suite: `#ComponentName-container .ReactStyleguidist-Playground__preview` instead of using `:nth-of-type()`. 

I understand that I can do it by having a custom component like https://github.com/sapegin/react-styleguidist/blob/master/examples/customised/styleguide/components/ReactComponent/index.jsx, but it will make updating styleguidist more difficult: I don't want to monitor upstream for changes in that file.

I also understand that isolated view exists but I can't use it for my task because I'm not using a webserver to host my docs for phantomjs, but feed them directly as `file:///` which makes navigating to #!/ComponentName a bit complicated.

(I can make the change myself and send a PR later if you so prefer)

n1313:
Fixes #238

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ examples/**/styleguide/index.html
 .publish/
 node_modules/
 Changelog.md
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ examples/**/styleguide/index.html
 .publish/
 node_modules/
 Changelog.md
+.DS_Store
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -15,7 +15,7 @@ const PlaygroundRenderer = ({
 	onCodeToggle,
 }) => (
 	<div className={s.root}>
-		<div className={s.preview}>
+		<div className={s.preview + ' rsg--example-preview'}>
 			{singleExample ? (
 				<a className={s.isolatedLink} href={'#!/' + name}>⇽ Exit Isolation</a>
 			) : (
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ examples/**/styleguide/index.html
 .publish/
 node_modules/
 Changelog.md
+.DS_Store
```
sapegin: Please you your global gitignore for that.

n1313:
Hmm, okay, that's unusual but I can redo the PR. Is there any harm in having it ignored explicitly and not depending on an implicit undocumented global setting, though?

sapegin:
It’s [well documented](https://help.github.com/articles/ignoring-files/) and I think it‘s better to configure it once for your system than add to each project separately. You can just revert this commit, I’ll squash everything anyway.

n1313:
I understand your point, and I agree that adding .DS_Store to a global ignore is a good idea. Unfortunately it is not a widely adopted approach (I think this is the only repo I know of that does that) and it is not documented anywhere (the decision to depend on a globally ignored .DS_Store, not the idea of having a global gitignore), so I think this kind of clash is bound to happen again, I'm afraid.

sapegin:
I should make a poll in Twitter ;-) For me it happens maybe once in a year and it could be any kind of artifacts of tools developers use on their computers. I wouldn’t expect every project have [all these things](https://github.com/sapegin/dotfiles/blob/master/tilde/gitignore) in their gitignore.

sapegin:
But probably you’re right and it‘s just my perfectionism ;-) I’ve checked a few popular JS project (React, Redux, Lodash, etc.): all of them have `.DS_Store`, many have `.idea` and other things.

n1313:
Well, I leave it to you to decide whether you want to have things like .DS_Store or npm-debug.log in there. I have reverted my .gitignore change. Please review the PR.

PR #244, opened 2016-11-19
Add mobile styles to improve display on narrow screens

n1313:
On screens narrower than 600px the sidebar will be moved to the top of the page and displayed as a static element to allow content column to take up all of the space.

### DIFF
--- a/src/rsg-components/ComponentsList/ComponentsList.css
+++ b/src/rsg-components/ComponentsList/ComponentsList.css
@@ -1,6 +1,5 @@
 .list {
-	margin: 0;
-	padding: 15px;
+	margin: 15px;
 }
 
 .item {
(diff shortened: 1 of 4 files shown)


src/rsg-components/ComponentsList/ComponentsList.css
```diff
@@ -16,3 +16,10 @@
 .heading {
 	font-weight: bold;
 }
+
+@media (max-width: 600px) {
+	.item .item {
```
sapegin: But why only inside sections?

n1313:
My intention is to have sections separated and sub-sections or components joined in a line.

sapegin:
Yeah, I understand but now if we have no sections, each component will be on it’s own line.

n1313:
Hmmm, right... I think I will have to add some extra classes for that. I will update the PR shortly.

n1313:
OK, done. The change to styles of `.list` and `.search` is done to make sure the sidebar displays properly if search returns no results.

PR #246, opened 2016-11-20
Fix path and styles in "customised" example

n1313:
This PR fixes two issues with `customised` example:
- wrong path to components (so running `npm run start:customised` doesn't show a "no components found" page)
- unrestricted width of component path string (so the headers and prop tables don't get wider than the enclosing container, [before](https://cloud.githubusercontent.com/assets/632081/20463341/6d95d4b6-af6b-11e6-8c97-f9ae3341f1bd.png)/[after](https://cloud.githubusercontent.com/assets/632081/20463344/7354c114-af6b-11e6-93e9-e3dd63889197.png))

### DIFF
--- a/examples/customised/styleguide.config.js
+++ b/examples/customised/styleguide.config.js
@@ -2,7 +2,7 @@ const path = require('path');
 
 module.exports = {
 	title: 'Style guide example',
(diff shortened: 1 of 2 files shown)


sapegin:
Thanks!

PR #247, opened 2016-11-20
Don't display sidebar and search field if there are no components found

n1313:
There is a search field on the "no components found" page (can be encountered in `customised` example before [PR #246](https://github.com/styleguidist/react-styleguidist/pull/246) It is lonely and somewhat useless, and probably shouldn't be there in the first place.

### DIFF
--- a/src/rsg-components/StyleGuide/StyleGuide.js
+++ b/src/rsg-components/StyleGuide/StyleGuide.js
@@ -64,7 +64,7 @@ export default class StyleGuide extends Component {
 				components={this.renderComponents(components, sections, sidebar, singleExample)}
 				sections={sections}
 				toc={this.renderTableOfContents(components, sections)}
-				sidebar={!isEmpty(components) || !isEmpty(sections)}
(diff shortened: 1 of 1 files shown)


src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -64,7 +64,7 @@ export default class StyleGuide extends Component {
 				components={this.renderComponents(components, sections, sidebar, singleExample)}
 				sections={sections}
 				toc={this.renderTableOfContents(components, sections)}
-				sidebar={sidebar}
+				sidebar={!isEmpty(components) || !isEmpty(sections)}
```
sapegin: You need to keep `sidebar` here too, otherwise it breaks isolated view.

n1313:
Whoops, my bad, I thought it was a boolean.

sapegin:
It is boolean, and it’s false isolated mode ;-)

n1313:
Right, it can also be false in isolated mode. It's getting late, I should get some sleep... Is the PR ok otherwise?

sapegin:
Yep, thanks!

Issue #261, opened 2016-12-13
Object references are broken in examples state


AoDev:
Hi, here is a simple example that you can copy as is and try in the styleguide.

It shows three buttons, one is "active". (red, green, blue)
When one is clicked, it is updating the example state so that the active item is a reference to some item in a list (array of objects).

The example fails because it seems that the state update mechanism clones objects or does anything that breaks references. This is a problem for any component example that relies on object references. But, they work correctly in the "real world".

```js
const items = [
  {id: 'red'},
  {id: 'green'},
  {id: 'blue'}
]

initialState = {activeItem: items[0]}

// update state, logs item clicked showing the reference is broken
function setActive (item) {
  console.log(item, item === items[0], item === items[1], item === items[2])
  setState({activeItem: item})
}

<div>
  <button type="text" onClick={setActive.bind(null, items[0])}>red {items[0] === state.activeItem ? 'active' : ''}</button>
  <button type="text" onClick={setActive.bind(null, items[1])}>green {items[1] === state.activeItem ? 'active' : ''}</button>
  <button type="text" onClick={setActive.bind(null, items[2])}>blue {items[2] === state.activeItem ? 'active' : ''}</button>
</div>
```

Note: It is broken right from the start when it is set in the initialState and also later when the state is updated.

sapegin:
It might be because we’re `eval`ing examples’ code. But `setState` is just a normal `setState` of a component that wraps example.

Anyway I’d avoid such code in any “real world” app — that looks very unreactish for me. Store an item ID in the state, it works fine.

AoDev:
@sapegin 
This is a contrived example simply to illustrate the problem.

And I don't know what you mean by Reactish, since a big point of React optimization is based on checking references in the render tree. Hence from the start and until now they have created and improved the pureRender mixin -> shallowCompare -> pureComponent.

I don't think it's a good idea to limit styleguidedist to what you think is the right way. If you don't want to handle this case at least a note should be added that any reference in examples won't work.

sapegin:
If you think it’s a bug please send a pull request with a fix or improved docs.

AoDev:
@sapegin I send a PR with docs updated with a note about this issue. Let me know if it's ok there or if any change is needed.

PR #262, opened 2016-12-13
docs(GettingStarted): state object reference behaviour caveat

AoDev:
Trying to rely on object references in the examples will not work.
@see #261

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -152,3 +152,26 @@ initialState = { key: 42 };
 ```
 
 You *can* use `React.createClass` in your code examples, but if you need a more complex demo it’s often a good idea to define it in a separate JavaScript file instead and then just `require` it in Markdown.
+
+## Code examples caveat
+
+Object references will not work as expected in examples state due to how the examples code is evaluated.
+
+```
+const items = [
+  {id: 0},
+  {id: 1}
+];
+
+initialState = {
+  activeItemByReference: items[0],
+  activeItemByPrimitive: items[0].id
+};
+
+<div>
+  {/* Will render "not active" because of object reference: */}
+  {state.activeItemByReference === items[0] ? 'active' : 'not active'}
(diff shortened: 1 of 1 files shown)


docs/GettingStarted.md
```diff
@@ -152,3 +152,31 @@ initialState = { key: 42 };
 ```
 
 You *can* use `React.createClass` in your code examples, but if you need a more complex demo it’s often a good idea to define it in a separate JavaScript file instead and then just `require` it in Markdown.
+
+## Code examples caveat
+
+Object references will not work as expected in examples state.
+
+This is currently unavoidable due to how the examples code is evaluated.
```
sapegin: Could you please merge these two short sentences into one paragraph?

docs/GettingStarted.md
```diff
@@ -152,3 +152,31 @@ initialState = { key: 42 };
...
 
 You *can* use `React.createClass` in your code examples, but if you need a more complex demo it’s often a good idea to define it in a separate JavaScript file instead and then just `require` it in Markdown.
+
+## Code examples caveat
+
+Object references will not work as expected in examples state.
+
+This is currently unavoidable due to how the examples code is evaluated. 
+
+```
```
sapegin: How about this example? A bit shorter and probably a bit clearer.

```js
const items = [
  {id: 0},
  {id: 1}
];
initialState = {
  activeItemByReference: items[0],
  activeItemByPrimitive: items[0].id
};
<div>
 {/* Will render "not active" because of object reference: */}
 {state.activeItemByReference === items[0] ? 'active' : 'not active'}
 {/* But this will render "active" as expected: */}
 {state.activeItemByPrimitive === items[0].id ? 'active' : 'not active'}
</div>
```

AoDev:
I have updated the PR with your comments.

sapegin:
Cool, thanks!

PR #263, opened 2016-12-15
Update examples deps, add webpack as dep

O4epegb:
Decided to add webpack as dependency for examples, since it is a peer dependency for styleguidist now. Easier to install everything in one go.
Also bumped up other deps, everything seems working correctly.


sapegin:
Thanks!

Issue #257, opened 2016-12-08
Allow code formatting to be customized


bebraw:
Currently it picks up code fragments automatically and skips static ones (i.e. those with language defined). This isn't cool in all cases as you might want to map static to dynamic in order to preserve GitHub highlighting for example.

It would be good if there was a way to define custom mappings against languages and adjust this behavior.

You can see how [Catalog](https://github.com/interactivethings/catalog) does it for reference.

just-boris:
Catalog names custom markdown content definitions as `specimens`. If you would like, we can do here the same. Let's say, that here will be a new option in `styleguide.config.js`:

```js
module.exports = {
  specimens: {
     html: (content) => {/* implementation */},
     color: (content) => {/* implementation */}
  }
};
```

By default, we are also going to enable `example` and `html` specimens by default, to make backwards-compatible behavior.

I have analyzed `examples.loader.js` source to check how hard it would be to implement and it seems like it needs big refactoring to accomplish this. Basically, we need to split up code evaluation logic and static highlight. Now they are both in the same module and depend on each other.

Before start actually doing this, I'd like to gather more demand from users, to be sure that it actually worths that efforts.

okonet:
Related to #269

okonet:
Also related to #267

sapegin:
I have an idea how to solve at least some of this issues:

1.  `javascript` `js`, `jsx` fenced code blocks are rendered as playgrounds by default.
2. Fist line can have a comment to change default behavior:

       ```js // noeditor
       <Button>Push me tender</Button>
       ```

Possible modifiers (names could be changed):

* noeditor — just renders a component, without code editor or any controls like an isolated mode button.
* static — just renders highlighted source

In that case specimens could be implemented as React components, like this:

    ```js // noeditor
    <Color hex="#f00" name="red" />
    ```

Custom modifiers will be easy to support too.

**This will be a breaking change.** What do you think?


(Thanks  @ndelangen for the idea)

n1313:
What will happen to non-tagged code blocks?

sapegin:
Nothing, I think but we won’t recommend them anymore. Will make transition easier.

ndelangen:
Non-tagged code blocks can't have comments, because if the parser can't know the language, it can't know the syntax for comments.

Just tag your code blocks 👍

sapegin:
[...]

PR #265, opened 2016-12-20
evaluate examples from JS-code blocks

just-boris:
Some code editors, (Atom and WebStorm as I know), understand code block language definition and are able to highlight code inside. Look at the screenshot bellow:

image

I'd like to mark my code examples as `js`, but currently React-Styleguidist doesn't make them live.

### DIFF
--- a/loaders/utils/chunkify.js
+++ b/loaders/utils/chunkify.js
@@ -23,7 +23,7 @@ module.exports = function chunkify(markdown) {
 	function processCode() {
 		return (ast) => {
 			visit(ast, 'code', node => {
-				if (node.lang && node.lang !== 'example') {
+				if (node.lang && node.lang !== 'example' && node.lang !== 'js') {
 					let highlighted;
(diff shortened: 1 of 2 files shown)


sapegin:
This looks a bit like a hack for me. But the feature itself is valuable. Could you please take a look at #257? This seems more flexible but default setting could allow `js` to be executed.

just-boris:
Left a comment there. Seems like it will be a big change, that will take a long to implement.

Also, I'd like not to configure this every time, but keep JS evaluable by default. That means, let's merge this PR first, then we can think about Specimens functionality.

sapegin:
But this is a breaking change in comparison to #257. So I can merge this only to the `next` branch now.

sapegin:
I‘m closing this for now. It turned out to be a big topic. We need to think what we really need here and how to implement it with minimum breaking changes.

Related issues: #257, #269, #382.

PR #272, opened 2016-12-28
Fix: Do not display Props heading and section when no props are defined.

### LINKED ISSUE #267: Code blocks with syntax doesn't get evaluated
It seems it only evaluates examples if there is no syntax specified for the code block. I.e. this

2016-12-22 at 17 34

results in following:

2016-12-22 at 17 33

And since I'd like to still have code highlight on GitHub.

I'd say it should be able to parse the language part of the code block and evaluate it if it's `jsx?` match.

And if the syntax isn't specified, it will try to evaluate the contents as JavaScript:

2016-12-22 at 17 30

I'd say the second isn't an issue, although it would be nice if it could check if it's a valid syntax (`try catch`?)

okonet:
Fixes #267

### DIFF
--- a/src/rsg-components/ReactComponent/ReactComponentRenderer.spec.js
+++ b/src/rsg-components/ReactComponent/ReactComponentRenderer.spec.js
@@ -1,0 +1,30 @@
+import test from 'ava';
+import React from 'react';
+import ReactComponentRenderer from './ReactComponentRenderer';
+
+test('should render props section', () => {
+	const actual = shallow(
+		<ReactComponentRenderer
+			name="Test"
+			pathLine="test"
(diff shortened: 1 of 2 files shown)


src/rsg-components/ReactComponent/ReactComponentRenderer.spec.js
```diff
@@ -0,0 +1,30 @@
+import test from 'ava';
+import React from 'react';
+import ReactComponentRenderer from './ReactComponentRenderer';
+
+test('should render props section if there are no props', () => {
```
sapegin: If there **are** props ;-)

okonet:
@sapegin updated.

PR #273, opened 2016-12-29
Support custom endpoints and configuration to the react-styleguidist server

cef62:
As discussed on [gitter](https://gitter.im/styleguidist/styleguidist?at=5864d4c29d4cc4fc5352a8ce) this PR enables styleguidist users to add new endpoints to the underlying express server.

A new field it's been added to `react-styleguidist` config module: `configServer`.

```javascript
module.exports = {
  components: './lib/components/**/[A-Z]*.js',
  configServer(app) {
    app.get(
      '/custom', 
      (req, res) => res.send({ response: 'Server invoked' })
    );
  },
};
```
The registered function receive a pointer to the instanced `express` server and can add to it new endpoints, configure plugins and middlewares.

This addition is vital for projects requiring access to server-side/filesytem features like the integration with `jest` snapshot testing ([snapguidist](https://github.com/MicheleBertoli/snapguidist)) @MicheleBertoli and I are working on.

Let me know if you feel the result is not what we discussed or if you would like some change.

### DIFF
--- a/examples/express/lib/components/CustomEndpoint/CustomEndpoint.js
+++ b/examples/express/lib/components/CustomEndpoint/CustomEndpoint.js
@@ -1,0 +1,31 @@
+import React, { Component } from 'react';
+
+export default class CustomEndpoint extends Component {
+
+	constructor(props) {
+		super(props);
+
+		this.state = { response: 'No Server Response' };
+		this.handleInvokeEndpoint = this.handleInvokeEndpoint.bind(this, 'custom');
+	}
+
+	handleInvokeEndpoint() {
+		fetch('http://localhost:3000/custom', { method: 'GET' })
+			.then(responseObj => responseObj.json())
+			.then(({ response } = {}) => this.setState({ response, error: null }))
+			.catch(() => this.setState({
+				response: null,
+				error: 'Ouch, something went wrong!',
+			}));
+	}
+
+	render() {
+		return (
+			<div>
+				<div>{this.state.response}</div>
+				{this.state.error ? <div>{this.state.error}</div> : null}
+				<button onClick={this.handleInvokeEndpoint}>Invoke server</button>
+			</div>
+		);
+	}
+}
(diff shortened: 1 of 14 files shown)


examples/express/lib/components/CustomEndpoint/CustomEndpoint.js
```diff
@@ -0,0 +1,31 @@
+import React, { Component } from 'react';
+
+export default class CustomEndpoint extends Component {
+
+	constructor(props) {
+		super(props);
+
+		this.state = { response: 'No Server Response' };
+		this.handleInvokeEndpoint = this.handleInvokeEndpoint.bind(this, 'custom');
```
sapegin: Looks like you don’t use this second parameter (custom).

examples/express/styleguide.config.js
```diff
@@ -0,0 +1,34 @@
...
+				test: /\.css$/,
+				include: dirs,
+				loader: 'style!css?modules&importLoaders=1',
+			}
+		);
+
+		return webpackConfig;
+	},
+
+	configServer(app) {
```
sapegin: I think it should be `configureServer`.

scripts/index.js
```diff
@@ -44,5 +45,16 @@ module.exports = function(config) {
 		makeWebpackConfig(env) {
 			return makeWebpackConfig(config, env || 'production');
 		},
+
+		/**
+		 * Return an info object containing the created server instance and the instanced webapck
```
sapegin: webapck

sapegin: Do we need this in the public API? What’s the use case?

sapegin:
Cool, thanks!

examples/express/styleguide.config.js
cef62: Great! I like `configureServer`

scripts/index.js
cef62: As discussed on gitter I'll remove it from the public API

PR #276, opened 2017-01-04
Fix: Style update to code toggle button

bennam:
The border around the show/hide code area doesn't quite line up with the border around the component in Safari on Mac OS 10.12.1. Fixed by adding margin: 0 to button.

screen shot 2017-01-04 at 14 00 06


sapegin:
Thanks! Could you please check in Chrome, Firefox and Opera too? Since it was fine in Chrome at least.

bennam:
Found a different issue in FF with the border misalignment which I've fixed. Should be fine in Chrome, FF, Safari and Opera now :-)

PR #280, opened 2017-01-08
Publish public methods

n1313:
This change adds (basic) support for documenting public methods of components using jsdoc tags `@public`, `@param` and `@returns`.

### DIFF
--- a/src/rsg-components/Methods/Methods.css
+++ b/src/rsg-components/Methods/Methods.css
@@ -1,0 +1,36 @@
+.table {
+	width: 100%;
+	border-collapse: collapse;
+}
+.tableHead {
+	composes: border from "../../styles/colors.css";
+	border-width: 0 0 1px 0;
+}
+.tableBody {
+}
+.cell {
+	padding-right: 15px;
+	padding-top: 6px;
+	vertical-align: top;
+	font-size: 13px;
+}
+.cellHeading {
+	padding-right: 15px;
+	padding-bottom: 6px;
+	text-align: left;
+	font-size: 13px;
+}
+.cellDesc {
+	padding-left: 15px;
+}
+
+.methodParam {
+	min-width: 200px;
+}
+
+.name {
+	composes: name from "../../styles/colors.css";
+}
+.type {
+	composes: type from "../../styles/colors.css";
+}
--- a/src/rsg-components/Methods/MethodsRenderer.js
+++ b/src/rsg-components/Methods/MethodsRenderer.js
@@ -1,0 +1,83 @@
+import React, { PropTypes } from 'react';
+import Code from 'rsg-components/Code';
+import Markdown from 'rsg-components/Markdown';
+import Group from 'react-group';
+
+import s from './methods.css';
+
+function renderRows(methods) {
+	const rows = [];
+	methods.map((method) => {
+		rows.push(
+			<tr key={method.name}>
+				<td className={s.cell}><Code className={s.name}>{method.name}()</Code></td>
+				<td className={s.cell}>{renderParameters(method)}</td>
+				<td className={s.cell + ' ' + s.cellDesc}>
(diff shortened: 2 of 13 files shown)


src/rsg-components/ReactComponent/ReactComponent.js
```diff
@@ -10,12 +11,16 @@ export default function ReactComponent({
 }) {
 	const { name, pathLine, examples } = component;
 	const { description, props } = component.props;
+	const methods = component.props.methods ? component.props.methods.filter((method) => {
```
sapegin: Maybe it’d be better to filter in the loader.

src/rsg-components/Methods/Methods.css
```diff
@@ -0,0 +1,36 @@
+.table {
```
sapegin: We need to extract table with most of this CSS to a separate component and use it for props and methods.

src/rsg-components/Methods/MethodsRenderer.js
```diff
@@ -0,0 +1,83 @@
...
+import s from './methods.css';
+
+function renderRows(methods) {
+	const rows = [];
+	methods.map((method) => {
+		rows.push(
+			<tr key={method.name}>
+				<td className={s.cell}><Code className={s.name}>{method.name}()</Code></td>
+				<td className={s.cell}>{renderParameters(method)}</td>
+				<td className={s.cell + ' ' + s.cellDesc}>
```
sapegin: Better to use `classnames` instead of concatenation.

src/rsg-components/Methods/Methods.css
n1313: I thought of that too, but I am not sure this should be done as a part of this PR.

sapegin: It introduces duplication, so that’d make sense. But it’s up to you ;-)

src/rsg-components/Methods/MethodsRenderer.js
n1313: I agree, but this is how it is done in PropsRenderer which I used as a base for my PR. I decided to not deviate from the "style" of it too much. I think a classname refactoring should be done as a separate PR across the entire codebase.

sapegin: Yeah, don’t worry, it’s just a minor thing ;-)

src/rsg-components/ReactComponent/ReactComponent.js
sapegin: We already have some methods processing in the loader and we never need private methods, so why not send only the data that’s needed by the frontend. But that’s just my opinion ;-)

n1313:
I can try to move it to the `next` branch, but I'd really like to have this change in v4 as well. Is there a way to achieve this painlessly?

sapegin:
Painlessly for me — send two PRs for both branches ;-) But why do you want it in 4.x? 5.x needs testing, more time we spend in 4.x, more we postpone 5.x final release. I already use it in my project without any issues but I’d like more people to try it.

n1313:
I have fixed two out of three things you've commented on and fixed the broken build (very embarrassing). I would very much like to have this change in v4 because I haven't touched v5 at all and have no idea how it works, or what the migration process is, or how difficult it will be to merge this PR into `next`. I will take a look at it as soon as possible, but I can't promise anything b/c I have no clue about v5 codebase at this moment.

sapegin:
Thanks, merged!

I’m going to write a changelog for 5.x in the next few days (it’s going to be hard — there are 144 commits already).

n1313:
Thanks!

sapegin:
Looks like you broke something:

image 2017-01-09 at 6 23 35 pm

n1313:
That's because I'm a doofus and empty arrays are truthy.

Issue #223, opened 2016-11-11
Simplify setup AKA 5.0.0


sapegin:
**Work in the [next](https://github.com/styleguidist/react-styleguidist/compare/next) branch**

Difficult first-time configuration is a known issue that we really want to solve. Here are some ideas for the next major release that might help:

- [x] CSS Modules → jss — simplify styling — #207.
- [x] Easier Webpack configuration — #221.
- [x] Compile Styleguidist’s code to ES5 — get rid of babel-loader.
- [x] Ret rid of most Webpack loaders except JSON loader.
- [x] Add default JSON loader for npm modules when needed.
- [x] Remove include/exclude option requirement for loaders.
- [x] Remove global Lodash, add context option — #249 
- [x] Default / base Webpack config — #222.
- [x] User Webpack config auto discovery — #255 

Postponed:
- CLI configuration tool (maybe) — #208.

Anything else? Any feedback?

/cc @okonet, @varya (as you have fresh experience ;–)

okonet:
Ideally for webpack users it should not require editing project's webpack config and work like storybook (takes your webpack config but allows to extend it). Even better would be just a plugin for webpack. In this setup is really a breeze: https://github.com/carteb/carte-blanche#setup

I can't speak a lot for non-webpack users but sane defaults sound good to me.

Styleguidist own code should definitively be compiled since it is also will increase build speed I think. Also, distributing it is much easier then it's browser ready. Not sure why css-modules is a problem here. You still can generate a separate CSS bundle and provide it along with the `index.html` and `main.js`. But it would be interesting to see what advantages JSS can give here and theming sounds very good.

I don't think CLI tool will help a lot, but it might be a nice addition. Good example is create-react-app.

sapegin:
**Webpack plugin**: I’m sure it’s possible to make a Webpack plugin for Styleguidist too. But I don’t think it should be the primary way of using Styleguidist. There are so many projects without Webpack (like component libraries or separate components). So it’s far from ideal but definitely nice to have.

**Storybook**: it sounds great except they don’t do that. They have default create-react-app-like config that works only when you don’t have any custom config. Otherwise it’s not different from what we have now.

**CSS**: yeah, the main benefit is easier theming. Plus no need for CSS loader (one of them) — less possible conflicts with user’s loaders (same benefit for compiling JS).

**CLI**: I hope it would make initial setup simpler and less scary. And looks great in docs and on site (which I want to make after this).

[...]

PR #283, opened 2017-01-11
Better support and tools for multiple exported components per file

madjam002:
This is more of an experiment but I'd like to know what your thoughts are on this.

There's a few things going on in this PR:

- [x] Fix multiple components from same file causing a duplicate React keys error
- [x] Allow for import statements in examples and for importing components
- [x] Allow for adjacent JSX by wrapping example JSX in an enclosing `<div />`
- [x] Add ability to import `@example` markdown by section

A few notes on a couple of things mentioned above:

### Allow for import statements in examples

When you export named components from a module, Styleguidist will require you to reference them within the module e.g

```js
// Header.jsx

export const H1 = props => (...)
export const H2 = props => (...)
```

```js
// Example
<Header.H1>Hello, world</Header.H1>
```

By allowing the use of import, the example can have an example of how the component might actually be imported in real world usage, and of course it is then able to correctly locate the component:

```js
// Example
import {H1} from 'components';

<H1>Hello, world!</H1>
```

Unfortunately, as you can see in the PR, the way I have done this is with some major hackery. I am more interested however with whether you think this is a good idea, and if so we can then look into refactoring.

### Ability to import `@example` by section

Just like wanting to have multiple small components per file, it might be desirable to have all of the documentation for these components in one file as well. This is difficult because if you create a `Readme.md`, it will be displayed for every component in the single file.

I've solved this by taking advantage of the existing ability to import an external example with `@example`, but now you can reference a section/anchor in that markdown file.

e.g:

```markdown
// Readme.md

This is an example README.

# H1

This is the documentation for `<H1 />`.

    <H1>Hello, world!</H1>

# H2

You get the idea...
```

```js
// Header.jsx

/** @example ./Readme.md#H1 */
export const H1 = props => (...)

/** @example ./Readme.md#H2 */
export const H2 = props => (...)
```

This enables all of the documentation for relevant components to be in one place, which is great for small components with a minimal API.

Let me know what your thoughts are, thanks!

### DIFF
--- a/loaders/examples.loader.js
+++ b/loaders/examples.loader.js
@@ -24,7 +24,7 @@ function examplesLoader(source) {
 	source = source.replace(COMPONENT_PLACEHOLDER_REGEXP, componentName);
 
 	// Load examples
-	let examples = chunkify(source);
+	let examples = chunkify(source, { section: query.section });
 
 	// We're analysing the examples' source code to figure out the require statements. We do it manually with regexes,
 	// because webpack unfortunately doesn't expose its smart logic for rewriting requires
--- a/loaders/props.loader.js
+++ b/loaders/props.loader.js
@@ -23,7 +23,7 @@ const REQUIRE_PLACEHOLDER = '<%{#require#}%>';
  */
 function parseProps(file, source, propsParser) {
 	try {
-		return castArray(propsParser(file, source)).map(propsToCode);
+		return castArray(propsParser(file, source)).map(propsToCode.bind(null, file));
 	}
 	catch (exception) {
 		console.log('Error when parsing', path.relative(process.cwd(), file));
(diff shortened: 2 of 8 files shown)


sapegin:
Thank you very much for your pull request! This all very interesting and I’m sure useful. 

Would you mind to send a new PR to the `next` branch? It’s a new major release which is almost ready (code and docs are 95% done and more testing is needed), see #223 for more details. The reason I’m asking you to do a lot more work is: every big pull request will delay new release more since I’ll have to port it to the new branch, and code base is very different in some parts (especially loaders). Probably even better to split this into multiple PRs to make it easier to review and merge.

Feel free to ping me in Gitter (or by email/Facebook/Skype) if you have any questions.

madjam002:
Sounds good, I'll also split it into multiple PRs. Thanks 😃

sapegin:
@madjam002 Thank you for understanding! :handshake:

madjam002:
Just FYI - I'll just wait for `next` to be merged into master as we're happy using my fork internally for now and it would be great to get some of these ideas explored further once `next` is merged.

sapegin:
@madjam002 Sounds good!

PR #285, opened 2017-01-12
Added es6-object-assign polyfill

bennam:
Previewing components doesn't work in IE 11 so added es6-object-assign polyfill.

PR #287, opened 2017-01-14
Fix production build

nkt:
#285 break my builds due `es6-object-assign` package used by production code.


sapegin:
Already fixed in 2bb09b7dcf70610d3b634825bdbc99abee795d47. New version should be available in 10 minutes.

PR #288, opened 2017-01-19
Publish public methods

n1313:
This PR brings the changes from https://github.com/styleguidist/react-styleguidist/pull/280 into `next` branch.


loaders/props-loader.js
```diff
@@ -33,6 +33,15 @@ module.exports = function(source) {
 		console.log();
 	}
 
+	if (parsedProps.length) {
```
sapegin: This check is unnecessary, forEach will do nothing for an empty array.

examples/basic/lib/components/CounterButton/CounterButton.css
```diff
@@ -0,0 +1,10 @@
+.button {
```
sapegin: We don’t use CSS Modules here and already have a `.button` class in `Button` component.

src/rsg-components/ReactComponent/ReactComponentRenderer.js
```diff
@@ -50,6 +50,9 @@ const styles = ({ font, monospace, light }) => ({
 	props: {
 		marginBottom: 30,
 	},
+	methods: {
```
sapegin: Could be `.section` or something like that, and would work for both props and methods.

sapegin:
Could you please also add tests for the loader?

loaders/props-loader.js
n1313: Without this check the test for props-loader fails. I don't understand the reason, but in test env `parsedProps` is an object, and in dev/prod env it's an array of objects. I was going to ask you about it in chat but it was late so I just slapped a check on it and went to bed -___-

n1313:
Yes, this is a copy of code from master with minor changes.

loaders/props-loader.js
sapegin: OK, I’ll take a look later.

src/rsg-components/ReactComponent/ReactComponentRenderer.js
sapegin: For some reason react-docs’ `parse` can return either array or object: 
https://github.com/reactjs/react-docgen/blob/master/src/parse.js#L52

And it’a actually expected, see `castArray` below your code. We just need to move it upper.

You can cherry-pick this commit: https://github.com/styleguidist/react-styleguidist/commit/d9c933bad017f2d4e629f0cc34f6d26369004dfe

sapegin: And this one: 53df991b4f1bc8d9d037ed70c3bb388ea8bf0c3a

n1313:
I've updated the PR. Please take a look!

sapegin:
Awesome! Can you add some tests and I’ll merge it?

PR #290, opened 2017-01-21
Add line break between description and extra when rendering props

goodmorninggoaway:


src/rsg-components/Props/PropsRenderer.js
```diff
@@ -59,6 +59,7 @@ function renderDescription(prop) {
 	return (
 		<Group>
 			{description && <Markdown text={description} inline />}
+			{extra && description && <br />}
```
sapegin: Better to use `<Group separator={<br />}>`.

sapegin:
Thanks!

PR #306, opened 2017-02-06
style: Use same code style to render `null` value

okonet:
Using Boolean expression as for props. Prior method was rendering `false` which isn't a valid React node value.

+ Add tests for methods section.


sapegin:
Thanks! Especially for new tests! 🦄

PR #308, opened 2017-02-08
Refactoring: sections and components relationship

okonet:
This PR aims to simplify rendering concerns by using only one data structure — `sections`. Prior there were 2 data structures: `components` and `sections`. The issue is that sections can also contain components which made it complicated IMO. By adding the root `components` section as a first "section" this allows to think about the whole structure as StyleGuide -> Sections -> Components which should make it easier to contribute.

This PR also makes some improvement in how some elements are rendered:

1. Table of contents now produces less markup by removing a `<span>` that was wrapping each `<a>`
1. TOC also uses `<ul>` and `<li>`s that also make it more semantic
1. Sections now use `<section>` HTML tag which makes it more semantic
1. It's now possible to have sections without `name` — makes it more ready for single component use-case (aka component web-site)
1. It tries to leverage React's composition approach by using `children` whenever possible instead of passing custom props.


src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -32,41 +33,26 @@ export default class StyleGuide extends Component {
...
 			<StyleGuideRenderer
 				title={config.title}
 				homepageUrl={HOMEPAGE}
-				components={this.renderComponents(components, sections, sidebar)}
-				sections={sections}
-				toc={this.renderTableOfContents(components, sections)}
-				sidebar={(isEmpty(components) && isEmpty(sections)) ? false : sidebar}
-			/>
+				toc={<TableOfContents sections={sections} />}
+				sidebar={isEmpty(sections) ? false : sidebar}
```
sapegin: When will it be empty if you put components as a first section?

sapegin: I see, not in a normal use case. Maybe extract to a variable like `noComponentsFound`?

n1313:
Could you please update the readme and faq files with examples of usage?

okonet:
@n1313 with examples of what? This is a refactoring work that doesn't change any of the public APIs.

sapegin:
@n1313 The only API change is that section name is now optional.

n1313:
Oh, I see. I misunderstood.

PR #309, opened 2017-02-09
Adds support for IE11 

### LINKED ISSUE #299: 5.x beta doesn‘t work in IE11
Would be cool if anyone have time to investigate and fix it :shipit:

n1313:
Hopefully, fixes #299

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "es6-object-assign": "~1.0.3",
     "escodegen": "^1.8.1",
     "findup": "^0.1.5",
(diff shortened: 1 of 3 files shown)


src/rsg-components/Styled/Styled.js
```diff
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import addStyles from '../../styles/addStyles';
+import 'function.name-polyfill';
```
sapegin: This should be in `src/index.js` — we already have some polyfills there.

PR #310, opened 2017-02-09
Feat: Add showSidebar config option

okonet:
This PR adds a new configuration option `showSidebar`. If set to `false` sidebar will never appear. By default it is set to `true` to preserve the original behavior.

Additionally, this PR refactors:

1. Use context instead of passing down `isolatedComponent` prop
1. Use less ambiguous variable names: 
- `sidebar` -> `isolatedComponent`
- `sidebar` -> `hasSidebar`
- `singleExample` -> `isolatedExample`


sapegin:
Awesome! Thank you! 🦄

Issue #314, opened 2017-02-10
How to add custom media-queries?


n1313:
...in the `next` branch? I can add custom styles by using `styles` property of the config, as described in the documentation, but adding `'@media (max-width: 800px)': { ... }` there doesn't seem to work.

sapegin:
I’m not sure what exactly are you doing and why it doesn’t work but there are some examples:

https://github.com/styleguidist/react-styleguidist/blob/next/src/styles/theme.js#L15
https://github.com/styleguidist/react-styleguidist/blob/next/src/rsg-components/StyleGuide/StyleGuideRenderer.js#L40-L44

n1313:
Ok, I see the problem now. According to JSS docs (https://github.com/cssinjs/jss/blob/master/docs/json-api.md), media queries go like this:

```
const styles = {
  button: {
    width: 100
  },
  '@media (min-width: 1024px)': {
    button: {
      width: 200
    }
  }
}
```
So here we have `styles['@media (min-width: 1024px)'].button { width: 200 }`. And this approach *didn't* work for me.

Using your example as a guide I wrote this:

```
const styles = {
  button: {
    width: 100,
    '@media (min-width: 1024px)': {
      width: 200
    }
  }
}
```
Here we have `styles.button['@media (min-width: 1024px)'] { width: 200 }`. And this *did* work.

Not sure if this is a bug in RSG implementation of JSS, or JSS has incorrect documentation. Or maybe I did something wrong. Anyway, I'm happy now.

sapegin:
We’re using jss-nested: https://github.com/cssinjs/jss-nested#use-at-rules-inside-of-regular-rules

n1313:
Then the "we're using JSS Syntax" link in the documentation (https://github.com/styleguidist/react-styleguidist/blob/next/docs/Configuration.md) should be corrected.

n1313:
Since JSS and JSS-nested have different, mutually exclusive, syntaxes when it comes to nested selectors, the docs should be pointing at JSS-nested documentation, not JSS's. Let me send a PR...

sapegin:
There are other plugins that changes syntax like camelizing selectors, etc. So probably better to list all plugins.

PR #315, opened 2017-02-10
Update docs to point to JSS-nested instead of JSS

n1313:
As mentioned in #314


sapegin:
This is even more confusing ;-)

> There are other plugins that changes syntax like camelizing selectors, etc. So probably better to list all plugins.

PR #318, opened 2017-02-12
Feat: Better anchors navigation with unique slugs

okonet:
This PR:

- Fixes anchors for sections with multiple words in name
- Fixes the case when components or sections with the same name could not be resolved by anchor
- Adds anchor navigation to the title of section/component


tizmagik:
This LGTM, although I wonder if the `slug` can be derived from `name` instead of requiring folks to explicitly define it? E.g. maybe it can be as simple as:

```js
const genSlug = (name) => name.replace(/\s/g, '_').toLowerCase()
```

In most cases that should do it, maybe if there's special characters or something we can put in a regex for that, etc? Just a thought!

okonet:
It is derived! Not sure what do you mean by explicitly requiring it.

src/utils/utils.js
```diff
@@ -16,6 +20,22 @@ export function setComponentsNames(components) {
 	return components;
 }
 
+export function setSlugs(sections) {
```
sapegin: Probably better to extract this to a separate method like `sectionSlugs` to make its singletonnes more clear. Also completely hide `slugger` and export two methods: `reset` and `slugifySections` (since it’s not generic).

Can we do it in the `styleguide-loader`?

src/rsg-components/Heading/HeadingRenderer.js
```diff
@@ -0,0 +1,53 @@
...
+		'&:hover, &:active': {
+			isolate: false,
+			color: linkHover,
+			cursor: 'pointer',
+		},
+	},
+});
+
+export function HeadingRenderer({ classes, children, slug, hierarchy, ...props }) {
+	return React.createElement(`h${hierarchy}`, {
```
sapegin: Any reason not to use JSX?

```js
const Tag = `h${hierarchy}`;
return (
  <Tag>...</Tag>
);
```

src/rsg-components/Heading/HeadingRenderer.js
```diff
@@ -0,0 +1,53 @@
...
+		children
+	);
+}
+
+HeadingRenderer.propTypes = {
+	children: PropTypes.node,
+	className: PropTypes.string,
+	classes: PropTypes.object.isRequired,
+	slug: PropTypes.string.isRequired,
+	hierarchy: PropTypes.oneOf([1, 2, 3, 4, 5, 6]).isRequired,
```
sapegin: Maybe `level` would be a better name.

src/rsg-components/Heading/HeadingRenderer.js
okonet: Additional variable allocation. Not even sure it's more readable. But sure, why not.

src/rsg-components/Heading/HeadingRenderer.js
okonet: It is a better name indeed!

src/rsg-components/Heading/HeadingRenderer.js
sapegin: Yeah, it’s wasn’t a request to change, just my opinion ;-)

src/utils/utils.js
sapegin: I’m thinking more about this: can we avoid singleton at all if we move it somewhere to `utils/getSections.js`. But I’m not sure I’d work and make code better.

src/rsg-components/Heading/HeadingRenderer.js
okonet: I tried and it actually looks better in code.

src/utils/utils.js
okonet: Yes, that would be great. As a follow up question: do we even need to transform components and sections in the `index.js`? Can this also be moved to the loader's code? https://github.com/styleguidist/react-styleguidist/blob/808fd34b4d27b5b8637700098d151dde7e0b7b02/src/index.js#L28-L28

Ideally index.js should not be concerned about it IMO.

sapegin: Yeah, I think this can be moved to loader / simplified. The only things that should be done there is globalization and name intospection.

okonet: Hmm, why can't this also be done in the loader? Can we pass the reference to `window` to it to make it work from one place? I assume the loader isn't a part of API so we can just use some `imports-plugin` for that, right?

sapegin: You can communicate with a loader only via query parameters and loaders return JS code as a string.

sapegin:
Wow, this GitHub conflict merging is really works! ;-)

n1313:
This doesn't bring any changes to public configurations, right?

okonet:
Yes, no changes to config.

sapegin:
@okonet You asked me to do that ;-)

PR #320, opened 2017-02-12
Show content recommended in previous step

coryhouse:
I found it confusing to show this snippet without the content from the previous step, so suggest showing it integrated.


sapegin:
Thanks! We have a new docs for the 5.x release. I hope there are much better, would be cool it can take a look:
https://github.com/styleguidist/react-styleguidist/tree/next/docs

PR #321, opened 2017-02-13
Fix readme links

n1313:


sapegin:
Thanks, it was bad copypaste ;-)

PR #329, opened 2017-02-17
Fix markdown headings

okonet:
Current version renders `<h3>` instead of `<h1>` for the following snippet `# Heading`. 

According to W3C this is the desired approach for outlining structure in HTML5: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines

This PR eliminated altering of markdown headings and uses more HTML5 compliant markup with `<article>` for markdown documents.

### DIFF
--- a/examples/sections/docs/Two.md
+++ b/examples/sections/docs/Two.md
@@ -1,1 +1,26 @@
-Second *Markdown* file.
+# Heading 1
+
+## Heading 2
+
+### Heading 3
+
+#### Heading 4
+
+##### Heading 5
+
+###### Heading 6
+
+foo|bar
+---|---
+1  |2
+
+* list 1
+* list 2
+  1. Sub-list
+  1. Sub-list
+  1. Sub-list
+* list 3
+
+* [ ] list 1
+* [ ] list 2
+* [x] list 3
--- a/src/rsg-components/Markdown/Markdown.js
+++ b/src/rsg-components/Markdown/Markdown.js
@@ -20,6 +20,16 @@ const styles = ({ font, monospace, link, linkHover, border, codeBackground }) =>
 		lineHeight: 1.5,
 	},
 	a: linkStyles({ link, linkHover }).link,
+	h1: {
+		composes: '$para',
+		fontSize: 36,
+		fontWeight: 'normal',
+	},
+	h2: {
+		composes: '$para',
+		fontSize: 32,
+		fontWeight: 'normal',
+	},
 	h3: {
 		composes: '$para',
 		fontSize: 26,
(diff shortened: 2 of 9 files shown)


src/rsg-components/Markdown/Markdown.js
```diff
@@ -20,6 +20,16 @@ const styles = ({ font, monospace, link, linkHover, border, codeBackground }) =>
 		lineHeight: 1.5,
 	},
 	a: linkStyles({ link, linkHover }).link,
+	h1: {
```
sapegin: I think we should rename existing styles, otherwise all existing headers will be rendered with increased size. What do you think?

examples/sections/docs/Two.md
```diff
@@ -1 +1,26 @@
...
+
+### Heading 3
+
+#### Heading 4
+
+##### Heading 5
+
+###### Heading 6
+
+foo|bar
```
sapegin: Does it render tables? 😮

src/utils/markdown-to-jsx.js
```diff
@@ -245,10 +244,8 @@ function getHTMLNodeTypeFromASTNodeType(node) {
 		case 'footnoteReference':
 			return 'a';
 
-		// Increase level of headings
 		case 'heading':
-			const depth = Math.min(node.depth + 2, 6);
-			return `h${depth}`;
+			return `h${node.depth}`;
```
sapegin: That’s good: closer to the original plugin. Eventually we’ll need to stop using our fork.

okonet: At this point, I'm not even sure why we need a fork...

examples/sections/docs/Two.md
okonet: Yes, but without styles 🤷‍♂️

src/rsg-components/Markdown/Markdown.js
okonet: What do you mean by "all existing headers"?

examples/sections/docs/Two.md
sapegin: I’ve removed styles because I couldn’t make it render a table 😭

src/rsg-components/Markdown/Markdown.js
sapegin: If I have `# Foo` in my style guide it’d be rendered bigger after this PR.

src/utils/markdown-to-jsx.js
sapegin: Probably we don’t, I should try to remove it.

src/rsg-components/Markdown/Markdown.js
okonet: Yeah I realize that but I don't think this a big issue. We still can put into a breaking change (which it isn't per se)

PR #330, opened 2017-02-18
WIP: Static rendering

okonet:
This PR implements static rendering of the styleguide when building for production.

🚧🚧🚧
Please don't merge yet. Still WIP!!!
🚧🚧🚧


examples/.gitignore
```diff
@@ -0,0 +1 @@
+**/styleguide/*.js
```
sapegin: We already have rules in global gitignore:

```
/examples/**/styleguide/build
/examples/**/styleguide/index.html
```

Scripts should be in a separate folder anyway.

okonet: Oh I see. Is there a reason for them to be in a separate dir BTW?

sapegin: Better UI: not to scare users with extra files and make `index.html` easier to find ;-)

okonet:
It looks like I was able to render everything besides the actual components (which is by design) statically using latest commit. 

This includes:

* TOC (Table of Contents) with proper slugs (this took some time)
* Markdown documents
* Code highlighted with highlight.js

We'll need to take some care of injecting CSS into `head` during the build time as well.

package.json
```diff
@@ -58,15 +57,18 @@
     "loader-utils": "^0.2.16",
     "lodash": "^4.17.4",
     "minimist": "^1.2.0",
+    "node-noop": "^1.0.0",
```
sapegin: `lodash/noop` should be enough ;-)

scripts/build.js
```diff
@@ -4,8 +4,16 @@ const webpack = require('webpack');
 const makeWebpackConfig = require('./make-webpack-config');
 
 module.exports = function build(config, callback) {
-	return webpack(makeWebpackConfig(config, 'production'), (err, stats) => {
-		// require('fs').writeFileSync('stats.json', JSON.stringify(stats.toJson()));
-		callback(err, stats);
+	// Firstly, build static "server" bundle
+	return webpack(makeWebpackConfig(config, 'production', true), (err, stats) => {
```
sapegin: Can we pass an array of webpack config here?

src/rsg-components/HtmlDocument/HtmlDocumentRenderer.js
```diff
@@ -0,0 +1,34 @@
+import React, { PropTypes } from 'react';
+import HTMLDocument from 'react-html-document';
+
+function getScrtips(assets) {
```
sapegin: This should be in `HtmlDocument.js`. The idea is to provide a simple way to redefine rendering of components, so data and markup should be separated.

src/rsg-components/HtmlDocument/HtmlDocumentRenderer.js
```diff
@@ -0,0 +1,34 @@
...
+	// TODO: Handle CSS and JS assets separately
+	return (
+		<HTMLDocument
+			title={ title }
+			scripts={getScrtips(assets)}
+			metatags={[
+				{ name: 'charset', content: 'utf-8' },
+			]}
+		>
+			{ children }
```
sapegin: I’m wondering why ESLint don’t complain about this spaces… 🤔

src/rsg-components/HtmlDocument/HtmlDocumentRenderer.js
okonet: Agreed! Will refactor appropriately.

src/rsg-components/HtmlDocument/HtmlDocumentRenderer.js
okonet: No rule defined?

scripts/build.js
sapegin: As I understand you can call webpack like this: `webpack([config1, config2])`.

src/rsg-components/HtmlDocument/HtmlDocumentRenderer.js
sapegin: I’ll check ;-)

scripts/build.js
okonet: Yes, but this isn't the same. This will produce 2 compiler's processes but we want to share one process.

sapegin: How do you share it? I see two separate `webpack` calls with different configs.

okonet: Actually you might be right about it. I was a bit confused. Let me try it out.

kof:
@okonet I am not aware of decision details, it seems like react-jss would be a good fit here. In any case its easy. You can always create a `SheetsRegistry` manually and add the sheets to it and then get them all together.

okonet:
@kof can you commit the solution you're suggesting?

sapegin:
@okonet @kof Basically [these two lines](https://github.com/styleguidist/react-styleguidist/blob/next/src/rsg-components/Styled/Styled.js#L10-L11). We need to merge styles with user’s styles and and also pass merged theme object. And strip `Renderer` for nicer class names. 

I’m not sure it’s all possible to do with react-jss and what would be the best solution.

kof:
Btw. `WrappedComponent.name` is not supported everywhere, haven't found caniuse data in this regard.

kof:
react-jss uses options.meta property if you pass it, otherwise it falls back to a component name https://github.com/cssinjs/react-jss/blob/master/src/index.js#L36

kof:
Why do you call them SomeethingRenderer anyways and then replace the Renderer part in order to have nicer meta attributes on style tags?

sapegin:
> Btw. WrappedComponent.name is not supported everywhere, haven't found caniuse data in this regard.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

> Why do you call them SomethingRenderer anyways and then replace the Renderer part in order to have nicer meta attributes on style tags?

Because I have components without `Renderer` but they don’t have any markup.

kof:
Btw. latest react-jss also passes classes object to the component additionally to the sheet.

kof:
Migration should be trivial.

sapegin:
@kof What about merging with custom styles and passing theme object?

kof:
> Because I have components without Renderer but they don’t have any markup.

So they are containers? Why is it important to identify a container by name? Why not call a container component  SomethingContainer?

sapegin:
Components without `Renderer` are “containers”:

```
function Button(props) {
  const stuff = do something with props
  return <ButtonRenderer {...stuff} />;
}
```

`Renderer`s contains only HTML so users could easily change style guide rendering.

sapegin:
And of course I don’t want to write `ButtonContainer` instead of `Button` everywhere ;-)

kof:
> @kof What about merging with custom styles and passing theme object?

Why are u using context to pass the config and take the user styles overrides if they are basically global?  Why not just merge them in each styles definition?

sapegin:
Because I don’t want users to do that when they override our components.

okonet:
@sapegin @kof let's move off-topic discussions to separate issues since they will be lost as soon we merge it.

kof:
Ok should we make this whole styles discussion a separate thing? I see 2 possibilities: we add SheetsRegistry right now here. Otherwise I need to implement a ThemeProvider in react-jss. It is planned anyways. Right now the first one is faster.

okonet:
Yes, let's move the discussion and implement the fastest way and refactor later.

sapegin:
Then let’s make fast now and then we’ll se if we can switch to react-jss :shipit:

kof:
So should I do it or it is already clear enough?

okonet:
Please do this @kof! I've tons of other things to do now.

kof:
Should it be on top of this branch?

okonet:
I'd say so since it belongs together.

sapegin:
@kof Yup, you should have commit right now. Welcome to the club but don’t talk about it ;-)

kof:
Ok I have added the code for sheets aggregation, but I couldn't test it, somebody need to show me how to kick this all off, or maybe it is right now in a broken state?

okonet:
@kof try `yarn run build`

kof:
Ok, I have finished the SSR integration for JSS. I needed to do 2 suboptimal things, because of "react-html-document"  component limitations.

1. We need to get the styles used by the application before we render the document. That module assumes we don't. So I needed to render the app to string first and then pass it to HtmlDocument inside of a div with dangerouslySetInnerHTML.
2. HtmlDocument abstraction doesn't give any option to define an attribute on server-side rendered style tag, in order to delete that tag once js kicks in I currently remove the first tag, but it will break if you add more style tags on the server and they will come first.

okonet:
Hmm, that makes me think if we still should use HtmlDocument. Right now it looks as hack. Are there any better options to do that @kof?

kof:
Yes, simple template string like [here](https://github.com/cssinjs/cssinjs/blob/master/src/server.js#L47)

kof:
Or optionally own document renderer component without that dependency.

okonet:
How own renderer would help us here?

kof:
> How own renderer would help us here?

- It would specify an id for the ssr style tag.
- It would accept content as an html string and incapsulate  `dangerouslySetInnerHTML`

kof:
Optionally this can be done by sending a pr to that original package and make this possible.

okonet:
@kof do you mind updating the PR by removing the `react-html-document` and introducing the template string? I'm away for a few days so won't be able to work a lot. 🏂

kof:
I think I can make it happen

sapegin:
@kof Awesome!

okonet:
@kof any news on this one?

kof:
@okonet not yet, could find time right now, because on the trip

okonet:
@sapegin NOOOO! Why did you close it?

sapegin:
I’ve removed the `next` branch because it was merged, so it wasn’t me ;-| Is it possible to change the target branch without resending a PR?

okonet:
It seems I can't even re-open the PR. We'll need to rebase it on master and make a new PR I guess.

okonet:
@kof do you think you could find some time to get this done any soon?

sapegin:
We probably need to create another branch first because this one is probably from 4.x and it would be a nightmare to merge master to it.

kof:
Will be hard, I have a great  deal of issues on jss project.

FDiskas:
Can by closed probably.

PR #337, opened 2017-02-26
Feat: Hide errors while typing in code editor

UsulPro:
Hi! First of all, thank you for a great project. I used storybook for my projects, and now I really like styleguidist and I'd like to spend some time for working on using these two projects together

Regarding this PR, I noticed that while typing in the code editor some errors can appear in the component preview. It could be a bit annoying in case if you just haven't a bracket closed for example. So this PR will change it to show an error only if you remove focus from the editor and if you still in editor only red marker will appear (it can be customized via css). If there was an error before you focus the code editor, it behaves as usual

[screenview](http://recordit.co/tHF1tlv7nj)

### DIFF
--- a/src/rsg-components/Editor/Editor.css
+++ b/src/rsg-components/Editor/Editor.css
@@ -1,14 +1,31 @@
 /* Tweak CodeMirror styles */
-.root :global .CodeMirror {
+.root :global .CodeMirror,
+.root_err :global .CodeMirror,
+.root_edit :global .CodeMirror {
 	height: auto;
 	padding: 5px 12px;
 	font-size: 12px;
 }
-.root :global .CodeMirror-scroll {
+.root_err {
+	border-left: red 4px solid;
+}
+.root_edit {
+	border-left: #1978c8 4px solid;
+}
+.root {
+	border-left: rgb(245, 245, 245) 4px solid;
+}
+.root :global .CodeMirror-scroll,
+.root_err :global .CodeMirror-scroll,
+.root_edit :global .CodeMirror-scroll
+ {
 	height: auto;
 	overflow-y: hidden;
 	overflow-x: auto;
 }
-.root :global .cm-error {
+.root :global .cm-error,
+.root_err :global .cm-error,
+.root_edit :global .cm-error
+ {
 	background: none !important;
 }
--- a/src/rsg-components/Editor/Editor.js
+++ b/src/rsg-components/Editor/Editor.js
@@ -23,6 +23,9 @@ export default class Editor extends Component {
 	static propTypes = {
 		code: PropTypes.string.isRequired,
 		onChange: PropTypes.func.isRequired,
+		onCodeTyping: PropTypes.func.isRequired,
+		isCodeValid: PropTypes.bool.isRequired,
+		isCodeTyping: PropTypes.bool.isRequired,
 	};
 	static contextTypes = {
 		config: PropTypes.object.isRequired,
@@ -33,8 +36,11 @@ export default class Editor extends Component {
 		this.handleChange = debounce(this.handleChange.bind(this), UPDATE_DELAY);
 	}
 
-	shouldComponentUpdate() {
-		return false;
+	shouldComponentUpdate(nextProps) {
+		return (
+			this.props.isCodeValid !== nextProps.isCodeValid ||
+			this.props.isCodeTyping !== nextProps.isCodeTyping
+		);
 	}
 
 	handleChange(newCode) {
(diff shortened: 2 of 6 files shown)


sapegin:
Hi Oleg! Thanks for the pull request!

I’m afraid such behavior will annoy more people than the current behavior ;-/ Personally I want to know if I have an error in the code and what / where is this error. I also think that showing an obsolete state in the preview and requiring users to check the color of the border isn’t very good idea.

You can increase `previewDelay` to show errors later. 

I’ll keep the PR open for a while — maybe I’m wrong and other contributors will say so ;-)

kof:
Why do you save your file with errors? In any case you need to fix that error and showing errors as fast as possible is a good thing.

okonet:
I agree on showing errors as soon as possible. Don't think this should be over complicated: if there is an error in the code then it should be displayed.

UsulPro:
Thank you for response, guys! I don't mind, I think it could be closed now 😄

tizmagik:
Hm... maybe just me, but I see value in this PR.

I do find it sort of jarring when editing code in the CodeMirror box it jumps around and shows a redbox because you haven't yet typed the closing bracket or whatever. I can certainly see the need to show the error when you're editing the actual source file and hitting save, but this is solving a different use case (iterating in the CodeMirror box). I also think it's kind of clever to show the error on blur -- maybe we can sow the error after some onChange throttle in the CodeMirror box in addition to the on blur?

This seems useful... am I missing something?

sapegin:
@tizmagik That’s exactly how it supposed to work: [debounce changes in textarea](https://github.com/styleguidist/react-styleguidist/blob/d0646ae50d9a33cfcfe3acb46861f4f303d41a6e/src/rsg-components/Playground/Playground.js#L62-L66).

tizmagik:
@sapegin right! I can see how that sort of does what's here but it's nice to be able to see valid changes shown immediately and only delay/move errors so you don't lose your place on the page. 

But maybe it's just me and @UsulPro 😅🤷‍♂️

kof:
We lack the right solution here, its obviously not removing fast errors, its making the errors reading experience better without jumps.

sapegin:
@tizmagik That might work too though may unnecessary complicate the code ;-)

@kof I agree that current solution is not perfect. Do you have any ideas how to improve it?

kof:
Mb separate view for errors, like a sidebar console

tizmagik:
Maybe a compromise here would be to retain the height of the example preview on errors (and set it again when there's valid code)?

That would nicely solve the pain point I have with the page jumping on errors as I'm typing. Thoughts?

kof:
Btw why not to rethrow errors into the dev tools console?

sapegin:
@kof That would hide them from the user ;-)

Maybe @okonet has some ideas.

kof:
Or render an overlay above the component in its previous working state and write the error there, so that there is no jumps and additionally throw it into the dev tools console so that we see the stack trace which is currently not available and I already had some pain with this.

sapegin:
That’s closer, yeah ;-) The only issue is when error is much taller than the rendered component.

kof:
> That’s closer, yeah ;-) The only issue is when error is much taller than the rendered component.

scroll it, set min height

tizmagik:
Could be scrollable

sapegin:
In case with no previous render or some tiny component that won’t work. `min-height` would cause the same jump we’re trying to avoid.

kof:
Btw. the overlay can be bigger then the orgiinal component, it can start at the same top position and go down, similar to how browser implement alert boxes, it can be really big. Just set the max height and make it scrollbar.

kof:
By min-height I ment a min-height that is always defined, not just when an error comes, but with my previous suggestion you don't need a min-height.

sapegin:
So it would overlap the code? But I’m also thinking about something like that.

kof:
> So it would overlap the code? But I’m also thinking about something like that.

yes

UsulPro:
Maybe [demo page](https://usulpro.github.io/react-styleguidist/) to play with this feature could help?

to clarify: 
it keeps the code only if an error appears after you focus CodeMirror box and while you're in the box.
if error was in the source code it shows immediately
if you blur CodeMirror box it shows immediately
if you blur CodeMirror box and then returned it still shows error

Since it could be easily enabled/disabled in `PlaygroundRenderer` and `EditorRenderer` maybe it worth to add some keyboard shortcut for it?

about notifiction via border color I was just looking for the most minimalistic way to don't destroy the existing design. of course this maybe not the final decision

sapegin:
I‘m closing this for now. We definitely can and should improve the error handling in the code editor but probably this is not the right direction for Styleguidist. Thank you for your work anyway!

PR #342, opened 2017-03-02
Move navigation ids from anchors to headings

### LINKED ISSUE #339: Hiding heading anchors breaks sidebar navigation
I don't like the way new heading anchors look like so I have hidden them by modifying `styles` in the config:

```
    Heading: {
      anchor: {
        display: 'none'
      }
    }
```

and now the sidebar navigation is broken (because it expects the anchors to be present and visible on the page.

/cc @okonet

n1313:
Fixes #339


sapegin:
Cool, thanks!

PR #343, opened 2017-03-02
Hide function sources for PropType.func

### LINKED ISSUE #197: Do not show function sources in props list
If default prop value is a function it renders as its source which is not very nice especially if the function is long.

Simple solution: just write “Function” instead of the code.

Perfect solution: show highlighted or at least formatted code in a tooltip.

image 2016-10-12 at 10 33 31 am

n1313:
Fixes #197

screen shot 2017-03-03 at 00 48 32


sapegin:
Awesome! Thanks!

PR #344, opened 2017-03-02
Replace static CSS class with a data attribute on Preview

### LINKED ISSUE #334: Define a separate attribute for accessing user rendered components
Right now class names are misused for that purpose.  The reasoning is - classes used internally for styling should be a private implementation detail of sg, they shouldn't be used at the same time for accessing containers for [visual testing](https://github.com/styleguidist/react-styleguidist/issues/238). It makes a false assumption for a user that he should override styles using class names.  It is better to use the default generateClassName function which generates unique class names, to make the intent clear.

In order to give a user an access to the container through the dom, we can use for e.g. a 'data-container' attribute with a specified container name or even path.

n1313:
Fixes #334

### DIFF
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -1,5 +1,4 @@
 import React, { PropTypes } from 'react';
-import cx from 'classnames';
 import Editor from 'rsg-components/Editor';
 import Link from 'rsg-components/Link';
 import Preview from 'rsg-components/Preview';
@@ -75,7 +74,7 @@ export function PlaygroundRenderer({
 }) {
 	return (
 		<div className={classes.root}>
-			<div className={cx(classes.preview, 'rsg--example-preview')}>
+			<div className={classes.preview} data-preview={name}>
 				<div className={classes.isolatedLink}>
 					{name && (
 						isolatedExample ? (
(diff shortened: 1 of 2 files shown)


src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -75,7 +74,7 @@ export function PlaygroundRenderer({
 }) {
 	return (
 		<div className={classes.root}>
-			<div className={cx(classes.preview, 'rsg--example-preview')}>
+			<div className={classes.preview} data-preview={name}>
 				<div className={classes.isolatedLink}>
 					{name && (
```
sapegin: Looks like `name` can be empty but I’m not sure why. Maybe we can remove this condition ;-)

n1313: Maybe we can, but that's out of scope of this PR :)

sapegin: Then you need to handle case when name is not there ;-)

n1313: Ah, now I see what you are talking about. OK!

n1313:
If `name` is not defined then data-preview attribute will be present but empty.

sapegin:
Thanks!

PR #353, opened 2017-03-14
Allow ignoring props

### LINKED ISSUE #331: Are private (or protected, or hidden) props possible?
In some cases a component might have a prop that should not be accessed by the user. It could be useful to hide this prop from the documentation or mark it in some other way.

An example of this would be a container component that is wrapped in Redux `connect()` with props that are managed by react-redux. I have a number of props that must be declared on the component for action dispatch to work and should not be tampered with by a user.

n1313:
This PR allows marking props with `@ignore` JSDoc tag to make RSG ignore them when rendering the documentation. Fixes #331

### DIFF
--- a/loaders/utils/getProps.js
+++ b/loaders/utils/getProps.js
@@ -8,6 +8,14 @@ const requireIt = require('./requireIt');
 
 const examplesLoader = path.resolve(__dirname, '../examples-loader.js');
 
+// HACK: We have to make sure that doclets is a proper object with correct prototype to
+// work around an issue in react-docgen that breaks the build if a component has JSDoc tags
+// like @see in its description, see https://github.com/reactjs/react-docgen/issues/155
+// and https://github.com/styleguidist/react-styleguidist/issues/298
+const getDocletsObject = (string) => {
+	return Object.assign({}, reactDocs.utils.docblock.getDoclets(string));
+};
+
 /**
  * 1. Remove non-public methods.
  * 2. Extract doclets.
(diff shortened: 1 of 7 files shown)


src/rsg-components/Props/__snapshots__/Props.spec.js.snap
```diff
@@ -1,5 +1,56 @@
...
+      <td
+        className="undefined undefined"
+      >
+        <Group
+          separator={<br />}
+        >
+          <_class
+            inline={true}
+            text="Label
+          @ignore"
```
sapegin: Looks like it renders it ;-)

loaders/utils/getProps.js
```diff
@@ -43,5 +45,17 @@ module.exports = function getProps(doc) {
...
 	}
 
+	if (doc.props) {
+		// Read doclets of props
+		Object.keys(doc.props).forEach(propName => {
+			const prop = doc.props[propName];
+			const doclets = getDocletsObject(prop.description);
+			// Remove ignored props
+			if (doclets && doclets.ignore) {
+				delete doc.props[propName];
```
sapegin: Looks like something is wrong with the code or the test — this line is not covered.

src/rsg-components/Props/__snapshots__/Props.spec.js.snap
n1313: Hmmm, that's weird.

loaders/utils/getProps.js
n1313: I was testing the wrong thing :(

n1313:
Updated the PR, please take a look.

sapegin:
Cool, thanks!

PR #372, opened 2017-04-04
Hash JS assets in production build

### LINKED ISSUE #371: Hash JS assets in production
When deploying a styleguide to production, the JS assets are not hashed / fingerprinted. When using a CDN, this means the assets are difficult to update since you have to manually expire the assets.

Relevant lines: https://github.com/styleguidist/react-styleguidist/blob/master/scripts/make-webpack-config.js#L27-L30

Below the above lines, there is a section for `isProd`. Adding a hash to the filename and chunk filenames should fix this issue. For reference, this is how create-react-app handles it: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpack.config.prod.js#L70-L71

If this is something you're interested in adding to the library, I'd be happy to create a PR for it 😄

green-arrow:
Fixes #371

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "babel-preset-env": "^1.3.2",
     "babel-preset-react": "^6.23.0",
     "babel-register": "^6.24.0",
+    "clean-webpack-plugin": "^0.1.16",
     "dog-names": "^1.0.2",
     "enzyme": "^2.8.0",
     "enzyme-to-json": "^1.5.0",
--- a/scripts/__tests__/create-server.spec.js
+++ b/scripts/__tests__/create-server.spec.js
@@ -20,7 +20,8 @@ test('createServer should return an object containing a production Webpack compi
 	const result = createServer(config, 'production');
 	expect(result).toBeTruthy();
 	expect(result.compiler).toBeTruthy();
-	expect(result.compiler.options.output.filename).toBe('build/bundle.js');
+	expect(result.compiler.options.output.filename).toBe('build/bundle.[chunkhash:8].js');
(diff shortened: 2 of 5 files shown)


sapegin:
Thanks for the PR!

One note: we probably should remove old files otherwise `bundle` folder will grow.

green-arrow:
@sapegin - should that be part of this PR? I can always bring in [rimraf](https://github.com/isaacs/rimraf) and just remove the styleguide directory before the build starts.

green-arrow:
@sapegin - Just FYI, I have a commit locally that uses `rimraf` to remove the `styleguide/build` directory before every build. If you'd like that to be in this PR, just let me know and I'll go ahead and push it.

sapegin:
Probably better to use [clean-webpack-plugin](https://www.npmjs.com/package/clean-webpack-plugin) to keep everything in webpack config.

green-arrow:
@sapegin - alrighty. Should that be a separate PR or included here?

sapegin:
Up to you, I’m fine either way ;-)

green-arrow:
@sapegin - I'll just pop it in this commit then 😄 

Give me a couple minutes and I'll push it out.

green-arrow:
@sapegin @okonet - Changes for cleaning up the `build` directory are pushed. Let me know if you see anything that needs changed / added.

sapegin:
Awesome, thanks!

sapegin:
I’m making the same mistake for the second time — missing that a new dependency was added to the wrong group (`dependencies` / `devDependencies`) because GitHub doesn’t show enough code in the diff to spot it 🤔

New path release should fix that very soon.

PR #374, opened 2017-04-05
Change default serverHost to 0.0.0.0

### LINKED ISSUE #373: Unable to access styleguide outside of local machine
In `scripts/schema/config.js`, the default `serverHost` is set to `localhost`. When using this host with express, it means that only the local machine can access the site. If users want to connect from another device, like a phone or a table, to test the style guide, they are unable to unless setting the `serverHost` manually to `0.0.0.0`.

Is this a case where the default should be changed to `0.0.0.0` to accommodate this use case, or should it just be left up to the user?

green-arrow:
This allows users to connect from outside their local machines

Resolves #373

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -242,7 +242,7 @@ See examples in the [Sections section](Components.md#sections).
 
 #### `serverHost`
 
-Type: `String`, default: `localhost`
+Type: `String`, default: `0.0.0.0`
 
 Dev server host name.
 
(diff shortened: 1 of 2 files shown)


sapegin:
Could we consider this a minor or patch change? Technically `localhost` still works.

/cc @okonet @n1313

green-arrow:
I would say it could be a minor update as it shouldn't break existing setups.

n1313:
I have no opinion but I welcome this change because this is exactly how my setup works :)

Issue #354, opened 2017-03-15
[discussion] Extensions API


okonet:
We might want to take a look at https://github.com/camwest/react-slot-fill for allowing extensions for SG.

sapegin:
We were talking about Storybook like plugins that would add tabs under examples. But this also can be useful.

okonet:
Yeah, I actually don't like how storybook does it since it's very restrictive. This one should theoretically allow writing custom visualizers for the components and create custom controlled components (like knobs?). We should try it out and see if it works well for us.

sapegin:
I think for knobs Storybook-style is quite good and enough. Why do you think it won’t work well?

okonet:
My concern is more UI-wise: storybook only allow you to extend by creating tabs _around_ the content there we are interested in creating extension _inside_ of the content.

sapegin:
Could you make some mockup of your ideas? That would really help ;-)

okonet:
I could when I get some more time :(

sapegin:
Have watched a [ReactConf talk](https://youtu.be/395ou6k6C6k?list=PLb0IAmt7-GS3fZ46IGFirdqKTIxlws7e0) about it and this is amazing! But looks like it only works with Fiber. Not sure they will / can support React 15.

sapegin:
What do we want to extend by the way?

1. Tabs under the component playground (Storybook like) seems useful: non-editable code, snapshots, knobs…
2. Access to style guide config.

What else?

MicheleBertoli:
+1 for tabs, it seems a great starting point to me.

cef62:
I agree with @okonet, I would love have both possibilities: 

- add isolated features/plugins in tabs (around content)
- extend/augment rendered content (inside content)

While working on `snapguidist` I had some idea on that. Now that I've closed a very important deadline at work I can finally work on it. Next days I will share more about this with you!

sapegin:
@cef62 Ping ;-) Do you have some time now?

cef62:
I've started digging through the new 5.x implementation. I really like the new  AST based parsing, great job @sapegin!
At the moment the part bugging me the most is that there's no effective way to assign a unique ID to every parsed snippet. This is something very important to support a plugin API in my opinion, in `snapguidist` for example we could use it to avoid unnecessary execution of unchanged snippets.

[...]

PR #379, opened 2017-04-08
Update PlaygroundRenderer.js

kulakowka:
I want to suggest a change in the styles of the playground:

### DIFF
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -6,6 +6,10 @@ import Styled from 'rsg-components/Styled';
 
 const styles = ({ font, link, linkHover, border, baseBackground, codeBackground }) => ({
 	root: {
+		backgroundColor: 'white',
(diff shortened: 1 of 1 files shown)


okonet:
I like the idea! But I think this should be customizable somehow. We might want to discuss features like this in https://github.com/styleguidist/react-styleguidist/issues/354

kulakowka:
@okonet It can be [customized](https://github.com/styleguidist/react-styleguidist/blob/next/docs/Cookbook.md#how-to-change-styles-of-a-style-guide).

n1313:
I have the same kind of background in my project as well, but it's diagonally striped. This one is also good.

n1313:
Although to be honest, I am not sure if this is going to be useful for **most** people. This feature is useful if you are working on some sort of a reusable UI component library which is supposed to work in unpredictable conditions on variety of backgrounds. I doubt this is what the majority of RSG users are building. I think the majority is targeting one environment with a specific background color, and so will have little use for this change, and even might be annoyed by it. I think this change could be a perfect example of how to customise your RSG for the cookbook, or an optional extension in the future, but I'm not 100% sold on it being in the main repo.

sapegin:
Thanks for the pull request. I don’t think it’d be a good default. I agree with @n1313, would you mind sending a new pull request with a cookbook example?

PR #381, opened 2017-04-10
Use `[name]` in output filename

jskeate:
If you try to build a project with multiple entries without this, you get an error about multiple entries being written to `build/bundle.js`.

While there may be something to be said about not having multiple entrypoints in a dev build, this should have zero impact on any project without multiple entries. And since `output` can't be changed with the config's `webpackConfig` option, the only other way is to use `dangerouslyUpdateWebpackConfig`, which I'd rather avoid.

### DIFF
--- a/scripts/__tests__/create-server.spec.js
+++ b/scripts/__tests__/create-server.spec.js
@@ -30,6 +30,6 @@ test('createServer should return an object containing a development Webpack comp
 	const result = createServer(config, 'development');
 	expect(result).toBeTruthy();
 	expect(result.compiler).toBeTruthy();
-	expect(result.compiler.options.output.filename).toBe('build/bundle.js');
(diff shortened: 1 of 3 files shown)


n1313:
Uh, tests?

jskeate:
Okay, so I updated the tests. I have no idea what the Travis failures are about. I must have forked this off of something other than `master`, but then trying to rebase it onto that I find a lot has changed (jest vs ava, for example). Recommendation?

sapegin:
Sorry for the delay, I was traveling ;-) Do tests pass locally?

jskeate:
Okay, seems to be all cleared up. Not quite sure what happened -- I blame the GitHub editor.

sapegin:
Awesome, thanks!

PR #386, opened 2017-04-11
Update Cookbook.md

kulakowka:
Update Playground background in styling example #379

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -179,6 +179,15 @@ module.exports = {
       '@keyframes blink': {
         to: { opacity: 0 }
       }
+    },
+    Playground: {
+      root: {
+        backgroundColor: 'white',
+        backgroundImage: 'linear-gradient(45deg,#efefef 25%,transparent 0,transparent 75%,#efefef 0,#efefef),linear-gradient(45deg,#efefef 25%,transparent 0,transparent 75%,#efefef 0,#efefef)',
+        backgroundPosition: '0 0,10px 10px',
+        backgroundSize: '21px 21px',
+        fontSize: '1rem'
+      }
     }
   }
 };


docs/Cookbook.md
```diff
@@ -179,6 +179,15 @@ module.exports = {
       '@keyframes blink': {
         to: { opacity: 0 }
       }
+    },
+    Playground: {
```
sapegin: A comment here would be very useful ;-)

docs/Cookbook.md
```diff
@@ -179,6 +179,15 @@ module.exports = {
...
+      root: {
+        backgroundColor: 'white',
+        backgroundImage: 'linear-gradient(45deg,#efefef 25%,transparent 0,transparent 75%,#efefef 0,#efefef),linear-gradient(45deg,#efefef 25%,transparent 0,transparent 75%,#efefef 0,#efefef)',
+        backgroundPosition: '0 0,10px 10px',
```
sapegin: This doesn’t look like a correct value.

kulakowka: Why? 

`0 0` - it's background position for first gradient
`10px 10px` - it's background position for second gradient

I have defined the styles in my project and it works for me.

docs/Cookbook.md
kulakowka: What should this comment be about?
Something like that:
```
/* For example: You can change the background in the component playground. */
```

docs/Cookbook.md
sapegin: I see, my mistake!

docs/Cookbook.md
sapegin: Yeah, something like that. And also what kind of background because it‘s impossible to see that it’s a background to test transparent components from the code ;-)

kulakowka:
In Google Chrome I got a strange bug:

PR #388, opened 2017-04-12
Add prop-types package

### LINKED ISSUE #380: Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
I updated the react to version 15.5.0 and saw an error in the console:

```
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
```

We need to update components to use [prop-types](https://www.npmjs.com/package/prop-types) package. More info in official blog: https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

```javascript
// Before (15.4 and below)
import React from 'react';

class Component extends React.Component {
  render() {
    return <div>{this.props.text}</div>;
  }
}

Component.propTypes = {
  text: React.PropTypes.string.isRequired,
}

// After (15.5)
import React from 'react';

class Component extends React.Component {
  render() {
    return <div>{this.props.text}</div>;
  }
}

Component.propTypes = {
  text: PropTypes.string.isRequired,
};```

kulakowka:
It fixes #380 

Should we update the version of react in peerDependencies?

### DIFF
--- a/examples/basic/lib/components/Button/Button.js
+++ b/examples/basic/lib/components/Button/Button.js
@@ -1,4 +1,5 @@
-import React, { PropTypes } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
 
 import s from './Button.css';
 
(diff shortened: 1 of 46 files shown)


sapegin:
> Should we update the version of react in peerDependencies?

Probably to 15+.

I don’t know what’s wrong again with conflicts ;-/ Maybe your fork’s master is behind?

sapegin:
Closing in favor of more complete #389. Thanks for your effort!

PR #392, opened 2017-04-12
Add buble configuration option

DylanMoz:
Allows for configuration of buble. Needed this for my use case so that my code examples could have tagged template literals. In buble the config is `transforms.dangerousTaggedTemplateString`


sapegin:
Thanks!

PR #389, opened 2017-04-12
React 15.5 & Prop type updates

ldabiralai:
This updates to using `prop-types` package and changes react versions to 15.5

PR #394, opened 2017-04-13
Section isolated mode

Knorcedger:
This is the implementation of the section isolated mode.

I was not able to write tests for it because I couldn't run the tests. I think it's something with `react-addons-test-utils` moving to `react-dom/test-utils`. Is this fixed in the master branch, or can someone take a look so that I can run the tests and add tests for this too?


sapegin:
Thanks!

I’m trying to merge #389 but it still doesn’t work as it should and I’m not sure why ;-|

src/rsg-components/Section/SectionRenderer.js
```diff
@@ -1,25 +1,53 @@
...
 	return (
 		<section className={classes.root}>
-			{name && (
-				<Heading level={1} slug={slug} className={classes.heading}>{name}</Heading>
-			)}
+			<div className={classes.titleWrapper}>
+				{name && (
+					<Heading level={1} slug={slug} className={classes.heading}>{name}</Heading>
+				)}
+				<div className={classes.isolatedLink}>
```
sapegin: Looks like it’s time to extract it to a separate component. We already have this pattern in two places I think.

src/utils/utils.js
```diff
@@ -148,22 +148,26 @@ export function filterComponentsInSectionsByExactName(sections, name) {
 	return components;
 }
 
+export function filterSections(sections, name) {
```
sapegin: Please add JDoc here.

Knorcedger:
Cool, I will wait for the other PR to be merged first and then add the tests. In the mean time will fix the Changes Requested :)

sapegin:
Awesome! 🍕

sapegin:
Tests in master should work now.

src/rsg-components/Section/SectionRenderer.js
Knorcedger: I created an IsolatedLink component that contains the markup and the CSS for it. I believe that the CSS rule that says "on hover of this element, reveal the IsolatedLink" should be inside the component (the component would just need the className for the element that will reveal the component once that element is hovered).

Do you agree with this approach?

Trying to implement this, I run into a problem with how to use this class inside the styles object. Following is the IsolateLinkRender where hoverClass is the class I'm getting from the parent component, and trying to use it in the styles object. Obviously the following doesn't work, since I don't know how to pass the hoverLink from the component to the styles object. Any help?
```javascript
const styles = ({ font }) => ({
	[hoverClass]: {
		'&:hover $isolatedLink': {
			isolate: false,
			opacity: 1,
		},
	},
	isolatedLink: {
		position: 'absolute',
		top: 0,
		right: 0,
		fontFamily: font,
		fontSize: 14,
		opacity: 0,
		transition: 'opacity ease-in-out .15s .2s',
	},
});

export function IsolatedLinkRenderer({ classes, hoverClass, isIsolated, link }) {
	return (
		<div className={classes.isolatedLink}>
			{link && (
				isIsolated ? (
					<Link href="/">⇽ Back</Link>
				) : (
					<Link href={'#!/' + link}>Open isolated ⇢</Link>
				)
			)}
		</div>
	);
}

IsolatedLinkRenderer.propTypes = {
	classes: PropTypes.object.isRequired,
	hoverClass: PropTypes.string.isRequired,
	isIsolated: PropTypes.bool,
	link: PropTypes.string,
};

export default Styled(styles)(IsolatedLinkRenderer);
```

sapegin: > I believe that the CSS rule that says "on hover of this element, reveal the IsolatedLink" should be inside the component

With this I agree.

> the component would just need the className for the element that will reveal the component once that element is hovered

I think you need to put both of them into this component. Sou it’d be not `IsolatedLink` but more like `IsolatedContainer`:

```html
<IsolatedContainer to="/foo">
  this content will have a link in the corner
</IsolatedContainer>
```

Knorcedger: @sapegin I created the IsolatedContainer component. Since the CSS rule for "onHover" is inside the component, that means that on heading hover show the "Open Isolated" link (it used to be onHover of the whole component info area show the link). That seems fine by me, but theres a problem with the Isolated example which has no title. It used to be on hover of the code example area but now the only "hover" area is the link itself which is hidden.

Any ideas? Do you want me to update the PR to see the code?

sapegin: Yes please!

Knorcedger:
@sapegin updated the PR. Let me know if you want me to clarify anything. Tests are still missing

sapegin:
That’s what I’d do: https://gist.github.com/9d02a2603619daa8de7f1287490c894f

Knorcedger:
I tried your proposed solution but that doesn't solve the problem that to open an isolated example you have to hover over the "opacity: 0" link.

sapegin:
You never hover on links, only on containers.

sapegin:
But probably you can hardcode these links back because of #426 ;-)

sapegin:
Works fine for me: 

image 2017-05-03 at 6 37 53 pm

But please take a look at #426, we’re probably trying to do something that would be removed soon. So if it‘s easier to revert, please revert it back, and sorry for inconsistent suggestions ;-)

Knorcedger:
Ok, cool. Shall I revert back to this [commit](https://github.com/styleguidist/react-styleguidist/pull/394/commits/de6dd2d819b6719b2314b920b992d247e2525a6c) and then add the tests?

sapegin:
Yep, thanks! 🍕

Knorcedger:
@sapegin I reverted to the commit you suggested and tried to adjust the tests, but I'm not familiar with Jest and it's snapshots so I'm not sure if it's 100% ok. At least `npm test` was reporting no errors locally :)

src/rsg-components/Section/__snapshots__/Section.spec.js.snap
```diff
@@ -1,15 +1,30 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`render should not render title if name is not set 1`] = `<section />`;
+exports[`render should not render title if name is not set 1`] = `
+<section>
```
sapegin: These empty divs look strange but they most likely will go away after UI 2.0 ;-)

src/utils/utils.js
```diff
@@ -149,21 +149,31 @@ export function filterComponentsInSectionsByExactName(sections, name) {
 }
 
 /**
- * Returns an object containing component name and, optionally, an example index
+ * Filters the sections to find the one with the matching name
+ * @param  {Array}  sections The styleguide sections
+ * @param  {string} name     The name to match
+ * @return {object}          The section found
+ */
+export function filterSections(sections, name) {
```
sapegin: Could you add a test for this function too? :-)

src/rsg-components/Section/SectionRenderer.js
```diff
@@ -2,24 +2,52 @@ import React from 'react';
...
+					{name && (
+						isolatedSection ? (
+							<Link href="/">⇽ Back</Link>
```
sapegin: And a test for back link.

sapegin:
Looks fine! There are two pieces uncovered by any test though.

src/utils/utils.js
Knorcedger: I added tests for this function.

src/rsg-components/Section/SectionRenderer.js
Knorcedger: I tried to understand how to add tests for links, but I couldn't figure it out. Do you mind adding it?

sapegin: That’s fine, we’re working on a new UI for these links anyway ;-)

Issue #393, opened 2017-04-13
Support @param in props description


ankri:
I was looking for a solution to have `@param` support in my props description, not only in public methods. My use case is that I want to document the parameters of my `PropType.func` props so the user of my component can see the arguments I will pass along to the function he provides.

Sadly react-docgen does not apply any specific treatment to the docblock of propTypes. [react-docgen#141](https://github.com/reactjs/react-docgen/issues/141). That's why you have to write your own parser for that.

I have a solution [here](https://github.com/ankri/react-styleguidist/tree/props-with-tags/examples/customised)

I added a custom [propsParser](https://github.com/ankri/react-styleguidist/blob/props-with-tags/examples/customised/styleguide.config.js#L49) and [PropsRenderer](https://github.com/ankri/react-styleguidist/blob/props-with-tags/examples/customised/styleguide/components/PropsRenderer.js#L186) to the `customised` example.

Code:
```javascript
/**
 * Gets called when the user clicks on the button
 *
 * @param { SyntheticEvent } event The react synthetic event
 * @param { Array } allProps All the props of Button
 */
onClick: PropTypes.func,
```
Output:
![image](https://cloud.githubusercontent.com/assets/2842920/24984062/6d3b624c-1feb-11e7-92e0-74d13f35dac3.png)

It should be trivial to add other tags like `@return`, `@see` or `@deprecated`.

---

I want to give back to this awesome project and I'm sure some people would benefit from this functionality. What should I do? Open a PullRequest? Create a new example? Add the code snippets to the `Cookbook.md`?

I would love to see `@param` support in the default version of the styleguidist but you would need the custom `propParser` for that and I do not know whether replacing the propsParser is the right way. You could also run `doctrine` on the `props.description` inside the `PropsRenderer` but that would be kind of ugly in my opinion. What do you guys think?

sapegin:
I think it’d be a good feature! We already do some description parsing [here](https://github.com/styleguidist/react-styleguidist/blob/master/loaders/utils/getProps.js#L37) so I’d start there ;-)

okonet:
@sapegin that's a lot of copy-paste. Should we include it by default?

ankri:
Thanks to @sapegin's tip I am working on a PR right now.

So far I have implemented support for
- [X] `@param`
- [X] `@deprecated`
- [ ] `@see`

Do you have suggestions for other [tags](http://usejsdoc.org/)?

sapegin:
[...]

PR #395, opened 2017-04-13
[WIP] JsDoc Tag support for Props, Components and Methods

ankri:
Adds the JsDoc Tag support mentioned in #393

Supports:
- [X] `@param`, `@arg`, `@argument` for props
- [X] `@author` for component
- [X] `@deprecated` for props, components and methods
- [X] `@see`, `@link` for props, components and methods
- [X] `@since`, `@version` for props, components and methods
- See [JsDoc homepage](http://usejsdoc.org/) for a list of other JsDoc tags. Did I miss a tag?

Tests are currently missing but I wanted to publish the PR to get your feedback on the styling. 

I've added some JsDoc Tags to the `basic` example.

Todo:
- [X] Tests
- [X] Documentation in `Documenting.md`


sapegin:
Could you please make a few screenshots?

ankri:
Here is a [jsbin](https://output.jsbin.com/vopadunaxu) of the rendered html with styling. I only added tags to `Button` and `CounterButton`

ankri:
# Button
image

# CounterButton
image

# Another look at Button
```javascript
import React, { PropTypes } from 'react';

import './Button.css';

/**
 * The only true button.
 *
 * @version 1.0.1
 * @see Check the [PR](https://github.com/styleguidist/react-styleguidist/pull/395#issuecomment-293943502) for more info
 * @author [@sapegin](https://github.com/sapegin)
 * @author [@ankri](https://github.com/ankri)
 */
export default function Button({
	color,
	size,
	onClick,
	children,
}) {
	const styles = {
		color,
		fontSize: Button.sizes[size],
	};

	return (
		<button className="button" style={styles} onClick={onClick}>{children}</button>
	);
}
Button.propTypes = {
	/**
	 * Button label.
	 */
	children: PropTypes.string.isRequired,
	/**
	 * The color for the button
	 *
	 * @see Check [Wikipedia](https://en.wikipedia.org/wiki/Web_colors#HTML_color_names) for a list of color names
	 */
	color: PropTypes.string,
	/**
	 * The size of the Button
	 *
	 * @since Version 1.0.1
	 */
	size: PropTypes.oneOf(['small', 'normal', 'large']),

	/**
	 * The width of the button
	 *
	 * @deprecated Do not use! Use size instead!
	 */
	width: PropTypes.number,

	/**
	 * Gets called when the user clicks on the button
	 *
	 * @param { SyntheticEvent } event The react `SyntheticEvent`
	 * @param { Object } props All the props of Button
	 */
	onClick: PropTypes.func,

	/**
	 * A prop that should not be visible in the doc.
	 * @ignore
	 */
	ignoredProp: PropTypes.bool,
};
...
}
```
image

ankri:
# omitting type on `@param`
image

src/rsg-components/JsDoc/ArgumentsRenderer.js
```diff
@@ -0,0 +1,57 @@
...
+		fontSize: 'inherit',
+	},
+});
+
+export function JsDocArguments({ classes, tags }) {
+	function renderParams(params) {
+		return (
+			params.map((param, index) => {
+				return (
+					<div key={`${param.name}-${index}`}>
```
sapegin: Do you need an index here? You can’t have several params with the same name.

src/rsg-components/JsDoc/ArgumentsRenderer.js
```diff
@@ -0,0 +1,57 @@
...
+						<Code className={classes.name}>{param.name}</Code>
+						{param.type && ':'}
```
sapegin: I’d add a space after : and maybe before.

src/rsg-components/JsDoc/ArgumentsRenderer.js
```diff
@@ -0,0 +1,57 @@
...
+						{param.type && <Code className={classes.type}>{param.type.name}</Code>}
```
sapegin: You can you `react-group` here instead of manual concatenation, otherwise if type of description is missed you’ll have extra : or -.

src/rsg-components/JsDoc/ArgumentsRenderer.js
```diff
@@ -0,0 +1,57 @@
...
+						{' - '}<Markdown text={param.description} inline />
```
sapegin: And a proper — ;-)

src/rsg-components/JsDoc/ArgumentsRenderer.js
```diff
@@ -0,0 +1,57 @@
...
+					</div>
+				);
+			})
+		);
+	}
+
+	return (
+		<div className={classes.argumentsBlock}>
+			<div className={classes.argumentsHeader}>Arguments</div>
+			{ tags.param && renderParams(tags.param) }
```
sapegin: Is it possible that all three param, arg and argument are empty?

src/rsg-components/Methods/MethodsRenderer.js
```diff
@@ -57,18 +64,32 @@ export function MethodsRenderer({ classes, methods }) {
...
 		});
 		return rows;
 	}
 
+	function renderMethodName(name, tags) {
+		if ('deprecated' in tags) {
+			return (<Code className={classes.deprecatedName}>{name}()</Code>);
+		}
+
+		return (<Code className={classes.name}>{name}()</Code>);
```
sapegin: `cx(classes.name, {[classes.nameIsDeprecated]: 'deprecated' in tags})`

loaders/utils/getProps.js
```diff
@@ -16,6 +17,21 @@ const getDocletsObject = (string) => {
 	return Object.assign({}, reactDocs.utils.docblock.getDoclets(string));
 };
 
+const getTagsFromDoctrine = (documentation) => {
```
ankri: It is the same as `_.groupBy` thanks.

ankri:
> There are quite a lot of markup repetition in JSDoc components: most of them use similar pattern and could be simplified to just this:

```jsx
  <Markdown text={`Since: ${tags.since[0].description}`} />
```
---
This was done on purpose.

By using this approach:

```jsx
<div className={classes.sinceBlock}>
  <span className={classes.sinceHeader}>Since:</span>
  {' '}
  <Markdown text={tags.since[0].description} inline />
</div>
```
... everyone can change the styling of the headers like "Since" or "Version" like they want to. With your approach one would lose the styling options.

But I've got no problem with changing it like you suggested. What do you say?
- Leave styling options but have more repetition in the code
- Remove styling options and have a cleaner code

src/rsg-components/JsDoc/ArgumentsRenderer.js
ankri: Done:
```jsx
<Group key={`${param.name}`}>
  <Code className={classes.name}>{param.name}</Code>
  {param.type && ':'}
  {param.type && <Code className={classes.type}>{param.type.name}</Code>}
  {'—'}
  <Markdown text={param.description} inline />
</Group>
```
I still need `{param.type && ':'}` and {param.type && <Code` though so that the output is `param — description` when `type` is `undefined`.

src/rsg-components/JsDoc/ArgumentsRenderer.js
ankri: They could all be `undefined`. But that will be handled by the `PropsRenderer`
```jsx
{(prop.tags.arg || prop.tags.argument || prop.tags.param) && <JsDocArguments tags={prop.tags} />}
```

In case you would have a documentation like this:
```javascript
static propTypes = {
  /**
  * @param
  */
  onClick: PropTypes.func
};
```
The `param` tag would be `undefined`

ankri: But when using
```javascript
static propTypes = {
  /**
  * @param { SyntheticEvent } event
  */
  onClick: PropTypes.func,
};
```

The output would be `event : SyntheticEvent - null`

And for:
```javascript
static propTypes = {
  /**
  * @param { SyntheticEvent }
  */
  onClick: PropTypes.func,
};
```
the output would be `SyntheticEvent - null`

What do you think: Should I remove the `null`s or should the `nulls`remain as a reminder that the user has forgotten to document their arguments correctly?

I would prefer to leave the `null` as reminder.

src/rsg-components/Methods/MethodsRenderer.js
ankri: Elegant. Fixed and thanks for the tip.

sapegin:
I’d start from simpler code.

ankri:
Added tests and documentation. Also fixed your suggestions.

Ready to review now.

sapegin:
Cool, much better now! But something is wrong with the branch — there are some unrelated changes from master.

src/rsg-components/JsDoc/ArgumentsRenderer.js
sapegin: I’d not render `null`s: there are no required parts, and requiring documentation for everything would decrease its quality.

ankri:
Hm. I needed the changes from master for the tests to work with React 1.15.5. Can I do something?

src/rsg-components/JsDoc/ArgumentsRenderer.js
ankri: In the newest version there are no `null`s rendered.

sapegin:
How did you merge master?

ankri:
I rebased it

sapegin:
I’ll try to merge it after 5.1 release.

PR #404, opened 2017-04-24
Support .jsx by default

KadoBOT:
Some style guides, require you to change the extension of files containing JSX to .jsx. So, I think it would be good if `styleguide server` also ignore files named .spec.jsx and .test.jsx.

### DIFF
--- a/docs/Components.md
+++ b/docs/Components.md
@@ -2,7 +2,7 @@
 
 ## Components
 
-By default Styleguidist will search components using this [glob pattern](https://github.com/isaacs/node-glob#glob-primer): `src/components/**/*.js` and `src/components/**/*.jsx`. It will pick up paths like `src/components/Button.js`, `src/components/Button/Button.js` or `src/components/Button/index.js`. It will also ignore test files (`__tests__` folder and file names containing `.test.js`, `.test.jsx`, `.spec.js` and `.spec.jsx`). If it doesn’t work for you, create a `styleguide.config.js` file in your project’s root folder and configure the patterns to fit your project structure.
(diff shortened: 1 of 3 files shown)


sapegin:
Thanks for the pull request! Could you also update `DEFAULT_COMPONENTS_PATTERN` and the docs?

docs/Components.md
```diff
@@ -2,7 +2,7 @@
 
 ## Components
 
-By default Styleguidist will search components using this [glob pattern](https://github.com/isaacs/node-glob#glob-primer): `src/components/**/*.js`. It will pick up paths like `src/components/Button.js`, `src/components/Button/Button.js` or `src/components/Button/index.js`. It will also ignore test files (`__tests__` folder and file names containing `.test.js` or `.spec.js`). If it doesn’t work for you, create a `styleguide.config.js` file in your project’s root folder and configure the patterns to fit your project structure.
+By default Styleguidist will search components using this [glob pattern](https://github.com/isaacs/node-glob#glob-primer): `src/components/**/*.js`. It will pick up paths like `src/components/Button.js`, `src/components/Button/Button.js` or `src/components/Button/index.js`. It will also ignore test files (`__tests__` folder and file names containing `.test.js`, `.test.jsx`, `.spec.js` and `.spec.jsx`). If it doesn’t work for you, create a `styleguide.config.js` file in your project’s root folder and configure the patterns to fit your project structure.
```
sapegin: `src/components/**/*.js` should be updated as well.

sapegin: And here too: https://github.com/styleguidist/react-styleguidist/blob/master/docs/Configuration.md#components

sapegin: And here: https://github.com/styleguidist/react-styleguidist/blob/master/docs/Configuration.md#ignore

KadoBOT: Since I've just changed the `ignore`, should `src/components/**/*.js` be changed? Because it won't look for .jsx unless it is specified in the styleguide config, right?

sapegin: That‘s `DEFAULT_COMPONENTS_PATTERN`, so I‘m for full support of `.jsx`, not only in ignores.

sapegin:
And the last one ;-)

> default Styleguidist will search components using this [glob pattern](https://github.com/isaacs/node-glob#glob-primer): `src/components/**/*.js`

sapegin:
Yeah, it should be the same as `DEFAULT_COMPONENTS_PATTERN`.

PR #406, opened 2017-04-25
add max-scale for mobile safari support

stepancar:
> Not all mobile browsers handle orientation changes in the same way. For example, Mobile Safari often just zooms the page when changing from portrait to landscape, instead of laying out the page as it would if originally loaded in landscape. If web developers want their scale settings to remain consistent when switching orientations on the iPhone, they must add a maximum-scale value to prevent this zooming, which has the sometimes-unwanted side effect of preventing users from zooming in

https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag

### DIFF
--- a/scripts/templates/index.html
+++ b/scripts/templates/index.html
@@ -3,7 +3,7 @@
 	<head>
 		<meta charset="utf-8">
 		<title><%=htmlWebpackPlugin.options.title%></title>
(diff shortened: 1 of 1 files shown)


sapegin:
If it means that the use can’t zoom it the page then I can’t accept this ;-|

PR #415, opened 2017-04-27
Use jss-global to target CodeMirror styles while still allowing theming

timsnadden:
Here is a starting point for allowing the themed monospace font to be used by CodeMirror. Because CodeMirror css is global it used jss-global.

### DIFF
--- a/src/rsg-components/Editor/EditorLoaderRenderer.js
+++ b/src/rsg-components/Editor/EditorLoaderRenderer.js
@@ -10,23 +10,21 @@ const styles = ({ font, monospace, light, codeBackground }) => ({
 		color: light,
 		backgroundColor: codeBackground,
 	},
-	// Tweak CodeMirror styles
+	// Tweak CodeMirror styles. Duplicate selectors increases specificity
 	'@global': {
-		body: {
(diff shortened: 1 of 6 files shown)


src/rsg-components/Editor/EditorLoaderRenderer.js
```diff
@@ -2,14 +2,33 @@ import React from 'react';
...
 	root: {
 		padding: [[7, 16, 10, 7]],
 		fontFamily: font,
 		fontSize: 12,
 		color: light,
 		backgroundColor: codeBackground,
 	},
+	// Tweak CodeMirror styles
+	'@global': {
+		body: {
```
sapegin: Can we use `.CodeMirror` instead of `body` + `& .CodeMirror`?

timsnadden: I've updated to use duplicate selectors (`.CodeMirror.CodeMirror`) to increase specificity enough to ensure overriding CodeMirror styles with any source order.

sapegin: At least it’s clear now ;-)

PR #421, opened 2017-04-30
added share code feature

stepancar:
Hello @sapegin!
Our team need in share code feature for sharing examples, bug stands for another developers.
I thing that this feature useful for all styleguidist users.
In playgroundRenderer component i added 
image share example button which copy code to url (maybe clipboard in the future).
Then user can copy url and pass to another developer which will past it to browser and see playground with example.
Please, review PR, and maybe you have suggestions how to implement this functionality more simple

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,8 @@ import {
 	processSections,
 	setSlugs,
 	slugger,
+	isPlayground,
+	getPlayground,
 } from './utils/utils';
 import StyleGuide from 'rsg-components/StyleGuide';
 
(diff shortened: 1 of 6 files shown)


sapegin:
Thank for the pull request! I think it’d be better to implement as a plugin, so I invite you to discuss possible plugin API for Styleguidist: https://github.com/styleguidist/react-styleguidist/issues/354

I don’t really want to add all possibly useful features to the core and I believe plugins would be much more powerful feature but I need help with defining an API. I’ve also seen theme switcher in your style guide which could be another plugin and I think you’d like to group these buttons somehow.

PR #425, opened 2017-05-02
Update Components.md

n1313:
This PR adds a bit of background about RSG and react-docgen and why some kinds of components are unfortunately not going to work for us.


docs/Components.md
```diff
@@ -57,3 +57,9 @@ module.exports = {
   ]
 };
 ```
+
+## Limitations
+
+React Styleguidist is built on top of [react-docgen](https://github.com/reactjs/react-docgen). It supports components defined via `React.createClass`, ES2015 class definitions and functions (stateless components), with optional FlowType annotations. If your components do not conform to these syntaxes, then they might not be correctly parsed by react-docgen and not be presented correctly by React Styleguidist.
```
sapegin: FlowType annotations → Flow type annotations

And also TypeScript via https://github.com/pvasek/react-docgen-typescript

> not be presented correctly by React Styleguidist.

In most (all?) cases they will be ignored with a warning in terminal.

sapegin:
Cool, thanks! Much better than I could write ;-)

docs/Components.md
n1313: > FlowType
I have copied that from docgen readme. What is the correct spelling?

> ignored with a warning
I'd rather not "overpromise" on this. It is possible to have a half-broken component that works in markdown but has incomplete proptypes (for example, if you declare them by reusing some dictionary instead of listing everything in `.oneOf([])`) or something minor like that.

sapegin: The project is called Flow: https://flow.org/

n1313:
Corrections were made.

sapegin:
🍕

PR #429, opened 2017-05-05
Docs: Fix question order in Cookbook

kohgpat:
Hello!

I found a little mistake IMO. The note after `styled-components` question said:

> **Note:** to change style guide styles use `theme` and `styles` options (see the next question).

But the next question is about `Preact` and the question right after is exactly about `theme` and `styles`. So I changed the order to match "note" in `styled-components` question.

Thank you for the project!

Best,
Nikolay.


sapegin:
Nice catch, thanks!

PR #431, opened 2017-05-06
Add logging for handled errors in Preview

nkt:
This change makes easier debugging.
Before this change you can't understand why and how did you get an error

After merge, developer would be able to discover errors in console:


src/rsg-components/Preview/Preview.js
```diff
@@ -129,6 +129,9 @@ export default class Preview extends Component {
 		this.setState({
 			error: err.toString(),
 		});
+
+		// eslint-disable-next-line
+		console.error(err);
```
sapegin: `// eslint-disable-line no-console` would be more explicit.

PR #433, opened 2017-05-08
Add the new logo as favicon to templates/index.html

ankri:
I added the cool new logo as the favicon to the `scripts/templates/index.html`

image

PS: Before I found the cleaner version with the transparent background I used the [logo](https://avatars2.githubusercontent.com/u/23550189?v=3&s=200) of the organization. You can check out what this would look like in my previous [commit](https://github.com/styleguidist/react-styleguidist/commit/de68e02d17a45e3c7e89adb70780eb1a94450939#diff-6b35c02d31b6cb9e72acd00eb00aed42L6)


sapegin:
Thanks for the pull request! I like the idea ;-) A few thoughts:

1. The scaled down logo doesn’t look good. I’ve tried to use an octopus emoji on the upcoming Styleguidist site and I like it more so far:

image 2017-05-09 at 9 49 36 am

2. I don’t like the idea of hardcoding the icon to the HTML, and keeping it as base64 in the source code as well.

3. There should be an easy way to use a custom favicon. It was possible to put an icon into a root folder. With this change the only way to change an icon will be using a custom template.

I don’t have a good solution to 2 and 3 now though.

ankri:
The logo does not scale well because the aspect ratio of the logo I used was not 1:1. Maybe @SaraVieira can chime in and provide a Logo with an aspect ratio of 1:1 ?

On the other hand I don't mind using the emoji. I will look into 2. and 3. 👍

SaraVieira:
hello @ankri !

You can find the logo here : https://github.com/styleguidist/brand/blob/master/rs-logo.png

Any modifications you need please let me know

sapegin:
I’m closing this for now because I don’t see an easy way to implement that yet. Feel free to submit a new pull request.

PR #434, opened 2017-05-09
fix(235) - Move open isolated

### LINKED ISSUE #235: New "Open isolated" encroaches on components
monosnap 2016-11-17 08-14-21

Maybe it should just be moved down next to Show code?

SaraVieira:
- Removes open isolated from preview

Fixes #235

### DIFF
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -1,6 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { MdFullscreen, MdFullscreenExit, MdArrowDropDown, MdArrowDropUp } from 'react-icons/lib/md';
+import MdFullscreen from 'react-icons/lib/md/fullscreen';
+import MdFullscreenExit from 'react-icons/lib/md/fullscreen-exit';
+import MdArrowDropDown from 'react-icons/lib/md/arrow-drop-down';
+import MdArrowDropUp from 'react-icons/lib/md/arrow-drop-up';
 import Editor from 'rsg-components/Editor';
 import Link from 'rsg-components/Link';
 import Preview from 'rsg-components/Preview';
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { MdFullscreen, MdFullscreenExit, MdArrowDropDown, MdArrowDropUp } from 'react-icons/lib/md';
 import Editor from 'rsg-components/Editor';
 import Link from 'rsg-components/Link';
 import Preview from 'rsg-components/Preview';
@@ -13,51 +14,42 @@ const styles = ({ space, color, fontFamily, fontSize, borderRadius }) => ({
 		border: [[1, color.border, 'solid']],
 		borderRadius: [[borderRadius, borderRadius, 0, borderRadius]],
 		marginTop: space[0],
-		'&:hover $isolatedLink': {
-			isolate: false,
-			opacity: 1,
-		},
 	},
 	preview: {
 		marginBottom: space[0],
 		padding: space[2],
 	},
 	codeToggle: {
-		position: 'absolute',
-		right: -1,
-		margin: 0,
-		padding: [[space[0], space[1]]],
+		outline: 'none',
+		border: 'none',
 		fontFamily: fontFamily.base,
 		fontSize: fontSize.base,
-		lineHeight: 1,
-		color: color.link,
-		border: [[1, color.border, 'solid']],
-		borderTop: 0,
-		borderBottomLeftRadius: borderRadius,
-		borderBottomRightRadius: borderRadius,
+		color: color.light,
+		transition: 'all 200ms ease',
 		cursor: 'pointer',
-		'&:hover, &:active': {
-			isolate: false,
+		'&:hover': {
 			color: color.linkHover,
+			fontFamily: fontFamily.base,
+			fontSize: fontSize.base,
 		},
 	},
-	showCode: {
-		composes: '$codeToggle',
-		backgroundColor: color.baseBackground,
+	toolbar: {
+		display: 'flex',
+		padding: [[space[0], space[2]]],
+		'& > *:not(:last-child)': {
+			marginRight: space[1],
+		},
 	},
-	hideCode: {
-		composes: '$codeToggle',
-		backgroundColor: color.codeBackground,
-	},
-	isolatedLink: {
-		position: 'absolute',
-		top: 0,
-		right: 0,
-		padding: [[space[0], space[1]]],
-		fontFamily: fontFamily.base,
-		fontSize: fontSize.base,
-		opacity: 0,
-		transition: 'opacity ease-in-out .15s .2s',
+	icons: {
+		width: 20,
+		height: 20,
+		color: color.light,
+		transition: 'all 200ms ease',
+		'&:hover': {
+			width: 20,
+			height: 20,
+			color: color.linkHover,
+		},
 	},
 });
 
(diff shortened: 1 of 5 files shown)


src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { MdFullscreen, MdFullscreenExit, MdArrowDropDown, MdArrowDropUp } from 'react-icons/lib/md';
```
sapegin: I think we should import particular icon modules, otherwise we’ll have the whole library in the bundle. And wouldn’t rely on tree shaking in Webpack 2 ;-)

src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -13,51 +14,42 @@ const styles = ({ space, color, fontFamily, fontSize, borderRadius }) => ({
...
-		border: [[1, color.border, 'solid']],
-		borderTop: 0,
-		borderBottomLeftRadius: borderRadius,
-		borderBottomRightRadius: borderRadius,
+		color: color.light,
+		transition: 'all 200ms ease',
 		cursor: 'pointer',
-		'&:hover, &:active': {
-			isolate: false,
+		'&:hover': {
```
sapegin: Probably better to disable isolation then to repeat all properties if it doesn’t break anything.

src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -13,51 +14,42 @@ const styles = ({ space, color, fontFamily, fontSize, borderRadius }) => ({
...
 	preview: {
 		marginBottom: space[0],
 		padding: space[2],
 	},
 	codeToggle: {
-		position: 'absolute',
-		right: -1,
-		margin: 0,
-		padding: [[space[0], space[1]]],
+		outline: 'none',
```
sapegin: If you disable default focus styles you need to provide some fallback to make it accessible with a keyboard.

src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -13,51 +14,42 @@ const styles = ({ space, color, fontFamily, fontSize, borderRadius }) => ({
...
 		},
 	},
-	showCode: {
-		composes: '$codeToggle',
-		backgroundColor: color.baseBackground,
-	},
-	hideCode: {
-		composes: '$codeToggle',
-		backgroundColor: color.codeBackground,
+	toolbar: {
```
sapegin: I’d use react-group here ;-)

sapegin:
Should we use a branch for the new UI? I don’t want it to block other features and don’t want to release many small changes separately.

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: react-group ? What do you mean ?

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: I will set a focus with the outline , good catch 👍

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: `isolate: false` what does this do ?

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: I haven't had problems in importing like this but sure , I can divide this in 4 lines 👍

SaraVieira:
@sapegin I think we should but maybe after this PR goes because this one actually closes an issue

src/rsg-components/Playground/PlaygroundRenderer.js
sapegin: It disables jss-isolate. That’s why you have to copy styles from non-hover state when you don’t have `isolate: false`.

src/rsg-components/Playground/PlaygroundRenderer.js
sapegin: https://sapegin.github.io/react-group/ — we already use it in Styleguidist.

sapegin:
This PR alone creates even more inconsistency with opening / closing isolation ;-|

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: OHHHHHHHHHHHHHHHHHH

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: Outline added 👍

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: Added 👍

SaraVieira:
@sapegin I can add the same icon at the top , I tought it would be done in phases but I can dp that really quickly 😄

sapegin:
That’s why I want to use a branch, so we could iterate not in production ;-)

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: But why use react group ?

src/rsg-components/Playground/PlaygroundRenderer.js
sapegin: Did you measure bundle size before and after? ;-|

src/rsg-components/Playground/PlaygroundRenderer.js
sapegin: Because it’s easy and you could avoid this ugly selector with `:not` ;-)

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: Fixed 👍

src/rsg-components/Playground/PlaygroundRenderer.js
SaraVieira: You don't like my css mastery ? :p

SaraVieira:
I created a ui branch and it's pointed to there now 👍

src/rsg-components/Playground/PlaygroundRenderer.js
sapegin: 👻

SaraVieira: Oh , you made this, you selling me your stuff :p

SaraVieira:
Changes made @sapegin

src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -73,8 +74,8 @@ export function PlaygroundRenderer({
 			<div className={classes.preview} data-preview={name ? name : ''}>
 				<Preview code={code} evalInContext={evalInContext} />
 			</div>
-			<section className={classes.toolbar}>
-				<div>
+			<Group separator=" " className={classes.toolbar}>
+				<div className={classes.toolbarItem}>
```
sapegin: Do we still need a margin here? If space isn’t enough then react-group is not the best choice because now we’re mixing spaces (a real space) with margins.

SaraVieira: I tried adding only the spaces and they were too close together and you could click one by accident

sapegin: Then maybe add padding to the button instead of margin to increase clickable area?

SaraVieira: The thing is I think it makes sense semantically ( not idea if that is spelled correctly) to have their own margins , that way when something is added it always has that margin associated

sapegin: That’s a tough question ;-) I often keep margins and layout in a separate layer. So do whatever you prefer now and when we extract the button to a separate component we’ll talk again what to do ;-)

SaraVieira: I added a div with a class of toolbarItem arround the component to actually give the margin some abstration from the component :)

sapegin:
I think it’s good enough for the very first iteration, merged, thanks! :shipit:

SaraVieira:
🎉 🎉

PR #437, opened 2017-05-10
Fixes a crash when using a defaultProp that is not listed in props

ankri:
This pull request fixes a crash, that occurs when you use a defaultProp that is not listed in the props. The cause was my previous PR not checking whether the incoming description might be `undefined`.

The error is kind of nasty because the styleguidist wont compile. 

I added the fix and the test to replicate the error.


sapegin:
Cool, thanks!

Issue #423, opened 2017-05-02
Examples are not published


Sector22:
Let me start off with that I think this is a great project and I'm really enjoying it so far! (:

I have been using it for some time now, but I've run into a bug I have some trouble figuring out what is going wrong.

I have a config that seems to be working just fine. The server and build commands are working and everything seems to be build correctly. That is until I try to build the styleguide inside a node docker container. I'm using a gitlab runner for building the styleguide and publishing it to a S3 bucket.
The files that it is outputting seem to be fine, I don't get any error or warnings. It can find all my components and sections. The only thing that is missing are the examples.

When inspecting the styleguide with the React devtools 2 things stand out. First, the ReactComponent holding my Component has a prop `hasExamples:false`. Second the ComponentRenderer does contains an example object with a prop `type:y`.

I suspect something might go wrong with some filepath names. 

Now I know this is not a whole lot to go on. Nor do I expect this to be solved very easily. Mostly I'm looking for some pointers for things I could try or look at to figure out where it is going wrong. I will continue to investigate later when I have more time. If I figure it out I'm happy to create a PR with a fix.

Thanks in advance for any help!

Sector22:
Just in case it help I uploaded my styleguide.config.js as well:

https://gist.github.com/Sector22/184e10ac7435e62da24d797a77c9ddba

sapegin:
Could you publish the generated style guide somewhere? Is Node version different inside the container?

Sector22:
It is running in a node 7.2.0 container, locally I'm running 7.3.0
Also using webpack 2.2.1

You can view the styleguide here: [http://test.snailhousecampervans.com.s3-eu-central-1.amazonaws.com/styleguide/index.html](http://test.snailhousecampervans.com.s3-eu-central-1.amazonaws.com/styleguide/index.html)

sapegin:
Is your files named `README.md` in upper case?

Sector22:
yes, they are.

sapegin:
That’s the reason I think. You’re using case-insensitive file system at home (Mac, Windows) and your Docker container has case-sensitive one (Linux). The default [getExampleFilename](https://github.com/styleguidist/react-styleguidist/blob/master/scripts/schemas/config.js#L60-L77) only checks for `Readme.md`. 

A fix to make the check case-insensitive is very appreciated ;-)

Sector22:
Ah there we go! Nice catch! I might have some time this weekend to fix it. (: Thanks for your help!

PR #440, opened 2017-05-12
Fix: Rewrite getExampleFilename logic to make it case insensitive (#423)

dotcs:
This PR fixes the `getExampleFilename` method and re-implements the logic such that the filename can be case-insensitive as requested here: https://github.com/styleguidist/react-styleguidist/issues/423#issuecomment-298536403.


scripts/schemas/config.js
```diff
@@ -62,15 +62,23 @@ module.exports = {
...
-				}
+			const files = fs.readdirSync(dirname);
+			const lowerFiles = files.map((name) => name.toLowerCase());
+
+			// Try to find a readme markdown file (`Readme.md`) or a markdown file
+			// named by the component (e.g. `Button.md`).
+			// Search is case-insensitive, so `README.md` or `BUTTON.md` is also
+			// valid.
+			const readmeIndex = lowerFiles.indexOf('readme.md');
+			const componentIndex = lowerFiles.indexOf(`${componentName.toLowerCase()}.md`);
```
sapegin: I’d move this line after the first condition; and you don't need an else because you have return.

scripts/schemas/config.js
```diff
@@ -62,15 +62,23 @@ module.exports = {
...
-				componentPath.replace(path.extname(componentPath), '.md'),
-			];
+			const componentName = path.basename(componentPath, path.extname(componentPath));
+			const dirname = path.dirname(componentPath);
 
-			for (const file of files) {
-				if (fs.existsSync(file)) {
-					return file;
-				}
+			const files = fs.readdirSync(dirname);
```
sapegin: Should we cache it somehow? Imagine a project with all components in a single folder.

dotcs:
@sapegin Thanks for the feedback. I'll have a look hopefully somewhen later this weekend. I'll see if I can implement cache and see what I can do about failing tests (somehow they worked locally ... ??). I'll come back to you when I'm done.

scripts/schemas/config.js
dotcs: Done. I think we still need else branch because it can be that neither a `readme.md` nor a `mycomponent.md` file is available. In this case `readmeIndex` and `componentIndex`  have both the value `-1` and both conditions are wrong. In this case `false` is returned as in the previous implementation. Please have a look in the updated code.

scripts/schemas/config.js
dotcs: I've had a look, but I'm not sure where caching could be done. `getExampleFilename` is called in `loaders/props-loader.js` and `processComponent.js`. The first one is already cached via webpack as it's a cachable loader (please correct me if I'm wrong). The second one is called to process a component in case it has changed, so I think in this case we should update anyway. Where in the application logic would you see the caching layer?

sapegin: I mean to cache `readdirSync` because it will be called for each component possible with the same results if all of them are in the same folder.

dotcs:
@sapegin Sorry for the long delay. I've been quite busy within the last two weeks. I've implemented a cache layer into the `getExampleFilename` implementation. Can you have a look if this fits what you have thought of?

The idea goes like this:
The `props-loader` is marked as cachable, so webpack will only call the real implementation if the underlying file system has changed somehow. In this case we have to invalidate the cache. But for all calls of `getExampleFilename` within one iteration after files have changed the cache can be used to reduce file io reads.

What this PR misses is documentation of the cache as the second param of the `getExampleFilename` implementation. I leave this up to you if you think it should be implemented.

Side note: I had to fix some lint issues (see commit dcf17a9) because code from master was invalid...?!

sapegin:
> Side note: I had to fix some lint issues (see commit dcf17a9) because code from master was invalid...?!

Looks like you need to `npm install` after merging master to your branch.

scripts/schemas/config.js
```diff
@@ -62,11 +62,19 @@ module.exports = {
 	},
 	getExampleFilename: {
 		type: 'function',
-		default: componentPath => {
+		default: (componentPath, cache) => {
```
sapegin: Could you please explain benefits of your implementation over _.memoize?

I’d make a cached case-insensitive version of `fs.existsSync` and pass it to to `getExampleFilename`. Passing just an object doesn’t look useful and will be hard to explain in the docs.

PR #452, opened 2017-05-21
Fix/exit isolated button fix

stepancar:
What happens:
When going into the isolation example and trying to go back button have text 'Exit isolated'.

What should happen: Button should have text 'Open isolated'

To reproduce,

go to https://react-styleguidist.js.org/examples/basic/#button

Press "Open isolated" in Playground
In the newly opened page, press "Exit isolated"
Observe you are now on component page, but button in playground still have "Exit isolated text"

This fix update Playground on context change.
Added test

### DIFF
--- a/src/rsg-components/Playground/Playground.js
+++ b/src/rsg-components/Playground/Playground.js
@@ -33,8 +33,12 @@ export default class Playground extends Component {
 		});
 	}
 
-	shouldComponentUpdate(nextProps, nextState) {
-		return nextState.code !== this.state.code || nextState.showCode !== this.state.showCode;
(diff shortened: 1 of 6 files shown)


sapegin:
Thanks for the pull request! I think it’s already fixed in the `ui` branch, I’ll check again.

stepancar:
as I see , it's not fixed https://github.com/styleguidist/react-styleguidist/blob/ui/src/rsg-components/Playground/Playground.js#L36

sapegin:
It’s not the latest code anyway, *I’ll check it again when I finish the new UI*.

sapegin:
I can’t reproduce it in the `ui` branch now (pushed the latest changes).

PR #474, opened 2017-06-07
fix(env): do not override NODE_ENV if already set

aversini:
This recent [styleguidist update](https://github.com/styleguidist/react-styleguidist/commit/c905c93372536bc7ce6f44a6040f5b26fb23b425) to support Create React App is overriding NODE_ENV, even if it's already set by the user.
 
This PR is an attempt to fix this and still support Create React App:
- If NODE_ENV is set, use it
- In ReactComponent, instead of checking for `development`, check for `not production`, allowing the use of `debug` for example and treat it the same as `development`

This is allowing a user to do this:
```js
NODE_ENV=debug styleguidist server
```
and configure their webpack/styleguidist configuration to act accordingly on 'debug'


src/rsg-components/ReactComponent/ReactComponent.js
```diff
@@ -5,7 +5,7 @@ import Methods from 'rsg-components/Methods';
 import Examples from 'rsg-components/Examples';
 import ReactComponentRenderer from 'rsg-components/ReactComponent/ReactComponentRenderer';
 
-const ExamplePlaceholder = process.env.NODE_ENV === 'development'
+const ExamplePlaceholder = process.env.NODE_ENV !== 'production'
```
sapegin: Good catch!

aversini: Actually that is what causing CI to fail. 

I'm assuming Jest snapshots have been taken with NODE_ENV set to 'production'. In CI, the test runs with NODE_ENV=test, so it makes my test (process.env.NODE_ENV !== 'production') truthy. That means that in test mode, component are not rendered in 'production' mode, and therefore the snapshots do not match.

Before my change, with NODE_ENV set to 'test' would render the component in production mode... quite a conundrum here. I have a feeling that it would make more sense to revert this change and keep snapshots 'production'. It means that ONLY when specifically set to 'development', component are rendered that special way? Let me know what you think @sapegin

sapegin: I thinks != production is more common way to check that but we’ll probably need to grep the code for `development` and update checks in other places too ;-)

aversini: You're right. So I grep'd around and didn't find any places that looked like they should be updated, beside that one line. So my suggestion is to consider the 'test' path by updating the condition to:

```js
const ExamplePlaceholder = process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test'
  ? require('rsg-components/ExamplePlaceholder').default
  : () => <div />;
```

This has no impact for production code and it should allow CI to pass.

sapegin: I think we can just update snapshots because it doesn’t really matter if it renders placeholder or not if it does the right thing ;-)

aversini:
@sapegin I finally got around this PR and by virtue of synchronizing it with the latest from master, CI tests are all passing 😄 Let me know if you need anything else. Thanks!

sapegin:
Awesome, thanks!

PR #477, opened 2017-06-08
changed es6-promise/auto to es6-promise.polyfill() to avoid build error

davidsims9t:


sapegin:
What kind of errors you have? According to the docs it should be the same — https://github.com/stefanpenner/es6-promise#auto-polyfill

davidsims9t:
I'm getting an error that the path es6-promise/auto doesn't exist. It only occurs after updating to react-styleguidist 5.3.2.

sapegin:
This is weird because it would break CI and won’t work for everybody else. Could it be something unusual in your webpack config?

davidsims9t:
I suppose it could be something in Webpack, but it's strange that it's only when I upgrade to 5.3.2 that I see this error.

sapegin:
Could you run `npm ls es6-promise`? It looks like it uses a different version of the package — may be because of `resolve.modules` in webpack config.

PR #482, opened 2017-06-09
Makes getExampleFilename case insensitive

verkholantsev:
Current implementation of `getExampleFilename()` works in different way on Linux (case sensitive filesystem) and OS X (case insensitive filesystem). This pull request fixes this issue.

### DIFF
--- a/scripts/schemas/config.js
+++ b/scripts/schemas/config.js
@@ -16,6 +16,9 @@ const createDisplayNameHandler = require('react-docgen-displayname-handler')
 const findUserWebpackConfig = require('../utils/findUserWebpackConfig');
 const getUserPackageJson = require('../utils/getUserPackageJson');
 const consts = require('../consts');
+const memoize = require('lodash/memoize');
+
+const memoizedReaddirSync = memoize(fs.readdirSync);
 
 module.exports = {
 	assetsDir: {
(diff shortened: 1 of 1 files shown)


sapegin:
Thanks for the pull request @verkholantsev! We already have an PR for that — #440 but I’m not sure if @dotcs is going to finish it. Please take a look at the comments there because your implementation has similar issues that I mentioned in the review ;-)

verkholantsev:
@sapegin Just added commit that uses `_.memoize` to create cache for `fs.readdirSync()`. Are there any more issues, that I didn't noticed it #440?

sapegin:
I think we should clean the cache on code updates otherwise it won’t work when you add new files when Styleguidist is running. But better to test it to make sure how exactly it works ;-)

verkholantsev:
@sapegin You are right, this implementation of cache breaks files watching in live server mode.

What way of implementation do you think is the best? Function `console.trace()` shows me two consumers of  `getExampleFilename()`:
1. `loaders/styleguide-loader`;
2. `loaders/props-loader`.

I've got an idea, but I think it is quite dirty:
1. Create module that exposes memoized readdirSync and method to reset the cache;
2. Import this method to that two loaders and reset the cache they are executed.

In this case there will no need to pass some parameters to `getExampleFilename()` from loaders, but it will create spaghetti-like dependency.

sapegin:
It should be enough to reset cache just in styleguide-loader I think.

verkholantsev:
My experiments are failed, I'm closing this PR.

PR #485, opened 2017-06-09
fix(471) - Copy path icon

### LINKED ISSUE #471: Copy path button
To be able to quickly copy import statements. Something like this:

image 2017-06-07 at 3 51 27 pm

* [ ] Choose an icon, [MdContentCopy](https://gorangajic.github.io/react-icons/md.html) should be good enough
* [ ] Add a button using [copy-text-to-clipboard](https://github.com/sindresorhus/copy-text-to-clipboard)

SaraVieira:
Adds copy path icon : 
copy

Closes #471


package.json
```diff
@@ -37,6 +37,7 @@
     "clean-webpack-plugin": "^0.1.16",
     "codemirror": "^5.26.0",
     "common-dir": "^1.0.1",
+    "copy-text-to-clipboard": "^1.0.0",
```
sapegin: Looks like it’s devDependencies.

src/rsg-components/ReactComponent/ReactComponentRenderer.js
```diff
@@ -21,6 +23,11 @@ const styles = ({ color, fontSize, fontFamily, space }) => ({
 		color: color.base,
 		fontSize: fontSize.text,
 	},
+	copyButton: {
+		cursor: 'pointer',
+		fill: color.light,
+		marginLeft: space[1],
```
sapegin: I think it should be orange on hover, like toolbar buttons.

src/rsg-components/ReactComponent/ReactComponentRenderer.js
```diff
@@ -38,7 +45,10 @@ export function ReactComponentRenderer({
 		<div className={classes.root} id={name + '-container'}>
 			<header className={classes.header}>
 				{heading}
-				<div className={classes.pathLine}>{pathLine}</div>
+				<div className={classes.pathLine}>
+					{pathLine}
+					<MdContentCopy className={classes.copyButton} onClick={() => copy(pathLine)} />
```
sapegin: And it should be a real <button> so probably ToolbarButton would work if add a prop for a small version.

SaraVieira:
All done :D

src/rsg-components/ReactComponent/ReactComponentRenderer.js
```diff
@@ -21,6 +24,20 @@ const styles = ({ color, fontSize, fontFamily, space }) => ({
 		color: color.base,
 		fontSize: fontSize.text,
 	},
+	copyButton: {
```
sapegin: You don’t need any extra styles if you’re using ToolbarButton — it already has everything. You only need to add a prop for a small version.

sapegin:
You also need to update npm lock file — it should fix CI failures.

SaraVieira:
@sapegin About that , I don't have npm5 because it breaks the now cli so I can't do that :(

I use yarn

sapegin:
Then move it to the right section in package.json and I’ll do that ;-)

SaraVieira:
I installed npm5 and updated the lock , also added the prop. Sorry about that , I was half asleep when I read your comment 😝

sapegin:
Now something is wrong on CI. Does tests pass for you locally?

sapegin:
Ha! I can’t even install dependencies locally, have the same `Cannot find module 'babel-helpers'` error.

SaraVieira:
@sapegin sorry , that's happening in mine too when I run the tests.
Can you please delete the lock and npm install again ?

I may have screwed up the lock

sapegin:
I’ve regenerated the lock file so `npm install` works now. But there are some test failures — most of them are just out of date snapshots (run `jest -u`). But the one with `undefined` class I think because of this:

```js
 	const classNames = cx({
  		[classes.button]: true,
  		[className]: true,
  		[classes.small]: small,
  	});
```

I’d try this instead (it’s more clear anyway ;-):

```js
 	const classNames = cx(classes.button, className, {
  		[classes.small]: small,
  	});
```

SaraVieira:
@sapegin Tests fixed 👍

package.json
SaraVieira: copy-text-to-clipboard is a FE dependency so it's not a dev dependency

src/rsg-components/ReactComponent/ReactComponentRenderer.js
SaraVieira: fixed 👍

package.json
sapegin: Sorry, my bad ;-)

sapegin:
Looks like we’ll have to fork copy-text-to-clipboard because they use arrow functions and it breaks UglifyJS (look at CI report) ;-(

https://github.com/sindresorhus/copy-text-to-clipboard/issues/5

sapegin:
Try this one: https://github.com/feross/clipboard-copy

SaraVieira:
@sapegin build seems to pass now , at least locally with the new one

SaraVieira:
I'll create some tests @sapegin

sapegin:
I’d mock clipboard-copy and test if it’s called with the right text when you click the button.

src/rsg-components/ReactComponent/ReactComponent.spec.js
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import slots, { DOCS_TAB_USAGE } from '../slots';
 import ReactComponent from './ReactComponent';
 import { ReactComponentRenderer } from './ReactComponentRenderer';
+import copy from 'clipboard-copy';
+
+jest.mock('clipboard-copy', () => jest.fn());
```
sapegin: `jest.mock('clipboard-copy')` should be enough.

src/rsg-components/ReactComponent/ReactComponentRenderer.js
```diff
@@ -38,7 +44,12 @@ export function ReactComponentRenderer({
 		<div className={classes.root} id={name + '-container'}>
 			<header className={classes.header}>
 				{heading}
-				<div className={classes.pathLine}>{pathLine}</div>
+				<div className={classes.pathLine}>
+					{pathLine}
+					<ToolbarButton small className={classes.copyButton}>
+						<MdContentCopy onClick={() => copy(pathLine)} />
```
sapegin: Just noticed that we still have a callback on an icon instead of a button ;-)

sapegin:
Merged manually in 5fbbc6773964042e350d12de8126f325232a48a5, thanks for the pull request!

Issue #483, opened 2017-06-09
Simpler way of replacing Styleguidist components


sapegin:
Right now [to use](https://react-styleguidist.js.org/docs/cookbook.html) Redux providers, React Intl, Relay etc. you need to use webpack aliases like this:

```js
// styleguide.config.js
const path = require('path');
module.exports = {
  webpackConfig: {
    resolve: {
      alias: {
        'rsg-components/Wrapper': path.join(__dirname, 'lib/styleguide/Wrapper')
      }
    }
  }
};
```

I think we have enough use cases to add a special config option for that. It could look like this:

```js
// styleguide.config.js
const path = require('path');
module.exports = {
  styleguideComponents: {
    Wrapper: path.join(__dirname, 'lib/styleguide/Wrapper')
  }
};
```

* [x] Add a new config option to the schema (scripts/schemas/config.js)
* [x] Generate webpack aliases based on the new option (scripts/make-webpack-config.js)
* [x] Add docs for the config option (docs/Configuration.md)
* [x] Update examples in the docs to use the new option instead of webpack aliases (mostly the cookbook, but better to check other files)

stepancar:
I think ,we don't need config option for that.
Webpack aliases is simple to understand.
And i think that aliases like this (with full path)
``` javascript
 alias: {
       './node_module/react-styleguidist/rsg-components/Wrapper': pathToYourComponent
 }
```
is more simple to understand and more simple for override.
when i see 'rsg-components/Wrapper' i can't understand what is rsg-components wrapper.
It looks like another package

sapegin:
That’s a good reason to have a simple to understand option ;-)

stepancar:
:D

sapegin:
There‘s a PR: https://github.com/styleguidist/react-styleguidist/issues/504 ;-)

sapegin:
Released in [5.5.0](https://github.com/styleguidist/react-styleguidist/releases/tag/v5.5.0).

PR #487, opened 2017-06-10
Make rsg components overridable using dangerouslyUpdateWebpackConfig

MicheleBertoli:
I know that the usage of `dangerouslyUpdateWebpackConfig` is discouraged, but I was playing around with it ([I like to live dangerously](https://www.youtube.com/watch?v=dVSjCvGFMnM) :)) and I found out that I couldn't override the rsg components.

### DIFF
--- a/scripts/make-webpack-config.js
+++ b/scripts/make-webpack-config.js
@@ -111,13 +111,13 @@ module.exports = function(config, env) {
 		});
 	}
 
-	// Add components folder alias at the end so users can override our components to customize the style guide
-	// (their aliases should be before this one)
-	webpackConfig.resolve.alias['rsg-components'] = path.resolve(sourceDir, 'rsg-components');
-
 	if (config.dangerouslyUpdateWebpackConfig) {
 		webpackConfig = config.dangerouslyUpdateWebpackConfig(webpackConfig, env);
 	}
+	
+	// Add components folder alias at the end so users can override our components to customize the style guide
+	// (their aliases should be before this one)
+	webpackConfig.resolve.alias['rsg-components'] = path.resolve(sourceDir, 'rsg-components');
 
 	return webpackConfig;
 };


sapegin:
I think `dangerouslyUpdateWebpackConfig` to be truly dangerous — you receive the final webpack config and do whatever you want — most likely shoot yourself in the leg ;-)

You can override components if you add them *before* the alias. Also take a look at #483.

What are you trying to do by the way? ;-)

MicheleBertoli:
> You can override components if you add them before the alias

Absolutely, but if you add the aliases inside `dangerouslyUpdateWebpackConfig` (in the "old" way) they don't work because of the order.

For example:

```js
updateWebpackConfig(webpackConfig) {
  webpackConfig.resolve.alias['rsg-components/StyleGuide/StyleGuideRenderer'] = path.join(__dirname, 'lib/styleguide/StyleGuideRenderer');
  return webpackConfig;
},
````
As I said, I was just experimenting, so feel free to close this if you think it doesn't provide any value.

Thank you very much :)

**Comment by sapegin** [comment]
They will work if you’ll add them *before*, something like this:

```js
updateWebpackConfig(webpackConfig) {
  webpackConfig.resolve.alias = {
       'rsg-components/StyleGuide/StyleGuideRenderer': path.join(__dirname, 'lib/styleguide/StyleGuideRenderer',
      ...webpackConfig.resolve.alias,
   };
  return webpackConfig;
},
```

MicheleBertoli:
+1 that's my current solution :)
The only problem is that if someone doesn't know about this and follows the "old" way, the aliases don't work.

sapegin:
There’s no *old way*, it’s a new option introduced in 5.0 ;-) It should work that way: it allows you to do anything and it won’t fix your mistakes. It’s usage is highly discouraged.

I’m still not sure why do you need to use it ;-)

MicheleBertoli:
"Old" way = the way we used to override the components before v5 with `updateWebpackConfig`.
Someone (like me) may think they could replace `updateWebpackConfig` with `dangerouslyUpdateWebpackConfig`.

MicheleBertoli:
"Old" way = the way we used to override the components before v5 with `updateWebpackConfig`.
Someone (like me) may think they could replace `updateWebpackConfig` with `dangerouslyUpdateWebpackConfig` but it doesn't work.

PR #488, opened 2017-06-11
fix(486) - Only show props if there are props

### LINKED ISSUE #486: Do not show Props & Methods tab when there’s no props and methods

SaraVieira:
Closes #486

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,7 +54,7 @@
         "readable-stream": {
           "version": "2.2.11",
           "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.2.11.tgz",
-          "integrity": "sha512-h+8+r3MKEhkiVrwdKL8aWs1oc1VvBu33ueshOvS26RsZQ3Amhx/oO3TKe4lApSV9ueY6as8EAh7mtuFjdlhg9Q==",
+          "integrity": "sha1-B5azH412iAB/8Lk6gIjTSqF8D3I=",
           "dev": true
         },
         "string_decoder": {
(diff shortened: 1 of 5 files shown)


src/rsg-components/ReactComponent/ReactComponent.js
```diff
@@ -74,6 +75,7 @@ export default class ReactComponent extends Component {
 						: <ExamplePlaceholder name={name} />
 				}
 				tabButtons={
+					showPropsButton &&
```
sapegin: There may be other tabs, not just props. I’m thinking that this may be not that easy to fix ;-)

SaraVieira:
What are the other other tabs?

sapegin:
Doesn’t matter (there are none now) but the `Slot` component can render any number of items. So the props tab button should contain a check if there are props or methods and render null if there are none.

src/rsg-components/ReactComponent/ReactComponent.js
```diff
@@ -44,6 +44,7 @@ export default class ReactComponent extends Component {
 		const { component } = this.props;
 		const { name, slug, pathLine } = component;
 		const { description, examples = [], tags = {} } = component.props;
+		const showPropsButton = component.props.props || (component.props.methods || {}).length > 0;
```
sapegin: Also this check won’t work:

* `!![] === true` 
* `({}).length === undefined`

sapegin: Here is the right check: https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Usage/Usage.js

SaraVieira:
Didn't know that about the slot but the props button receives the whole props including the methods and props of the component so I updated the code to only show or hide the methods component @sapegin

package-lock.json
```diff
@@ -54,7 +54,7 @@
         "readable-stream": {
           "version": "2.2.11",
           "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.2.11.tgz",
-          "integrity": "sha512-h+8+r3MKEhkiVrwdKL8aWs1oc1VvBu33ueshOvS26RsZQ3Amhx/oO3TKe4lApSV9ueY6as8EAh7mtuFjdlhg9Q==",
+          "integrity": "sha1-B5azH412iAB/8Lk6gIjTSqF8D3I=",
```
sapegin: Looks like you have some weird version of npm. Why it changes all hashes to SHA1? ;-/

sapegin:
I’ve made some tweaks (see my commits) and merged. Thanks!

SaraVieira:
@sapegin Awesome !

PR #490, opened 2017-06-12
Remove border from TabButton

n1313:
This PR explicitly removes border from TabButton to make it look good on Windows.


sapegin:
I think it actually the same issue as #491 but it won’t hurt to have it explicitly removed.

PR #496, opened 2017-06-14
Allow markdown in pathLine

bluetidepro:
This doesn't change any of the current way pathLines are handles, but also allows you to be able to style them with markdown if you use a custom `getComponentPathLine`.

For example, this would produce a list:
```javascript
getComponentPathLine(componentPath) {
	const name = path.basename(componentPath, '.js');
	let dir = path.dirname(componentPath);
	dir = dir.replace(/src\/components\//, '../');
	return `* **Blah:** \`import { ${name} } from 'blah';\`
* **BlahBlah:** \`import ${name} from '${dir}/${name}';\``
},
```
The output would look something like this:

screen shot 2017-06-14 at 10 18 51 am

So, it would just allow people to return custom Markdown. It's a nice little extra feature, if anyone needs it. I had been using it as my own custom Renderer but figured I'd contribute back up to the project since it doesn't hurt anyone currently, just adds the ability if you want it.

(_This is my 1st contribution PR, so let me know if there is anything else I need to do in the PR_)


bluetidepro:
(It looks like the CI failed, but the tests were already failing locally, even before this change, so I don't think it has anything to do with my PR changes)

sapegin:
Tests shouldn’t fail locally — they don’t fail in master, but they are obviously failing because the code was changed ;-) You need to update snapshots (run `jest -u`).

sapegin:
In [5.5.0](https://github.com/styleguidist/react-styleguidist/releases/tag/v5.5.0) you can easily redefine Pathline component, so I’m closing this as we discussed before ;-)

Issue #497, opened 2017-06-16
Auto fence @example without a path to an example file


ambidexterich:
When an @example doclet has an example and not a path to a file:
```
/**
 * A button example
 * @example
 * <Button>My Text</Button>
 */
```
The contents of the example gets appended to the description which causes strange rendering

![image](https://user-images.githubusercontent.com/675222/27232145-e6bb919a-5282-11e7-8617-d8ac02ce7c09.png)

If you add the fence as part of example as is supported, JSDocs will render the fence inside it's own code block.

Adding the fence to code blocks automatically will allow proper rendering it both cases.

sapegin:
Would you mind send a pull request for that? Doesn’t look very complicated you already know what to do ;-)

ethanh1223:
Looks like PR was merged -- this closed?

sapegin:
@ethanh1223 Yup :shipit:

Issue #472, opened 2017-06-07
Make logging customizable


sapegin:
Right now Styleguidist uses `console.log` to show some potentially useful messages like:

```
No webpack config found. You may need to specify "webpackConfig" option in your style guide config:
https://github.com/styleguidist/react-styleguidist/blob/master/docs/Webpack.md
```

We should do it only when Styleguidist is started from CLI, not from the API (or tests). Would be nice to be able to customize the logging somehow — I don’t really know what would be the right approach — maybe pass a logging function ~~everywhere~~ or something like that.

sapegin:
This look very close: https://github.com/js-cli/glogg

adityavohra7:
Not having `react-styleguidist` output anything when its tests run definitely makes sense, but I wonder if there are situations wherein people may want the API to output stuff similar to what the CLI does? 

Say I wanted to abstract away or package up my JS infra (build + configs), sort of like what `create-react-app` does, and provide the consumers of this JS infra package with an npm run script like `start-workbench`, which under the hood uses the `react-styleguidist` API. 

Having this run script output warnings or compilation errors would definitely be useful, and would only be possible if the API had output (I think?). Though at the same time this `start-workbench` script could just shell out to the `react-styleguidist` CLI, or it could re-implement some of the output forwarding of the CLI [implemented here](https://github.com/styleguidist/react-styleguidist/blob/master/bin/styleguidist.js#L197). So I'm not sure 😅

sapegin:
@adityavohra7 That’s exactly what I want to do! I just not sure about the API and implementation. 

Probably pass a logger object to the API like this:

```js
const styleguide = styleguidist({
  components: './lib/components/**/*.js',
  logger: {
    error: msg => console.error(msg)
  }
});
```

sapegin:
There’s a PR: #510.

PR #500, opened 2017-06-19
Check Example File Exists

ambidexterich:
I had to do some massaging to get the example file to resolve to the correct directory but you might have an easier way to do it.

### DIFF
--- a/loaders/utils/getProps.js
+++ b/loaders/utils/getProps.js
@@ -23,20 +23,14 @@ const getDoctrineTags = documentation => {
 	return _.groupBy(documentation.tags, 'title');
 };
 
-const validateExampleFile = (baseFile, exampleFile) => {
-	// In case there is no basefile such as running tests but exampleFile might still exist
-	const componentFile = baseFile || '';
-	if (typeof exampleFile !== 'string') {
-		return false;
-	}
-
-	const cleanFilePath = exampleFile.trim();
-	const exampleFilepath = path.resolve(path.dirname(componentFile), cleanFilePath);
+const doesExampleFileExist = (basepath, exampleFile) => {
+	const exampleFilepath = path.resolve(path.dirname(basepath), exampleFile);
 	const doesFileExist = fs.existsSync(exampleFilepath);
 
+	// Only warn when all conditions are met but file still isn't found
 	if (!doesFileExist) {
-		// eslint-disable-next-line
-		console.warn(`${cleanFilePath} is an invalid filepath.`);
+		// eslint-disable-next-line no-console
+		console.warn(`An example file ${exampleFile} defined in ${basepath} component not found.`);
 	}
 	return doesFileExist;
 };
@@ -74,9 +68,16 @@ module.exports = function getProps(doc, filepath) {
 
 		doc.description = highlightCode(removeDoclets(doc.description));
 
-		const exampleFileExists = validateExampleFile(filepath, doc.doclets.example);
+		let exampleFileExists = false;
+		let exampleFile = doc.doclets.example;
+		// doc.doclets.example might be a boolean or undefined
+		if (typeof doc.doclets.example === 'string') {
+			exampleFile = doc.doclets.example.trim();
+			exampleFileExists = doesExampleFileExist(filepath, exampleFile);
+		}
+
 		if (exampleFileExists) {
-			doc.example = requireIt(`!!${examplesLoader}!${doc.doclets.example}`);
+			doc.example = requireIt(`!!${examplesLoader}!${exampleFile}`);
 			delete doc.doclets.example;
 		}
 	} else {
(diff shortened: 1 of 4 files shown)


loaders/utils/getProps.js
```diff
@@ -22,16 +24,29 @@ const getDoctrineTags = documentation => {
...
 };
 
+const validateExampleFile = (baseFile, exampleFile) => {
+	const componentFile = baseFile || '';
+	if (typeof exampleFile !== 'string') {
+		return false;
+	}
+
+	// Remove newline characters included from react-docgen
+	const cleanFilePath = exampleFile.replace(/\n/g, '');
```
sapegin: `.trim()` should be enough here. Also I’d do it outside the function so you could remove the same code when you print a warning message.

loaders/utils/getProps.js
```diff
@@ -49,14 +64,21 @@ module.exports = function getProps(doc) {
...
 		doc.tags = getDoctrineTags(documentation);
 
 		doc.description = highlightCode(removeDoclets(doc.description));
-
-		if (doc.doclets.example) {
+		if (exampleFileExists) {
 			doc.example = requireIt(`!!${examplesLoader}!${doc.doclets.example}`);
 			delete doc.doclets.example;
+		} else if (doc.doclets.example) {
+			/* eslint-disable no-console */
```
sapegin: eslint-disable-nextline

loaders/utils/getProps.js
```diff
@@ -49,14 +64,21 @@ module.exports = function getProps(doc) {
...
+			console.warn(
+				`${chalk.bold(doc.doclets.example.toString().replace(/\n/g, ''))} is an invalid filepath.`
```
sapegin: I’d not use styling here, it may make #472 harder.

loaders/utils/getProps.js
```diff
@@ -22,16 +23,35 @@ const getDoctrineTags = documentation => {
 	return _.groupBy(documentation.tags, 'title');
 };
 
+const validateExampleFile = (baseFile, exampleFile) => {
+	// In case there is no basefile such as running tests but exampleFile might still exist
```
sapegin: Could we pass a name from tests? I’d remove both conditions.

loaders/utils/getProps.js
```diff
@@ -22,16 +23,35 @@ const getDoctrineTags = documentation => {
...
+	if (typeof exampleFile !== 'string') {
+		return false;
+	}
+
+	const cleanFilePath = exampleFile.trim();
+	const exampleFilepath = path.resolve(path.dirname(componentFile), cleanFilePath);
+	const doesFileExist = fs.existsSync(exampleFilepath);
+
+	if (!doesFileExist) {
+		// eslint-disable-next-line
```
sapegin: This way you’re disabling ESLint entirely for the next line — there may be other warnings.

loaders/utils/getProps.js
```diff
@@ -22,16 +23,35 @@ const getDoctrineTags = documentation => {
...
+		console.warn(`${cleanFilePath} is an invalid filepath.`);
```
sapegin: If file doesn’t exist it doesn’t mean the path is invalid. To make it useful for the user I’d try something like this:

> An example file FILEPATH defined in COMPONENTPATH component not found.

loaders/utils/getProps.js
```diff
@@ -22,16 +23,35 @@ const getDoctrineTags = documentation => {
 	return _.groupBy(documentation.tags, 'title');
 };
 
+const validateExampleFile = (baseFile, exampleFile) => {
```
sapegin: I’d call it something like `isExampleFileExists` and do only that. Right now it does too many things:

1. Check if example defined.
2. Check if file exists.
3. Prints a warning.
4. Do other magic like trimming file names and checking if component path was passed or not.

ambidexterich:
I've implemented all of your feedback. The warning is a stopgap until https://github.com/styleguidist/react-styleguidist/issues/497. I can't start using styleguidist in our product until this one is fixed.

sapegin:
The initial version was closer to what I’d merge unfortunately ;-|

loaders/utils/getProps.js
ambidexterich: Good catch

loaders/utils/getProps.js
ambidexterich: The reason for the check that example file is a string and ensure that the base file exists is that `path.resolve` fails if either one is not a string like when @example is true and passed from a test. Would you prefer that moved to the body of `getProps`?

sapegin: Definitely! ;-)

loaders/utils/getProps.js
ambidexterich: Updated

ambidexterich:
Submitted another attempt.

sapegin:
Cool, thanks!

Issue #499, opened 2017-06-17
Deprecate webpack 1


sapegin:
Show a warning that webpack 1 support will be removed in the next (?) major release.

https://twitter.com/iamsapegin/status/856846088447414273

Do not remove support until:

* [x] Create React App supports webpack 2 
* [ ] [webpack-blocks 1.0](https://github.com/andywer/webpack-blocks/tree/release/1.0) final release
* [x] [Webpack 3](https://github.com/webpack/webpack/releases) final release

PR #502, opened 2017-06-20
Tweak Webpack 2 conditions

adityavohra7:
`react-styleguidist` seems to have an incompatibility with Webpack 3 ([released today](https://github.com/webpack/webpack/releases/tag/v3.0.0)). Specifically, the issue I experienced was that the Webpack config used with Webpack 3 had its `config.resolve.extensions` [set to](https://github.com/styleguidist/react-styleguidist/blob/master/scripts/make-webpack-config.js#L32) `['.js', '.jsx', '.json', '']` (which Webpack 3 complains about):
```
$ ./node_modules/.bin/styleguidist server --config ./node_modules/js-infra/config/styleguidist/index.js
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.resolve.extensions[3] should not be empty.
```
I've tried tweaking the Webpack 2 conditions in [scripts/make-webpack-config.js](https://github.com/styleguidist/react-styleguidist/blob/master/scripts/make-webpack-config.js) a little so as to make it compatible with Webpack 3. Let me know what you think!

### DIFF
--- a/scripts/make-webpack-config.js
+++ b/scripts/make-webpack-config.js
@@ -12,7 +12,7 @@ const getWebpackVersion = require('./utils/getWebpackVersion');
 const mergeWebpackConfig = require('./utils/mergeWebpackConfig');
 const StyleguidistOptionsPlugin = require('./utils/StyleguidistOptionsPlugin');
 
-const isWebpack2 = getWebpackVersion() === 2;
+const isWebpackLessThan2 = getWebpackVersion() < 2;
 const sourceDir = path.resolve(__dirname, '../lib');
 const htmlLoader = require.resolve('html-webpack-plugin/lib/loader');
 
@@ -29,7 +29,7 @@ module.exports = function(config, env) {
 			chunkFilename: 'build/[name].js',
 		},
 		resolve: {
-			extensions: isWebpack2 ? ['.js', '.jsx', '.json'] : ['.js', '.jsx', '.json', ''],
+			extensions: isWebpackLessThan2 ? ['.js', '.jsx', '.json', ''] : ['.js', '.jsx', '.json'],
 			alias: {
 				'rsg-codemirror-theme.css': `codemirror/theme/${config.highlightTheme}.css`,
 			},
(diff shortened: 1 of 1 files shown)


scripts/make-webpack-config.js
```diff
@@ -12,7 +12,7 @@ const getWebpackVersion = require('./utils/getWebpackVersion');
 const mergeWebpackConfig = require('./utils/mergeWebpackConfig');
 const StyleguidistOptionsPlugin = require('./utils/StyleguidistOptionsPlugin');
 
-const isWebpack2 = getWebpackVersion() === 2;
+const isWebpackLessThan2 = getWebpackVersion() < 2;
```
n1313: So, `isWebpack1`? :)

adityavohra7: Heh I thought about that, but wasn't sure if there were any Webpack releases prior to 1.x.x. I guess it's fair to assume that most people are using 1.x.x? If that's the case, I can change this to `isWebpack1`. 😅

sapegin: Doesn’t matter — we don’t support them and we’ll remove webpack 1 support soon anyway ;-)

sapegin:
Thanks for the pull request! I’ve merged it as is because we’re going to remove this conditions soon anyway.

Are you interested in working on #499 as well? ;-)

adityavohra7:
@sapegin, sure! Is #499 mostly about showing a deprecation warning? 

Also, as far as supporting Webpack@1 goes, are we just waiting for a non-beta 1.x.x of [`webpack-blocks`](https://github.com/andywer/webpack-blocks) (they seem to have a [beta](https://github.com/andywer/webpack-blocks/releases/tag/v1.0.0-beta) out)?

sapegin:
@adityavohra7 Yes, and it should be a major release. 

Examples rework may be a good candidate for the next major: https://github.com/styleguidist/react-styleguidist/milestone/6

PR #505, opened 2017-06-22
Plugin API first draft

sapegin:
* [x] `plugins` config option
* [x] Load plugins and pass to the style guide
* [x] Convert slots to core plugins
* [ ] Ability to modify style guide config from a plugin (?)
* [x] Containers (similar to slow but wraps children in all passed wrappers)
* [ ] Add validation
* [ ] Add tests
* [ ] Add docs

I have no idea how to do:

* [ ] Store and change state for plugins (I suppose it should be separate state for style guide / component / example).
* [ ] Change style guide in any way (for example something similar to the `Wrapper` component).


src/rsg-components/Playground/Playground.js
```diff
@@ -71,21 +71,24 @@ export default class Playground extends Component {
 				preview={<Preview code={code} evalInContext={evalInContext} />}
 				tabButtons={
 					<Slot
-						name="exampleTabButtons"
+						name="exampleTabButton"
```
n1313: I think a more specific name like "playgroundTabButton" would be better here

src/rsg-components/Slot/Slot.js
```diff
@@ -9,30 +9,26 @@ export default function Slot({ name, active, onlyActive, className, props = {} }
...
-				// Pass fill ID to onClick event handler
-				// eslint-disable-next-line react/prop-types
-				onClick: onClick && ((...attrs) => onClick(id, ...attrs)),
-			};
-
-			Fill = render;
+	// { id: 'pizza', render: ({ foo }) => <div>{foo}</div> }
+	const rendered = fills.map(({ id, render }, index) => {
+		// Render only specified fill
+		if (onlyActive && id !== active) {
```
n1313: I'm afraid I don't fully understand this bit. Can there be a case when `active` is set, but `onlyActive` isn't?

src/rsg-components/Slot/Slot.js
```diff
@@ -9,30 +9,26 @@ export default function Slot({ name, active, onlyActive, className, props = {} }
 		throw new Error(`Slot "${name}" not found, available slots: ${Object.keys(slots).join(', ')}`);
```
n1313: The logic of this warning is probably incorrect: it warns if an existing slot can't find any fills (which should be totally ok: you could have an extension point but nothing occupying it at the moment), but instead it should warn if a fill requests a  slot that doesn't exist.

n1313:
This looks like a great first step. Some parts are a bit unclear to me, like how would a third-party plugin be registered and configured, but with a bit of documentation it should be fine.

I think it would be great if styles or classnames were somehow exposed to the plugins, so that they could use them to look like a natural part of the UI. A third-party button plugin will not have easy access to TabButton component and won't be able to easily follow any future changes in RSG UI (or any customizations that an end-user might have). Passing refs for relevant HTML nodes would be useful, too.

src/rsg-components/Playground/Playground.js
sapegin: 👍

src/rsg-components/Slot/Slot.js
sapegin: Yup, tab buttons for example: we need to show all of them but highlight only one.

src/rsg-components/Slot/Slot.js
sapegin: All slots are empty objects by default, so the check will work (it’s already working in 5.4 ;–)

sapegin:
> how would a third-party plugin be registered (webpack loaders will have to be reconfigured for this, probably?) and configured

Like this, via style guide config:

```js
module.exports = {
  plugins: {
    // These are Node module names
    'my-awesome-plugin': true,
    'my-another-plugin': {
      // Some options here
    }
  }
};
```

> I think it would be great if styles or classnames were somehow exposed to the plugins, so that they could use them to look like a natural part of the UI. A third-party button plugin will not have easy access to TabButton component and won't be able to easily follow any future changes in RSG UI (or any customizations that an end-user might have).

Plugins are run in the same webpack context as the Styleguidist UI so you could import `rsg-components/TabButton` or `rsg-componets/Styled` to have access to all theme variables.

> Passing refs for relevant HTML nodes would be useful, too.

Not sure about that yet ;-)

n1313:
Are plugins going to be loaded in alphabetical order? Should `plugins` in config be an array?

sapegin:
Alphabetical doesn’t make any sense ;-) It should work similar to PostCSS. Or we can make an array,  similar to Babel:

```js
module.exports = {
  plugins: [
    // These are Node module names
    'my-awesome-plugin',
    ['my-another-plugin', {
      // Some options here
    }]
  ]
};
```

n1313:
Yes, I think the babel way is good.

sapegin:
Cool, I’m fine with it too — both are familiar for many users.

MicheleBertoli:
Well done, @sapegin. 
This looks pretty cool, thank you very much.

My only concerns are about the props the plugins' components receive (I'm thinking about Snapguidist).
I'm also missing a lot of context here, so these might be dumb questions:
- which props do the components receive? different props for each slot?
- are the actions (e.g. hide/show for docs) pre-defined by the slots?
- is it possible for components in different slots communicate to each other (I guess is covered by "Store and change state for plugins")?

sapegin:
> which props do the components receive? different props for each slot?

All fills in a slot receive the same props except `active`, passed via `props` prop. Right now they don’t receive much but we just need to pass more ;-)

> are the actions (e.g. hide/show for docs) pre-defined by the slots?

There’s no such action ;-) Tabs are using the same `Slot` component and parent component stores active tab ID in its state and changes it on tab button click.

> is it possible for components in different slots communicate to each other (I guess is covered by "Store and change state for plugins")?

It should be possible and most likely will be covered by that, at least partially. But I need more feedback and ideas here ;-| @MicheleBertoli

n1313:
I have played a bit with this branch and it looks promising.

It is easy to start developing a test plugin: just checkout the branch and work in one of the examples. The plugin loader happily accepts local paths so you don't have to touch npm while you develop:
```js
  // in styleguide.config.js
  plugins: {
    //...
    './plugins/test': {
      options: { /* plugin configuration options */ }
    }
  }
```

The configuration options from above can be accessed in your plugin like this: 
```js
// in ./plugins/test/index.js
//...
const testPlugin = (globalConfig, pluginConfig) => {
  // plugin code
}
export default testPlugin;
```

I haven't tried doing this yet but it looks like you can export multiple components (or "fills") from one plugin module and they will share the module code and state, @MicheleBertoli.

I think that the next steps for this proposal should be:

1. considering plugin load order and changing plugin list into an array as discussed above.
1. standardising and implementing consistent "fill addresses": every fill should receive component name, path and markdown block index (if applicable), so that a plugin could be configured to only work on certain components/markdowns. Maybe even going as far as putting this into standard plugin config.
1. making a decision on what data and methods should be available to plugins, and whether all plugins should receive the same data regardless of their slot. For example, should props data ("usage") be available to toolbar buttons? I think it should be possible to pass this data "inside" a plugin by exporting two fills, one for `docsTabButton` and one for `exampleToolbarButton` and then sharing it between them, but is this a good way? Can it be done better?
1. coming up with a way to change the styles of rsg-components like Playground from plugins. Consider a plugin that switches Playground background color from plain white to striped, or that switches its width from 100% to 320px to simulate mobile view. A hacky way to do this would be via DOM manipulation. A better way would be to give plugins some sort of an API to do that, maybe by passing refs?
1. considering sharing the "parent rsg-component" to plugins, so that they have easy access to its props and state, and can modify it by setState. One interesting possibility here is if we do this and rewrite rsg-components so that their styles are in their state, then modifying their styles from plugins will be very easy.
1. improving plugin loader so that class components (`class Something extends Component`) can be used as fills. Currently this is not supported (only simple functions that export JSX are allowed), or I couldn't understand how to make it work.

sapegin:
> considering plugin load order and changing plugin list into an array as discussed above.

Feel free to do it if you have time! ;-)

> every fill should receive component name, path and markdown block index (if applicable), so that a plugin could be configured to only work on certain components/markdowns. Maybe even going as far as putting this into standard plugin config.

Yup, right now they receive some random data ;-|

> making a decision on what data and methods should be available to plugins, and whether all plugins should receive the same data regardless of their slot.

> For example, should props data ("usage") be available to toolbar buttons?

If it makes implementation simpler why not?

> I think it should be possible to pass this data "inside" a plugin by exporting two fills, one for docsTabButton and one for exampleToolbarButton and then sharing it between them, but is this a good way? Could it be done better?

I think we should try to implement some useful plugins and then we’ll see what work and what doesn’t.

> coming up with a way to change the styles of rsg-components like Playground from plugins. Consider a plugin that switches Playground background color from plain white to striped, or that switches its width from 100% to 320px to simulate mobile view. A hacky way to do this would be via DOM manipulation. A better way would be to give plugins some sort of an API to do that, maybe by passing refs?

I think JSS dynamic properties is a way to go here. We’ll need to pass more data to some component and make plugins be able to pass styles.

> considering sharing the "parent rsg-component" to plugins, so that they have easy access to its props and state, and can modify it by setState.

Something like that will be necessary. I’d rather pass `state` and `setState` implicitly than the whole `this`.

> improving plugin loader so that class components (class Something extends Component) can be used as fills. Currently this is not supported (only simple functions that export JSX are allowed), or I couldn't understand how to make it work.

It supposed to work already, I’ve probably missed something ;-)

n1313:
> It supposed to work already

It expects `render` [to be a function](https://github.com/styleguidist/react-styleguidist/pull/505/files#diff-cca5424950fd6a01dc641abd32827197R23), so it complains if I export a component there. How should it be done?

sapegin:
But this is the check for plugins itself, not for a fill. I don’t see any checks for fills.

n1313:
...right. Huh. I guess I'm doing something wrong somewhere else.

sapegin:
It may be me I just don’t see where yet ;-)

n1313:
What do you think about swapping the order of arguments in plugin constructor? Now it is `globalConfig, pluginConfig`, I suggest `pluginConfig, globalConfig`. I think most people will be interested in `pluginConfig` and not `globalConfig`.

sapegin:
Maybe, I don’t have a strong preference ;-)

n1313:
@sapegin, I've added two commits to change the order of arguments and change the object into array.

scripts/schemas/config.js
```diff
@@ -105,13 +105,13 @@ module.exports = {
 		default: 'base16-light',
 	},
 	plugins: {
-		type: 'object',
-		default: {},
+		type: 'array',
+		default: [],
 		process: val =>
 			CORE_PLUGINS.reduce((allPlugins, plugin) => {
```
sapegin: I think now just `concat` should be enough ;-)

sapegin:
screen recording 2017-06-27 at 10 44 pm

Plugin code:

```js
import React from 'react';
import MdBorderAll from 'react-icons/lib/md/border-all';
import MdBorderClear from 'react-icons/lib/md/border-clear';
import ToolbarButton from 'rsg-components/ToolbarButton';

const BackgroundButton = ({ state, setState }) =>
	state.__backgroundIsEnabled
		? <ToolbarButton onClick={() => setState({ __backgroundIsEnabled: false })} title="Show normal background">
		<MdBorderAll />
	</ToolbarButton>
		: <ToolbarButton onClick={() => setState({ __backgroundIsEnabled: true })} title="Show transparent background">
		<MdBorderClear />
	</ToolbarButton>;

const backgroundPlugin = () => ({
	fills: [
		{
			type: 'exampleToolbarButton',
			render: BackgroundButton,
		},
	],
});

export default backgroundPlugin;
```

And style guide config:

```js
const path = require('path');
module.exports = {
	styles: {
		Playground: {
			preview: {
				background: props => {
					console.log(props);
					return props.__backgroundIsEnabled ? 'deepskyblue' : 'transparent';
				},
			},
		},
	},
	plugins: [
		path.resolve(__dirname, 'src/plugin.js'),
	],
};
```

There’s no way to define styles from plugins yet. I think we should (at least for start) reuse the config somehow if you don’t have better ideas ;-)

image 2017-06-27 at 10 50 48 pm

n1313:
I would like the "background plugin" to be less rigid and more configurable. I wanted to allow passing arbitrary CSS via plugin configuration. At least the color/gradient should be configurable.

> There’s no way to define styles from plugins yet

Can we add/remove custom class names to components and assign custom CSS properties to them instead?

n1313:
@sapegin what are your plans for maintaining plugins? Do you want them to be kept and maintained by individual owners, or do you want them as repos inside the styleguidist org, or be pushed into the main repo?

sapegin:
> I would like the "background plugin" to be more configurable. I wanted to allow passing arbitrary CSS via plugin configuration. At least the color/gradient values should be externally configurable.

This will be easy to do I think.

> Can we add/remove custom class names to components and assign custom CSS properties to them instead?

Why do you want to do that? ;-) What’s the benefit and what use cases can’t be implemented using dynamic properties?

> Do you want them to be kept and maintained by individual owners, or do you want them as repos inside the styleguidist org

Both would work. Background plugin sounds useful enough to have it in the main organization.

n1313:
> What’s the benefit and what use cases can’t be implemented using dynamic properties?

The benefit of having a plugin like that is that we don't have to make a new separate plugin and separate API every time someone comes up with a new idea about changing styles of playgrounds. Currently we have two ideas: changing the background color and changing the width of the container to simulate narrower screens. If we can pass arbitrary styles then we need only one plugin, otherwise two.

sapegin:
I’m afraid it will be hard to implement without much benefit. But you can easily generate functions from user styles inside the plugin, and this is simpler than writing class name changing logic.

n1313:
If a ref to the element DOM node could be exposed to the plugin then manipulating class names would be trivial with element.classList APIs.

sapegin:
It won’t solve the problem of multiple plugins defining the same CSS properties and you can already do this with dynamic properties. I still don’t see what use cases aren’t covered by them. I’d suggest to try dynamic properties and see what’s really missed.

sapegin:
Accessing DOM in React is a bad practice in general and would like to avoid that if possible.

n1313:
Right now I can't come up with a realistic example of two plugins wanting to change the same CSS property on the same Playground element for different reason, but that doesn't mean that this will never happen, and we might want to be prepared for that just in case. But, on the other hand, the responsibility of proper cleanup and general compatibility between plugins ultimately lies with authors of plugins, so I suppose this question doesn't have to be resolved here right now.

sapegin:
I just don’t see much difference between JSS merging and CSS class names here: in both cases you’ll have only one value in the end ;-) And I’d say merging is more predictable — we may tweak it or even show a warning when several plugins tries to change the same property.

Anyway let’s try to make it work first and then make it perfect ;-)

scripts/schemas/config.js
```diff
@@ -93,6 +109,15 @@ module.exports = {
 	logger: {
 		type: 'object',
 	},
+	plugins: {
+		type: 'array',
+		default: [],
+		process: val =>
+			CORE_PLUGINS.reduce((allPlugins, plugin) => {
```
okonet: `val.concat(CORE_PLUGINS)`?

src/utils/loadPlugins.js
```diff
@@ -0,0 +1,45 @@
...
+		exampleTab: [],
+		docsTabButton: [],
+		docsTab: [],
+	};
+
+	plugins.forEach(({ module, options }) => {
+
+		const func = module.default || module;
+		if (!isFunction(func)) {
+			throw Error('Plugin not a function'); // TODO
```
okonet: `Plugin is not a function` is more correct although still too cryptic. But I see the TODO so hopefully it will be changed before the release :)

sapegin: Treat this file pseudocode for now ;-)

oyeanuj:
@sapegin For some of the questions mentioned in the thread above, if you are still looking for ideas, I'd suggest [Slate Editor's plugin system](https://docs.slatejs.org/slate-react/plugins). Having used it extensively, I can vouch that it is well thought-out and implemented.

sapegin:
I’ve merged the master branch here, going to add a few more things and release with 6.1.0. It will be very alpha but you’ll be able to implement some very basic plugins.

sapegin:
I’ve also fixed the CI accidentally ;-)

isratc:
hiya how can  getaccess to this repository  so I merge this pull request? 

Thankyou kindly

sapegin:
@isratc That’s an interesting way of managing open source ;-)

I’m planning this for [6.1.0](https://github.com/styleguidist/react-styleguidist/milestone/8) which depends on several PRs so I can’t say anything more that ”soonish”. Could you share your use case? Is there any missing features? It’s very very limited for now. In any case you can try to write your plugin using this branch — your feedback will be very valuable.

isratc:
Ha I am very new to github =( and a newbie dev

Basically I'm doing some experimental learning, and im trying to create a project which will also have a styleguide using your lovely styleguidist, I am using https://ant.design/ = antd library to create the components quickly. 

Issue is that is errors in the styleguide config when I would like to add 

`plugins: {
   'antd': true,
 }

or 
 {
  "plugins": [
    ["import", { libraryName: "antd", style: "css" }] 
  ]
}`

the components arent outputting in the styleguide
screen shot 2017-11-15 at 14 57 23

@sapegin  Thankyou in advance, really appreciate the help :)

sapegin:
@isratc I’m not sure what you’re trying to do and how is it related to Styleguidist plugins API. Please open a new issue and make a demo project that we can debug based on this repo: https://github.com/styleguidist/example

jspears:
You might consider using the plugin system from MrBuilder
[mr-builder](https://mr-builder.github.io/#mrbuilder-optionsmanager) .  It has babel style plugins, but also (optional) autoinstall of plugins, command line configuration, env configuration and config file configuration.     It was designed to be a generic plugin system, and is also used in the next version of [subschema](https://github.com/jspears/subschema-devel/tree/master/plugins/subschema-plugin-autoloader) to allow for injection of components.

tuchk4:
@sapegin any plans to merge it?

sapegin:
I'll need some help to make it ready, even as an MVP.

mehrad77:
Oh :( sad to see this MR is closed

PR #509, opened 2017-06-23
Isolate search placeholder styles

### LINKED ISSUE #491: input::placeholder selector breaks display in Chrome 55
Some elements of UI are looking weird in Chrome 55 (Windows) because of this [`input::placeholder`](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/TableOfContents/TableOfContentsRenderer.js#L28) selector. It should either be vendor-prefixed or separated from the rest.

n1313:
Fixes #491


sapegin:
Cool, thanks!

PR #511, opened 2017-06-23
feat(*): allow to pass settings for markdown code viewer

stepancar:
https://github.com/styleguidist/react-styleguidist/issues/257
1) render preview for jsx, js, javascript by default
2) allow to pass settings for code viewer
```
    ```jsx // { "static": true, "wrapInIframe": true }
    <span></span>
    ```
```

### DIFF
--- a/loaders/utils/chunkify.js
+++ b/loaders/utils/chunkify.js
@@ -14,16 +14,35 @@ const CODE_PLACEHOLDER = '<%{#code#}%>';
  */
 module.exports = function chunkify(markdown) {
 	const codeChunks = [];
+	const codeLanguages = ['javascript', 'js', 'jsx'];
 
 	/*
 	 * - Highlight code in fenced code blocks with defined language (```html) if the language is not `example`.
-	 * - Extract indented and fenced code blocks without language or if language is `example`.
+	 * - Extract indented and fenced code blocks with lang javascript | js | jsx or if language is `example`.
 	 * - Leave all other Markdown or HTML as is.
 	 */
 	function processCode() {
 		return ast => {
 			visit(ast, 'code', node => {
-				if (node.lang && node.lang !== 'example') {
+				let lang = node.lang || '';
+				let settings;
+				try {
+					const settingsString = lang.slice(lang.indexOf('{'), lang.lastIndexOf('}') + 1) || '{}';
+					settings = JSON.parse(settingsString);
+				} catch (exception) {
+					node.value = `Settings not parsed! Use JSON to pass settings!
+						\`\`\`jsx // { "static": false }
+							...
+						\`\`\`
+					`;
+				}
+				lang = lang.slice(0, lang.indexOf(' ') !== -1 ? lang.indexOf(' ') : lang.length);
+				node.lang = lang;
+				if (
+					lang &&
+					lang !== 'example' &&
+					(codeLanguages.indexOf(lang) === -1 || (settings && settings.static))
+				) {
 					let highlighted;
 					try {
 						highlighted = hljs.highlight(node.lang, node.value).value;
@@ -32,7 +51,11 @@ module.exports = function chunkify(markdown) {
 					}
 					node.value = highlighted;
 				} else {
-					codeChunks.push(node.value);
+					codeChunks.push({
+						type: 'code',
+						content: node.value,
+						settings,
+					});
 					node.type = 'html';
 					node.value = CODE_PLACEHOLDER;
 				}
@@ -54,10 +77,7 @@ module.exports = function chunkify(markdown) {
 		}
 		const code = codeChunks.shift();
 		if (code) {
-			chunks.push({
-				type: 'code',
-				content: code,
-			});
+			chunks.push(code);
 		}
 	});
 
--- a/loaders/utils/chunkify.js
+++ b/loaders/utils/chunkify.js
@@ -15,7 +15,7 @@ const CODE_PLACEHOLDER = '<%{#code#}%>';
 module.exports = function chunkify(markdown) {
 	const codeChunks = [];
 	const codeLanguages = ['javascript', 'js', 'jsx'];
-
+	const settingsModifiers = ['static', 'noEditor'];
 	/*
 	 * - Highlight code in fenced code blocks with defined language (```html) if the language is not `example`.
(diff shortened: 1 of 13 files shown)


sapegin:
Wow, that’s cool! Do we really need JSON here? List of space separated keywords would be enough probably.

stepancar:
@sapegin,
1) in our case we want to wrap PreviewRenderer with custom,  for allow to switch demo to mobile view and it actual only for several component examples.
For this case we need something like this:
```
    ```jsx // { "mobile": true }
     <MyComponent />
    ```
```
2) For several examples we need to show code by default, in another we don't need it.
For this case we need something like this:
```
    ```jsx // { "showCode": true }
     <MyComponent />
    ```
```
3) Several components support themes for this components we render additional buttons - theme switcher
We need any mechanism to pass example settings to PreviewComponent

sapegin:
For all your use cases a space-separated list of modifier would be enough ;-)

stepancar:
@sapegin 
```
    ```jsx // { frame: { width: '300px', height: '800' }, themeSwitcher: {themes: ['white', 'black'] }}
    ```
```

sapegin:
Then I’d implement both because even a single modifier covers most of the use cases and for anything extra complicated you could use JSON.

stepancar:
@sapegin, ok. I will implement both

stepancar:
@sapegin,
1) added support for static, noEditor single modifiers
2) allow to override default settings for showCode option

src/rsg-components/Playground/Playground.js
```diff
@@ -33,6 +44,21 @@ export default class Playground extends Component {
...
 
+	getChildContext() {
+		if (this.props.settings && this.props.settings.noEditor) {
+			const slots = this.context.slots || {};
+			return {
+				...this.context,
+				slots: {
+					...slots,
+					exampleTabButtons: (slots.exampleTabButtons || [])
+						.filter(slot => slot.id !== 'rsg-code-editor'),
```
sapegin: No editor means no chrome at all: no tabs, toolbar or border around a component. To use components to render parts of a style guide, not to show them.

loaders/utils/chunkify.js
```diff
@@ -14,16 +14,42 @@ const CODE_PLACEHOLDER = '<%{#code#}%>';
...
+						\`\`\`jsx // static noEditor
+							...
+						\`\`\`
+					`;
+				}
+				lang = lang.slice(0, lang.indexOf(' ') !== -1 ? lang.indexOf(' ') : lang.length);
+				node.lang = lang;
+				if (
+					lang &&
+					lang !== 'example' &&
```
sapegin: I think we don’t need this anymore, it’s a braking change anyway.

loaders/utils/chunkify.js
```diff
@@ -14,16 +14,42 @@ const CODE_PLACEHOLDER = '<%{#code#}%>';
...
+	 * - Extract indented and fenced code blocks with lang javascript | js | jsx or if language is `example`.
 	 * - Leave all other Markdown or HTML as is.
 	 */
 	function processCode() {
 		return ast => {
 			visit(ast, 'code', node => {
-				if (node.lang && node.lang !== 'example') {
+				let lang = node.lang || '';
+				let settings = {};
+				try {
```
sapegin: I’d simplify this part: if a string starts from `{`, try to parse the whole string as JSON, otherwise do `split(' '). We’ll also need custom modifiers for plugins.

loaders/utils/chunkify.js
stepancar: @sapegin this comment relates to empty lang or to 'example'?

sapegin: To `example`.

stepancar:
@sapegin, updated

loaders/utils/chunkify.js
stepancar: ok, fixed

sapegin:
Awesome, thanks! PhantomJS tests are failing because of some JS error — could you take a look? See the Travis report for details.

stepancar:
@sapegin , fixed js errors.
Added examples for static and noEditor modifiers. Could you please correct example text If it is necessary?

sapegin:
Thank you very much, this is super cool! 🍕

I’ll mere it into a new branch, tweak a bit and think what other breaking changes we could include into the next major release.

okonet:
I'm wondering if we could use something like front matter YAML format for that: https://jekyllrb.com/docs/frontmatter/ or even mixed thing: https://github.com/dworthen/js-yaml-front-matter

sapegin:
1. That was my first idea when I saw JSON, but no ;-) I don’t want to bring YAML parse to the client.
2. For most use cases space-separated modifiers would be enough. External files would be the most common exception I guess.
3. Looks like Markdown parsers don’t render anything you write after language tag on the same line. At least it doesn’t affect GitHub in any way so it doesn’t break browsing docs on GitHub.

@stepancar Looks like we don’t really need `//` and it would work for any language, not just JS.

stepancar:
@sapegin , removed //

stepancar:
@sapegin , maybe I can help you in release of this feature?

sapegin:
That would be awesome!! But it’s a breaking change and a major release. And I’m not 100% sure what I want to include into it so I’ve created an issue to track it — #517 ;-)

n1313:
Could you guys please clarify what is going to be the default mode for tagged/untagged code blocks when this is released?

sapegin:
js/jsx/javascript will be rendered as interactive examples, the same way as code examples without the language now.

okonet:
Hmm, but there will be a way to disable that, right?

n1313:
And what about untagged ones?

sapegin:
@okonet Why do you want to disable that? :-) `static` modifier will render highlighted code.

@n1313 No changes.

okonet:
@sapegin I assume if I use `import` etc it won't work?

sapegin:
I’m not sure what do you mean here ;-) We’re changing only the default behavior — you’ll have more options, not less.

sapegin:
If you use `import` in interactive examples it won’t work *today*, no changes here ;-)

okonet:
Yes, that's why I would disable it and make it non-interactive.

sapegin:
Yup, that’s why we have the `static` modifier ;-)

PR #516, opened 2017-06-27
Feat: Allow devServer.watchConfig to be configured

### LINKED ISSUE #515: Support Vagrant with VirtualBox
Hi,

I just tried to make styleguidist work with Vagrant, and it mostly works. There is a pretty deal-breaking issue with hot reloading, however.

When using the VirtualBox provider (which most people do because it's cross-platform) in Vagrant, you need to set `devServer: { poll: true }` for it to work because file system events do not work properly. See https://webpack.js.org/guides/development-vagrant/. This is not possible to set in styleguidist because of the way the server config is merged. See [create-server.js](https://github.com/styleguidist/react-styleguidist/blob/master/scripts/create-server.js#L15-L17).

I tried patching the file to include `poll: true`, and then it works just fine. It seems trivial to support this usecase by changing the config merging to the following:

```js
const webpackDevServerConfig = Object.assign({}, webpackConfig.devServer, {
		noInfo: true,
		compress: true,
		clientLogLevel: 'none',
		hot: true,
		quiet: true,
		watchOptions: Object.assign({}, {
			ignored: /node_modules/,
		}, webpackConfig.devServer ? webpackConfig.devServer.watchOptions : {}),
		contentBase: config.assetsDir,
		stats: webpackConfig.stats || {},
});
```

Shall I make a PR for this, or do you propose another solution?

esphen:
## Why

Currently, Vagrant with the VirtualBox provider does not work with styleguidist, as there are issues with webpack's default behavior. The way around this is to make webpack use a polling scheme for checking for changed files. This is currently not configurable in styleguidist.

## How

This PR makes the `devServer.watchConfig` configurable so that it is possible to use styleguidist with Vagrant

Fixes #515

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -390,6 +390,20 @@ initialState = {
 </div>
 ```
 
+## How to use Vagrant with Styleguidist?
+
+First of all, make sure you read [this guide](https://webpack.js.org/guides/development-vagrant/) from the webpack documentation.
+
+Then enable polling in your webpack config:
+
+```js
+devServer: {
+  watchOptions: {
+    poll: true,
+  },
+},
+```
+
 ## Are there any other projects like this?
 
 * [Atellier](https://github.com/scup/atellier), a React components emulator.
--- a/scripts/create-server.js
+++ b/scripts/create-server.js
@@ -2,11 +2,12 @@
 
 const webpack = require('webpack');
 const WebpackDevServer = require('webpack-dev-server');
+const merge = require('webpack-merge');
 const makeWebpackConfig = require('./make-webpack-config');
 
 module.exports = function createServer(config, env) {
 	const webpackConfig = makeWebpackConfig(config, env);
-	const webpackDevServerConfig = Object.assign({}, webpackConfig.devServer, {
+	const webpackDevServerConfig = merge(webpackConfig.devServer, {
 		noInfo: true,
 		compress: true,
 		clientLogLevel: 'none',


scripts/create-server.js
```diff
@@ -2,22 +2,29 @@
...
+				ignored: /node_modules/,
+			},
+			contentBase: config.assetsDir,
+			stats: webpackConfig.stats || {},
 		},
-		contentBase: config.assetsDir,
-		stats: webpackConfig.stats || {},
-	});
+		{
+			watchOptions: webpackConfig.devServer ? webpackConfig.devServer.watchOptions : {},
```
sapegin: I think this part isn’t required because webpack-merge should do recursive merge.

docs/Cookbook.md
```diff
@@ -390,6 +390,22 @@ initialState = {
 </div>
 ```
 
+## How to use Vagrant with styleguidist?
```
sapegin: styleguidist → Styleguidist

docs/Cookbook.md
```diff
@@ -390,6 +390,22 @@ initialState = {
...
+
+First of all, make sure you read [this guide](https://webpack.js.org/guides/development-vagrant/) from the webpack documentation.
+
+The article mentions that you need to use `--watch-poll` when using the
```
sapegin: I’d write it shorter like this: 

> Then enable polling in your webpack config:

sapegin:
And thanks for the pull request! 🍕

scripts/create-server.js
esphen: I'm sorry, not quite sure what you mean here. Would you like to merge the entire `webpackConfig.devServer` into the styleguidist dev server config, not just `watchConfig`? Here I am checking for `devServer` being defined in the user config so it doesn't throw when it isn't.

I have not used `webpack-merge` before, so I am assuming it works like `Object.assign` in most cases

sapegin: I’d try to remove this whole second block and see if it still works because webpack-merge does recursive merge and it should work ;-)

esphen: Ah, I see! Thanks for the ELI5. It does seems to work without the last block because of the recursive merging.

esphen:
Amended the commit with your requested changes

Issue #517, opened 2017-06-27
6.0.0 release


sapegin:
It’s all about Markdown examples improvements:

- [x] Modifiers for fenced code examples (#257, #511)
- [x] Default modifiers: `static` and `noeditor`
- [x] Ability to add custom modifiers via style guide config (#269)
- [x] Make modifiers case-insensitive
- [x] Update docs

Optional:

- [x] Remove webpack 1 support (#512)

Not sure about the last one because it turned out to be a lot of work. If anyone wants to help with that — we can do it, otherwise I’d postpone it to the 7.0.0.

**All work will be in the [six](https://github.com/styleguidist/react-styleguidist/tree/six) branch.**

n1313:
Why the major bump?

sapegin:
Because we’re changing default behavior for fenced code blocks with js/jsx/javascript language.

n1313:
How much time do you think it'll take this branch to be fully done, and would it be possible to coincide it with initial offering of plugin API?

sapegin:
Depends on (1) how many people will want to help, (2) will we include webpack removal (I guess not ;-) and (3) what we want to include in initial plugins API.

I guess it’s possible to have *some* undocumented plugin API ;-)

stepancar:
@sapegin  , what reason to "Make modifiers case-insensitive"?

stepancar:
"Ability to add custom modifiers via style guide config" 
Just merge modifiers from styleguide.config with passed to example in .md?

sapegin:
I would ask for a reason to make it case-sensitive ;-) It improves the UI — users don’t have to remember the correct casing. But I think it’d be enough to make space-separated modifiers case-insensitive.

> Just merge modifiers from styleguide.config with passed to example in .md?
> what we should implement?

I mean some kind of filters or transformers that could be defined in the config (and later in plugins) that would allow users to implement thing like this file loader. Something like this could work:

```js
function({ contents, lang, modifiers }) {
  if (typeof modifiers.file === 'string') {
    const filepath = modifiers.file;
    delete modifiers.file;
    return {
       contents: fs.readFileSync(filepath),
       modifiers,
       lang,
  }
}
```

stepancar:
@sapegin, I dont like idea with case-insensitive. Because if implement something like loop on example settings object keys and override all keys with lowercase in example plugins(read as preprocessors) developers will should work with case insensitive parameters. 
```
[...]

Issue #269, opened 2016-12-27
Ability to display source code of the component


okonet:
During my work on [react-dropzone styleguide](http://react-dropzone.netlify.com/), which isn't a conventional styleguide, one of the missing pieces is an ability to inline the source code of the examples into the styleguide.

So, in my case, the config looks like this (excerpt):

```js
module.exports = {
  title: 'react-dropzone',
  showCode: false,
  sections: [
    {
      name: 'Installation',
      content: 'README.md'
    },
    {
      name: 'PropTypes',
      components: './src/index.js'
    },
    {
      name: 'Examples',
      components: './examples/**/*.js',
      context: {
        Dropzone: './src/index'
      }
    }
  ]
};
``` 

The idea is to display examples and allow opening them in the isolation mode. But one of the important things I'd like to achieve is to have example's code inlined into the documentation. Right now this is only possible by having it twice — once in `/examples/**/index.js` and one more time in `/examples/**/Readme.md` inside the code block with `jsx` type (so it won't get evaluated).

It would be great to have an option for each `section` that would allow inlining the source code of component's file to the styleguide.

Worth mentioning, that I don't want this to be editable, just a raw contents rendered with code highlight. This should be easy to achieve with `raw-loader`. Although it might be interesting to experiment with editable source code as well.

okonet:
Another idea would be to have some special markup for the markdown that would add the source code of the component as a code block below it.

okonet:
Random file? I'm not sure I get this.

sapegin:
Instead of rendering only component‘s source, which looks very specific and rare, just allow to specify a path.

okonet:
Yes, that would work as well!

sapegin:
[](./file.js)

okonet:
After some 1x1 discussion, it seems that the most appropriate and flexible way would be to have some extended markdown syntax that would allow inlining the contents of the linked file. So, the user would write something like `> ../path/to/my/file.js` and it would render file contents using code fence block.

There are few open questions like:

1. What this marker should look like?
1. How to determine the syntax of the file? Should user define it?
1. It would be great if those links would still work on GitHub as well

okonet:
One of the possible solutions could be as simple as having a fence block with the path to a file that should be inlined.

![2016-12-27 at 11 43](https://cloud.githubusercontent.com/assets/11071/21498086/c3411724-cc29-11e6-90a9-7f737b9af659.png)

[...]

PR #525, opened 2017-07-04
examples preprocessor and case insensitive settings for example

stepancar:
https://github.com/styleguidist/react-styleguidist/issues/269
https://github.com/styleguidist/react-styleguidist/issues/517

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -183,18 +183,23 @@ module.exports = {
 ```javascript
 module.exports = {
   examplePreprocessor: function(props) {
-      const lang = props.lang;
-      const content = props.content;
-      const settings = props.settings
-      if (typeof settings.file === 'string') {
+    const lang = props.lang;
+    const content = props.content;
+    const settings = props.settings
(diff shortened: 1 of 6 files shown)


stepancar:
@sapegin , added examplesPreprocessor setting to config which allow to override/customize examples

sapegin:
Awesome, thanks! I don’t like the name though, but don’t have any better ideas now. Will think ;-)

sapegin:
Thanks! We could change the name later, still have not better ideas ;-|

stepancar:
Why you don't like this name?

sapegin:
“Preprocessor” is way to complex word for a simple function ;-)

stepancar:
hm)
I exampleReplacer, exampleOverrider, exampleUpdater?
or just
```
getExample({lang, content, settings})=>
    ({
        lang,
        content,
        settings
    })
```

sapegin:
I’d try something like `updateExample`. Better matches other settings.

stepancar:
@sapegin updarted in six branch

PR #528, opened 2017-07-05
React-docgen version upgraded to v3 to support Flow annotations

danielfx90:
# Problem
`<*>` tokens throw syntax error. For example when doing:
```
import type { Children } from "react";
...
type Props = {
   children?: Children<*>,
...
```

# Solution
`react-docgen` package version upgraded from v2 to v3: https://github.com/reactjs/react-docgen/issues/181


package.json
```diff
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "ast-types": "^0.9.11",
+    "babylon": "^7.0.0-beta.8",
```
sapegin: Babylon is a dependency of react-docgen, why is it here?

package.json
```diff
@@ -42,6 +43,7 @@
     "es6-object-assign": "~1.1.0",
     "es6-promise": "^4.1.0",
     "escodegen": "^1.8.1",
+    "eslint": "^3.19.0",
```
sapegin: Should stay in devDependencies.

package.json
danielfx90: Sorry, I was trying something and I forgot to remove it. Thanks!

package.json
danielfx90: True that!

danielfx90:
@sapegin all done!

danielfx90:
@sapegin Sorry, still getting the hang of it. Now all three checks have passed.

sapegin:
Merged, thanks!

sapegin:
Small issue with this change: https://github.com/nerdlabs/react-docgen-displayname-handler/blob/master/package.json#L43

```
npm ERR! peer dep missing: react-docgen@2.x, required by react-docgen-displayname-handler@1.0.0
```

Not sure if it breaks anything, if not probably worth sending a PR to react-docgen-displayname-handler to relax required react-docgen version.

danielfx90:
I'll check it out!

PR #530, opened 2017-07-06
Output correct protocol when starting RSG server

n1313:
This PR changes "Style guide server started at:" notification to correctly state https protocol when starting RSG server with `{ devServer: { https: true } }`.

### DIFF
--- a/bin/styleguidist.js
+++ b/bin/styleguidist.js
@@ -106,7 +106,7 @@ function commandServer() {
 		if (err) {
 			console.error(err);
 		} else {
-			const isHttps = config.webpackConfig && config.webpackConfig.devServer && config.webpackConfig.devServer.https;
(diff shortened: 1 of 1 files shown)


bin/styleguidist.js
```diff
@@ -106,8 +106,9 @@ function commandServer() {
 		if (err) {
 			console.error(err);
 		} else {
+			const isHttps = config.webpackConfig && config.webpackConfig.devServer && config.webpackConfig.devServer.https;
```
sapegin: I think it’s better to use `compiler.options` (see a few lines below) — it contains an actual webpack config, so you at least can avoid `config.webpackConfig` check.

n1313: Done.

sapegin:
Thanks!!

PR #532, opened 2017-07-06
Dynamically update document.title when switching to isolated mode

### LINKED ISSUE #531: Display component name in <title> when opening a component in isolated mode
I spend most of my time with RSG switching between a handful of components in isolated mode. Unfortunately, page title doesn't change when I do that, so my browser history consists of the same page name repeated many times. Not very helpful. I would like my page `<title>` to display component name when I open it in isolated mode: `React Style Guide Example - Button` or something similar.

n1313:
Fixes #531

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -64,6 +64,14 @@ function renderStyleguide() {
 	slugger.reset();
 	sections = setSlugs(sections);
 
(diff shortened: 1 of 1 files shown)


sapegin:
Thanks! 🍕

PR #535, opened 2017-07-07
Feat: Improve shape defaultProp rendering (#524)

### LINKED ISSUE #524: Improve printing of objects as `defaultProps`
Hi,

Before continuing, I would just like to say that I love react-styleguidist. it's an amazing product, very well put together. We evaluated many styleguide-libraries, but yours was by far the better in terms of producing something that was easy for the team to consume while working.

On to the issue at hand. My suggestion is quite simply to improve the rendering of `defaultProps` that are shapes. In the image below, you see that in this very simple example, it is breaking over several lines in a not very appealing way due to RSG falling back to rendering it as a string.

screen shot 2017-07-03 at 21 26 43

In this example, it's not **that** bad - but in another project I am working on, I would like to pass an object with several large strings and many properties, and it looks absolutely awful. The cell ends up taking too much space, and is generally unappealing - so we opted for not using `defaultProps` for our configuration objects for now. That's not to say that we don't **want** to.

What I propose is that the `defaultProps` field is rendered as the string `shape` or `Object` when an object is passed, and then on hover, focus or click, a popover is opened next to the string with the pretty printed corresponding `defaultProp` object.

Is this currently possible? Do you like the idea? I looked into the code, but it seems that the default `propsParser` does not return an immediately useful `defaultValue`. It would have to be parsed for it to be pretty printed. See below:

```js
{ type: { name: 'shape', value: {} },
  required: false,
  description: '',
  defaultValue: { value: '{\n    foo: \'bar\'\n}', computed: false },
  tags: {} }
```

esphen:
# What

This PR improves the printing of objects in `defaultProps`, which had some issues in the existing version. See #524 for further details.

# How

This PR removes the fallback of printing objects as strings. Instead, objects will be parsed, beautified, then finally displayed in a title. This mirrors the existing method used by `function` props.

The improved preview

Parsing is done with the [safe-eval](https://github.com/hacksparrow/safe-eval) library. Not sure if we actually need this library, though, as we are not in a node.js context here. Wanted to get some feedback first. We could use `eval` directly, or just render the unprettified source code representation in the title. Thoughts?

Fixes #524

### DIFF
--- a/src/rsg-components/Props/PropsRenderer.js
+++ b/src/rsg-components/Props/PropsRenderer.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Group from 'react-group';
+import objectToString from 'javascript-stringify';
 import Arguments from 'rsg-components/Arguments';
 import Code from 'rsg-components/Code';
 import JsDoc from 'rsg-components/JsDoc';
@@ -85,12 +86,27 @@ function renderDefault(prop) {
 	if (prop.required) {
 		return <Text>Required</Text>;
 	} else if (prop.defaultValue) {
-		if (prop.type && prop.type.name === 'func') {
-			return (
-				<Text underlined title={showSpaces(unquote(prop.defaultValue.value))}>
-					Function
-				</Text>
-			);
+		if (prop.type) {
+			const propName = prop.type.name;
+
+			if (propName === 'func') {
+				return (
+					<Text underlined title={showSpaces(unquote(prop.defaultValue.value))}>
+						Function
+					</Text>
+				);
+			} else if (propName === 'shape' || propName === 'object') {
+				// We eval source code to be able to format the defaultProp here. This
+				// can be considered safe, as it is the source code that is evaled,
+				// which is from a known source and safe by default
+				// eslint-disable-next-line no-eval
+				const object = eval(`(${prop.defaultValue.value})`);
+				return (
+					<Text underlined title={objectToString(object, null, 2)}>
+						Shape
+					</Text>
+				);
+			}
 		}
 
 		return (
--- a/examples/sections/styleguide.config.js
+++ b/examples/sections/styleguide.config.js
@@ -31,8 +31,8 @@ module.exports = {
 			env === 'development'
 				? false
 				: {
-						maxAssetSize: 670000, // bytes
-						maxEntrypointSize: 670000, // bytes
+						maxAssetSize: 680000, // bytes
+						maxEntrypointSize: 680000, // bytes
 						hints: 'error',
 					},
 	}),
--- a/package-lock.json
+++ b/package-lock.json
@@ -249,9 +249,9 @@
       "dev": true
     },
     "ansi-escapes": {
-      "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz",
-      "integrity": "sha1-06ioOzGapneTZisT52HHkRQiMG4=",
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-2.0.0.tgz",
+      "integrity": "sha1-W65SvkJIeN2Xg+iRDj/Cki6DyBs=",
       "dev": true
     },
     "ansi-html": {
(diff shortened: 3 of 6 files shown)


SaraVieira:
@esphen Will this only show on hover ?

esphen:
@SaraVieira Yes, which brings me to an issue I have with the way these popovers are handled today. Seeing as you have to hover over the element to show the data, this excludes a potentially large audience from seeing this information. To quote the [HTML spec](https://www.w3.org/TR/html/dom.html#the-title-attribute):

> Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet). 

I tried this in ChromeVox to verify, and was not able to read the text. So this solution is better for mouse users, but not not accessible, which is far from ideal. I would much prefer this text to be a `<button>` (which makes it focusable and tabbable), which shows a popover element on focus.

src/rsg-components/Props/PropsRenderer.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Group from 'react-group';
+import safeEval from 'safe-eval';
+import prettyFormat from 'pretty-format';
```
sapegin: This looks like way too many new frontend dependencies for such feature. Have you compared the bundle size?

sapegin:
> I tried this in ChromeVox to verify, and was not able to read the text. So this solution is better for mouse users, but not not accessible, which is far from ideal. I would much prefer this text to be a <button> (which makes it focusable and tabbable), which shows a popover element on focus.

Sounds like a good idea for a separate PR if we keep hover ;-)

sapegin:
> Parsing is done with the safe-eval library. Not sure if we actually need this library, though, as we are not in a node.js context here.

I don’t think so since we’re rendering source code which I’d say is safe by default ;-)

> just render the unprettified source code representation in the title. Thoughts?

This or find some very simple library to do that. Pretty-print is very nice for Node but way too big for frontend. Something like [stringify-object](https://github.com/yeoman/stringify-object) maybe.

esphen:
Thanks @sapegin. Based on your feedback I'll use `window.eval` for parsing and look into a smaller library for formatting.

esphen:
@sapegin: I amended the commit with the requested changes. 

There were a lot of changes in the `package-lock.json` after I `rm package-lock.json && npm i` however. I wonder what that's about. Looks like it pruned all the `requires` fields, apart from that it looks pretty consistent with my expectations. I used npm@5.0.0 here.

sapegin:
I would recommend to update to npm 5.2 — it’s much more stable but honestly I don’t really understand some changes it does to the lock file ;-|

sapegin:
Oops, I forget that Sindre Sorhus doesn’t care about webpack users and his modules often breaks UglifyJS because they contains some ES6 code (see Travis CI logs) ;-(

Looks we have to choose yet another library ;-|

esphen:
I found [javascript-stringify](https://github.com/blakeembrey/javascript-stringify), it looks pretty good. It has no dependencies. Pushed a new commit to the PR.

This is the result:

2017-07-14-194649_swaygrab

n1313:
@esphen @sapegin do you think this change should be applied to the function popup as well? I would be nice to have these popups look the same.

sapegin:
@n1313 I think he’s on Linux or something because it’s just a `title` ;-)

n1313:
Titles on Linux look like *that*? o__o

esphen:
@n1313 lol. Yes they do :)

esphen:
@sapegin Travis seems to fail because it is now 5K over the recommended build size. What do you think?

sapegin:
It’s fine to change it. This check is to prevent accidental adding of something huge to the bundle, like the whole Lodash ;-)

esphen:
@sapegin: I pushed another commit which fixes the travis failure and fixes a merge conflict on `package-lock.json`. Are we good now?

sapegin:
Looks good for me, thanks! Could you please add some tests too?

esphen:
@sapegin Done, the new branches in the diff now have code coverage

sapegin:
Merged, thanks!

PR #537, opened 2017-07-07
Set line-height to inherit

### LINKED ISSUE #536: Change line-height from normal to inherit
**styleguide.config.js**
```js
const path = require('path');

module.exports = {
  title: 'BSA React Web Components',
  sections: [
    {
      name: 'Introduction',
      content: 'docs/Introduction.md',
    },
    {
      name: 'Reboot',
      content: 'docs/Reboot.md',
    },
    {
      name: 'Components',
      components: 'src/**/[A-Z]*.js',
    },
  ],
  require: [
    path.join(__dirname, 'docs/assets/styles/index.less'),
  ],
  assetsDir: 'docs/assets',
  webpackConfig: {
    module: {
      loaders: [
        {
          test: /\.js$/,
          exclude: /node_modules/,
          loader: 'babel-loader',
          query: {
            presets: ['es2015', 'stage-3', 'react'],
          },
        },
        {
          test: /\.css$/,
          loaders: ['style-loader', 'css-loader'],
        },
        {
          test: /\.less$/,
          use: [
            'style-loader',
            'css-loader',
            {
              loader: 'less-loader',
              options: {
                sourceMap: true,
              },
            },
          ],
        },
      ],
    },
  },
};
```

webpack version: ^3.0.0
OS, browser version: OS X, Chrome 58.0.3029.110 (64-bit)

Hi! I like react-styleguidist a lot.

I'm trying to set up a page where I showcase how headers, lists and other elements look by importing the library (and its stylesheet).

Currently it looks like this 

screen shot 2017-07-07 at 3 40 44 pm

If I change [this div](https://github.com/styleguidist/react-styleguidist/blob/a010018f6672af81332018399fc0e595e81d4b24/src/rsg-components/Playground/PlaygroundRenderer.js#L26) `line-height` from `normal` to `inherit` it looks how I would expect it to.

screen shot 2017-07-07 at 3 43 34 pm

Could we change it so the default style for line-height is inherit?

SaraVieira:
Closes #536

### DIFF
--- a/src/styles/nonInheritedProps.js
+++ b/src/styles/nonInheritedProps.js
@@ -57,5 +57,5 @@ export default {
 	'unicode-bidi': 'normal',
 	'vertical-align': 'baseline',
 	'width': 'auto',
-	'z-index': 'auto'
+	'z-index': 'auto',
 };
--- a/src/styles/nonInheritedProps.js
+++ b/src/styles/nonInheritedProps.js
@@ -29,7 +29,6 @@ export default {
 	'height': 'auto',
 	'hyphens': 'none',
 	'left': 'auto',
-	'line-height': 'inherit',
 	'margin': '0',
 	'max-height': 'none',
 	'max-width': 'none',
--- a/src/styles/nonInheritedProps.js
+++ b/src/styles/nonInheritedProps.js
@@ -29,6 +29,7 @@ export default {
 	'height': 'auto',
 	'hyphens': 'none',
 	'left': 'auto',
+	'line-height': 'inherit',
 	'margin': '0',
 	'max-height': 'none',
 	'max-width': 'none',
@@ -57,5 +58,5 @@ export default {
 	'unicode-bidi': 'normal',
 	'vertical-align': 'baseline',
 	'width': 'auto',
-	'z-index': 'auto',
+	'z-index': 'auto'
 };
--- a/src/styles/setupjss.js
+++ b/src/styles/setupjss.js
@@ -20,7 +20,7 @@ jss.setup({
 
 				// Allow inheritance because it may be set on body and should be available for user components
 				fontFamily: 'inherit',
-				'line-height': 'inherit',
+				lineHeight: 'inherit',
 			},
 		}),
 		nested(),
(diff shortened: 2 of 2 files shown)


SaraVieira:
@sapegin Changed ! Thanks , it makes more sense there :)

sapegin:
I hope it won’t break the UI ;-)

src/styles/nonInheritedProps.js
```diff
@@ -57,5 +57,5 @@ export default {
 	'unicode-bidi': 'normal',
 	'vertical-align': 'baseline',
 	'width': 'auto',
-	'z-index': 'auto',
+	'z-index': 'auto'
```
sapegin: This change is weird ;-)

SaraVieira:
From what I checked it didn't seen to break :)

A good idea maybe would be to put something like netifly on PR's always mounting one of the instances so we can always be sure :)

cc @sapegin

sapegin:
@SaraVieira Could you take a look at my last comment? Is it Prettier or some accidental change?

sapegin:
> A good idea maybe would be to put something like netifly on PR's always mounting one of the instances so we can always be sure :)

Yeah, we use Netlify for the site anyway ;-)

src/styles/nonInheritedProps.js
SaraVieira: I thought I answered this 😢 

I think this was prettier should I remove it ?

SaraVieira:
@sapegin Can I do this ? Never used any GH integration so it would be a cool experience :)

sapegin:
@SaraVieira You can at least think how to do it, I’m not really sure myself ;-) Could you open a new issue?

SaraVieira:
Sure! As soon as I get home :)

src/styles/nonInheritedProps.js
sapegin: I was wandering why Prettier didn’t put it back ;-) Better to revert it ;-)

SaraVieira:
@sapegin All changes made and creating the issue now

src/styles/setupjss.js
```diff
@@ -20,6 +20,7 @@ jss.setup({
 
 				// Allow inheritance because it may be set on body and should be available for user components
 				fontFamily: 'inherit',
+				'line-height': 'inherit',
```
sapegin: Sorry, just noticed: it should be `lineHeight` of course.

SaraVieira: Also fixed 👍

PR #539, opened 2017-07-10
Create a jss instance for react-styleguidist

timsnadden:
This helps to ensure that config/class names don't collide if components being presented by react-styleguidist also use jss.


sapegin:
Thanks for the pull request! Please check test failures.

PR #540, opened 2017-07-11
Fix: Preview should unmount the Wrapper component when it unmounts

wenboyu2:
`Preview` doesn't unmount `Wrapper` it mounts when `Preview` is unmounted. As a result, clicking `IsolateButton`, all `Wrapper`s remain on the DOM including those irrelevant.

In this PR `Preview` is changed to clean up after itself.


sapegin:
Thanks for the pull request!

PR #542, opened 2017-07-12
Fix duplicating of keys for 'arrayOf' prop type in the unions rendering

psdcoder:
We get an error if we have in the component propTypes something like this: 
```
items: PropTypes.oneOfType([
  PropTypes.arrayOf(PropTypes.object),
  PropTypes.arrayOf(PropTypes.array),
]).isRequired,
```
 image
It happens because at the moment styleguidist use just type of prop (`arrayOf`), and when we have two or more same types in the `oneOfType`- type can't be used as key.
So this PR just adds an index as a suffix to key.

### DIFF
--- a/src/rsg-components/Props/PropsRenderer.js
+++ b/src/rsg-components/Props/PropsRenderer.js
@@ -156,8 +156,8 @@ function renderUnion(prop) {
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


sapegin:
Cool, thanks!

Issue #519, opened 2017-06-28
Run console.clear() when components are reloaded


birjolaxew:
It's a bit irritating to see an error in your console and not know if it's from the current component that was hot-loaded in or from an older version.

This would be solved by running `console.clear()` once a component has been reloaded (or if multiple components are being reloaded, after they're all reloaded).

sapegin:
Sounds like a good feature. Feel free to send a pull request ;-)

ethanh1223:
@sapegin Brand new to open-source and trying to get my first few commits with you guys, so bear with me :)

I've played with a few options and clearing the console seems to work best in the PlaygroundRenderer. When re-compiling multiple components, errors from prior compilations are cleared but errors from _all_ components in current compilation are visible. 

Any objections to this method? Tried to submit a PR but linter doesn't allow for "console" statements in the Component :(

sapegin:
@ethanh1223 Awesome! 🍕

I think [Preview](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Preview/Preview.js) would be a better place.

For ESLint you need to disable this check for a specific line:

```js
console.clear(); // eslint-disable-line no-console
```

Conscious use of `console.log` is OK, ESLint guards us from committing accidental debug code ;-)

wayspurrchen:
Is it possible to disable this functionality on-demand? I want to try to develop against components directly within my styleguidist server, and runtime errors that I want to diagnose and fix are cleared whenever I refresh the page.

PR #544, opened 2017-07-17
Clear console on hot reload

ethanh1223:
Adds #519

Clearing console in `ComponentDidMount` of Preview makes a lot more sense than Playground -- now understand how the two are related :)

Didn't think a doc update was necessary here.

### DIFF
--- a/src/rsg-components/Preview/Preview.js
+++ b/src/rsg-components/Preview/Preview.js
@@ -49,6 +49,7 @@ export default class Preview extends Component {
 	};
 
 	componentDidMount() {
+		console.clear(); // eslint-disable-line no-console
 		this.executeCode();
 	}
 
--- a/src/rsg-components/Preview/Preview.js
+++ b/src/rsg-components/Preview/Preview.js
@@ -49,7 +49,9 @@ export default class Preview extends Component {
 	};
 
 	componentDidMount() {
-		console.clear(); // eslint-disable-line no-console
+		if (console) {
+			console.clear(); // eslint-disable-line no-console
+		}
 		this.executeCode();
 	}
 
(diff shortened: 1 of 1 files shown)


src/rsg-components/Preview/Preview.js
```diff
@@ -49,6 +49,9 @@ export default class Preview extends Component {
 	};
 
 	componentDidMount() {
+		if (console.clear) { // eslint-disable-line no-console
```
sapegin: Why do we need this condition?

birjolaxew: I believe it is because the tests [fail with a `console.clear is not a function`](https://travis-ci.org/styleguidist/react-styleguidist/jobs/254586655#L3112) if it isn't included.

sapegin: Then it should be polyfilled in `test/jestsetup.js`, not hacked around in the main code.

tizmagik: You can also mock/spy on that function within the test itself, if you wanted:

```js
global.console.clear = jest.fn();
```

Then you can assert that `console.clear` was called 😁

ndelangen:
Hey @ethanh1223 it looks like this PR kind of got forgotten, do you think it would be possible for you to finish this one?

It would definitely be awesome if you could. If not I'm sure @sapegin can take over to get this work merged!

sapegin:
@ndelangen What’s your interest here? ;-) 

@ethanh1223 Let me know if you need any help. I would try what @tizmagik is suggesting. If you don’t have time to finish the PR soon — let us know, we’ll find someone else.

sapegin:
Finished in #638.

Issue #545, opened 2017-07-18
Preview not updating when editing initialState.


kumarabhishek:
When changing initialState in code editor, its not updating the preview.
For example in code editor of Button let's write following:
```js
initialState={v: 'Hello'};
<Button>{state.v}</Button>
```
Now when we edit code to following, it does not reflect in Button preview:

```
initialState={v: 'Hello1'};
<Button>{state.v}</Button>
```

kumarabhishek:
It seems I do not have permission to push. Getting following error:

```bash
~/dev/react-styleguidist $ git push origin preview-fix
ERROR: Permission to styleguidist/react-styleguidist.git denied to kumarabhishek.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

```

kumarabhishek:
@sapegin I got fix for https://github.com/styleguidist/react-styleguidist/issues/519 also. If you provide me privilege, i will include even this fix in same PR.

n1313:
@kumarabhishek please fork the repo, push your fix to a branch in your fork and send a PR from your fork to the main repo.

sapegin:
@kumarabhishek Thanks! There’s a PR for #519 already: #544.

PR #546, opened 2017-07-19
fix(Preview): Preview not updating when initialState is edited in code-editor

kumarabhishek:
Fixing #545 - Preview not updating when initialState is edited in code-editor

### DIFF
--- a/src/rsg-components/Preview/Preview.js
+++ b/src/rsg-components/Preview/Preview.js
@@ -95,8 +95,8 @@ export default class Preview extends Component {
 		);
 
 		window.requestAnimationFrame(() => {
+			this.unmountPreview();
 			try {
-				this.unmountPreview();
 				ReactDOM.render(wrappedComponent, this.mountNode);
 			} catch (err) {
 				this.handleError(err);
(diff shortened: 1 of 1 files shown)


src/rsg-components/Preview/Preview.js
```diff
@@ -96,6 +96,7 @@ export default class Preview extends Component {
 
 		window.requestAnimationFrame(() => {
 			try {
+				this.unmountPreview();
```
sapegin: It shouldn’t be inside try / catch.

kumarabhishek: If I put this outside try, then in case of exception this.unmountPreview(); will be executed twice (one in this method and other in handleError). Also by having it inside try does not harm in anyway. Its my proposition though. Let me know if you still want it outside try, I will do the change and update PR.

sapegin: That’s fine if it’s executed twice (it has a condition inside anyway) but having it inside try / catch makes understanding of the code harder because we use try / catch here to catch errors in user’s code.

kumarabhishek: Oh ho. Now i got it. Basically its call stack in catch handler you want to keep clean and only related to user code. Yeah, you are right. I will update the code. Thanks for clarifying.

kumarabhishek:
Incorporated review comment.

sapegin:
Thanks!

PR #550, opened 2017-07-26
Remove typo from testing section of the developer guide

natearn:
I found a small typo in the filename. `.spece.js` should be `.spec.js`.


sapegin:
Thanks!

PR #552, opened 2017-07-28
Update nvmrc to use node 8

esphen:
Since you are using `package-lock`, I wanted to suggest updating the `.nvmrc` to reflect this. As you know, and note in the `dangerfile-js`, `package-lock.json` requires npm v5. This is not reflected in the the `.nvmrc`, which states node version 5, which equals npm version 3.

```
$ nvm use 5
Now using node v5.12.0 (npm v3.8.6)
$ nvm use 6
Now using node v6.7.0 (npm v3.10.3)
$ nvm use 7
Now using node v7.10.0 (npm v4.2.0)
$ nvm use 8
Now using node v8.2.1 (npm v5.3.0)
```


sapegin:
Thanks! I don’t use nvm myself so I always forget to update it ;-)

PR #556, opened 2017-07-31
Render null and undefined defaultValues as code

esphen:
Hi,

I found another thing I found to be lacking with `defaultProps`. In the case where you specify a default value that is `null` or `undefined`, I would expect it to show that value. Instead, what you get is a title with that value, see below.

2017-07-31-192516_swaygrab

Here's a PR that makes an exception  for `null` and `undefined` values, and renders them as code instead.

### DIFF
--- a/src/rsg-components/Props/PropsRenderer.js
+++ b/src/rsg-components/Props/PropsRenderer.js
@@ -82,6 +82,8 @@ function renderShape(props) {
 	return rows;
 }
 
+const defaultValueBlacklist = ['null', 'undefined'];
+
 function renderDefault(prop) {
 	if (prop.required) {
 		return <Text>Required</Text>;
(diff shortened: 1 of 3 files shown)


src/rsg-components/Props/PropsRenderer.js
```diff
@@ -82,14 +82,22 @@ function renderShape(props) {
...
+
 function renderDefault(prop) {
 	if (prop.required) {
 		return <Text>Required</Text>;
 	} else if (prop.defaultValue) {
 		if (prop.type) {
 			const propName = prop.type.name;
 
-			if (propName === 'func') {
+			if (defaultValueBlacklist.includes(prop.defaultValue.value)) {
```
sapegin: You need to use `indexOf` because Node 4 doesn’t support `includes`.

esphen: Thanks, I'm too spoiled by babel

esphen:
@sapegin I amended the commit with your requested change

src/rsg-components/Props/PropsRenderer.js
sapegin: Babel won’t transpile this without a runtime / polyfill / whatever, will it?

sapegin:
Thanks, merged!

Issue #551, opened 2017-07-27
Doc is not getting updated when changes are not from react component file


pasupuletics:
@sapegin , I have a scenario that react component imports `propTypes` from external file and doc is not getting generated which is expected, And I have written a [react-docgen](https://github.com/reactjs/react-docgen) custom handler, which would generate a doc for external `propTypes` and everything went smooth. Now the issue with live reload, When I made any changes to external `propTypes` file changes are reflecting on live example, but documentation is not getting updated as as react-docgen [parse](https://github.com/styleguidist/react-styleguidist/blob/master/loaders/props-loader.js#L28) method is not invoking(props-loader.js) unless changes are from actual react component file. Is there any way to invoke [parse](https://github.com/styleguidist/react-styleguidist/blob/master/loaders/props-loader.js#L28) method(props-loader.js) when a change is not form actual react component file?

##### Example:
```javascript
import React from 'react';
import propTypes from './button-proptypes'

export default class Button extends React.component{
...
}

Button.propTypes = propTypes;
```
##### button-proptypes.js
```javascript
import PropTypes from 'prop-types';

export const propTypes = {
/*Type of a button*/
propA: PropTypes.oneOf(['simple', 'primary'])
propB: ...
}
```

When I made any changes to `button-proptypes.js` live reload is working fine, but documentation is not getting Updated.

sapegin:
Maybe `contextDependencies` could help here: https://react-styleguidist.js.org/docs/configuration#contextdependencies

pasupuletics:
@sapegin , Thanks for your quick response, But configuring `contextDependencies` didn't serve my purpose. But I got the root cause that the issue with webpack HMR(not really an issue, its an expected behaviour). HMR has enabled for all the files and working as expected. When component file changes  webpack re-builds and pushes updated file into browser same happens for external imported files as well. 

Is there any possibility to update doc/props property when any of the component dependent file gets changed? I mean when react component gets updated webpack invokes `props-loader` and pushes changes to browser like wise, can we invoke the same for component dependent files too?

sapegin:
Could you try to implement it? I don’t really understand how it should work ;-|

pasupuletics:
Sure @sapegin will give it a try. Could you please clarify me on below:
When component file changes how the props-loader.js is pitching in(might be because of HMR), could you please help me to find where that peace of code reside(actual logic).

pasupuletics:
[...]

PR #557, opened 2017-08-01
Added support to enable HMR for component dependencies

pasupuletics:
Added support to re-compile component (invoking props-loader) when any of its dependencies gets changed. This will enable only when [`contextDependencies`](https://react-styleguidist.js.org/docs/configuration#contextdependencies) is configured.

Issued in favor of this [discussion](https://github.com/styleguidist/react-styleguidist/issues/551)


pasupuletics:
Added test case and pushed changes to upstream, @sapegin

sapegin:
Thanks, merged!

Issue #195, opened 2016-10-12
Improve single-component usage


sapegin:
React Styleguidist could be awesome for documenting components, [like this](https://sapegin.github.io/react-group/), but now it’s not.

* [ ] New option to disable sidebar.
* [ ] New option, `getComponentName()`, to change the title.
* [ ] “Fork me” ribbon with customisable URL.
* [ ] New option to hide isolated link.
* [ ] Hide path line if it’s empty.

Anything else? Any feedback?

natearn:
Is this change still desired? I'd like to contribute and I'm looking for an easy place to start.

sapegin:
@natearn Yes, please do it! Feel free to ping me on Gitter if you have any questions ;-)

snmslavk:
@sapegin Hi, where do you want to add this "fork me" ribbon?

> Hide path line if it’s empty.

I think it's done, could we make this checkbox checked?

sapegin:
@snmslavk I think a new option in style guide config would work:

```js
ribbon: {
  url: 'http://example.com/',
  text: 'Fork me on GitHub', // Default if the ribbon option passed
  color: '#fff', // Default if the ribbon option passed
  background: '#24292e' // Default if the ribbon option passed
}
```

You can use this CSS as a starting point:

https://github.com/sapegin/social-likes-next/blob/7018e8d51c05853eb04b437d8fce7bc0a34a663e/gh-pages/gh-pages.styl#L97-L123

![image 2017-08-30 at 2 31 34 pm](https://user-images.githubusercontent.com/70067/29872436-f418dcce-8d8f-11e7-9b3d-4425834845d6.png)

M-ZubairAhmed:
@sapegin are all tasks done? if not i would like to pick one

sapegin:
@M-ZubairAhmed Only checked ones are done. And there’s an abandoned PR for the ribbon.

kizu:
What I would love to have is to have some way to configure example both per example (maybe in a way the `static` keyword is done? i.e. in the first line after the language), and per section (so we could have one style of example in one part of docs, and another — at another).

Things I would like to configure:

1. The default state of the example: if the code is visible or folded. Right now we can control this option only globally.
2. To have some layouts to choose from: the current one, with full-width code, the side-by-side (see how it looks like there: http://kizu.ru/bemto-components/#html-structure), or in some other way.
[...]

PR #558, opened 2017-08-01
Hide empty path line

natearn:
This is an implementation of one of the features from #195.

The Pathline component will not render if the path line is empty. I'm not sure if this requires an update to the docs.

Please let me know if anything needs changing.

### DIFF
--- a/src/rsg-components/ReactComponent/ReactComponentRenderer.js
+++ b/src/rsg-components/ReactComponent/ReactComponentRenderer.js
@@ -37,9 +37,10 @@ export function ReactComponentRenderer({
 		<div className={classes.root} id={name + '-container'}>
 			<header className={classes.header}>
 				{heading}
-				<Pathline>
-					{pathLine}
-				</Pathline>
+				{(pathLine || null) &&
+					<Pathline>
+						{pathLine}
+					</Pathline>}
 			</header>
 			{(description || docs) &&
 				<div className={classes.docs}>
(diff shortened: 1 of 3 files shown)


src/rsg-components/ReactComponent/ReactComponentRenderer.js
```diff
@@ -37,9 +37,10 @@ export function ReactComponentRenderer({
 		<div className={classes.root} id={name + '-container'}>
 			<header className={classes.header}>
 				{heading}
-				<Pathline>
-					{pathLine}
-				</Pathline>
+				{(pathLine || null) &&
```
sapegin: Just `pathLine &&` should be enough. And we should remove `isRequired` from PropTypes.

natearn: The `null` is so to tell React not to render anything. Otherwise, an empty string will still get rendered as an empty string.

sapegin: Why do you care about that? ;-) `!!pathLine &&` would be still more readable.

natearn: Fair enough :)

natearn:
Alright, I've implemented the changes and fixed the snapshot.

sapegin:
Thanks, merged!

Issue #456, opened 2017-05-26
Using devtools with react-stylegudist


benedyktdryl:
Hi, 

I know that `devtools` are ignored along with some other configuration keys.

**My question is:**

- Is there any particular reason for that? Is it breaking `react-stylegudist` entirely or there are some edge cases?
- Can we disable ignoring somehow or you will consider PR with disabling it optionally feature?
- Is there kind of recipe for having sourcemaps on development when using your library? 

Thanks for your time!

sapegin:
For `devtools` it should not be necessary, so feel free to send a pull request ;-)

benedyktdryl:
So I should go with optional disabling ignoring or I can just remove devtools from ignored keys list?

sapegin:
@n1313 Did you want to be able to change `devtools`? What should we do for production build?

n1313:
Hmmm, no, I don't think devtools matters for what I do.

sapegin:
@n1313 Sorry, it was @cef62 ;-)

benedyktdryl:
Any updates on that? 

CC: @cef62 @sapegin

sapegin:
I think it should explicitly disable `devtool` in production, otherwise it would complicate style guide configuration and will be easy to miss.

PR #569, opened 2017-08-09
Using devtools with react-stylegudist #456

AttackTheDarkness:
The `devtool` setting in your webpack config will now get passed through for non-production environments, e.g., running the hot-reloading server.

It will still get ignored for production builds.

Docs updated, and added a test.

### DIFF
--- a/scripts/utils/__tests__/mergeWebpackConfig.spec.js
+++ b/scripts/utils/__tests__/mergeWebpackConfig.spec.js
@@ -39,12 +39,20 @@ it('should ignore certain Webpack plugins', () => {
 	expect(result.plugins[1].constructor.name).toBe('MyPlugin');
 });
 
-[
-	{ env: 'development', action: 'pass', expected: 'source-map' },
-	{ env: 'production', action: 'ignore', expected: false },
-].forEach(t => {
-	it(`should ${t.action} devtool settings in ${t.env}`, () => {
-		const result = mergeWebpackConfig({ devtool: false }, () => ({ devtool: 'source-map' }), t.env);
-		expect(result).toEqual({ devtool: t.expected });
-	});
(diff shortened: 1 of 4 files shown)


scripts/utils/__tests__/mergeWebpackConfig.spec.js
```diff
@@ -38,3 +38,13 @@ it('should ignore certain Webpack plugins', () => {
 	expect(result.plugins[0].constructor.name).toBe('UglifyJsPlugin');
 	expect(result.plugins[1].constructor.name).toBe('MyPlugin');
 });
+
+[
```
sapegin: It would be almost shorter and much more readable if you write two separate tests without the loop ;-)

scripts/utils/mergeWebpackConfig.js
```diff
@@ -46,6 +38,11 @@ const merge = mergeBase({
  */
 module.exports = function mergeWebpackConfig(baseConfig, userConfig, env) {
 	const userConfigObject = isFunction(userConfig) ? userConfig(env) : userConfig;
-	const safeUserConfig = omit(userConfigObject, IGNORE_SECTIONS);
+	let safeUserConfig = omit(userConfigObject, IGNORE_SECTIONS);
+	// For production builds, we'll ignore devtool settings to avoid
+	// source mapping bloat.
+	if (env === 'production') {
+		safeUserConfig = omit(safeUserConfig, ['devtool']);
```
sapegin: I’d define a new const for that, like IGNORE_SECTIONS_PROD for consistency. Or even:

```js
const IGNORE_SECTIONS_ENV = {
  development: [],
  production: ['devtool']
}
```

In this case you could avoid a condition.

sapegin:
Thanks, merged!

PR #577, opened 2017-08-16
Change require to show absolute path

skube:
As a best practice one should use absolute paths for Webpack stuff

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -239,7 +239,7 @@ Modules that are required for your style guide. Useful for third-party styles or
 module.exports = {
   require: [
     'babel-polyfill',
-    '/absolute/path/to/styles.css',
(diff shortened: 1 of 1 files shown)


docs/Configuration.md
```diff
@@ -239,7 +239,7 @@ Modules that are required for your style guide. Useful for third-party styles or
 module.exports = {
   require: [
     'babel-polyfill',
-    'path/to/styles.css',
+    '/absolute/path/to/styles.css',
```
sapegin: Would be better to show how to easily have an absolute path without hardcoding it:

```
path.join(__dirname, 'styleguide/styles.css'),
```

skube: Ok. Good idea!

sapegin:
Thanks, merging!

skube:
Thank you! 🥇

PR #581, opened 2017-08-16
Update GettingStarted.md with actual packages required

csmosx:
It's impossible to get a truly clean npm install due to the conflicting react-docgen dependencies, but at least this is closer to sanity.


sapegin:
This is 100% incorrect. Looks like you have some non-default npm settings — unsatisfied peer dependency should be a warning, not an error. Please open an issue with more details.

Issue #584, opened 2017-08-19
CSS EM unit is not isolate


papasmurf17:
In my react components I use `EM` as font size unit, and when the components are rendered all font sizes are wrong.

I know that `EM` is relative unit and if the container changes font-size all childs inherit this change, but the final aim of react-styleguidist is:

> Isolated React component development environment

My workaround to bypass this problem was add this piece of configuration to the **styleguide.config.js**

```js
styles: {
    Examples: {
      root: {
        'font-size': '1em !important'
      }
    },
    Playground: {
      root: {
        'font-size': '1em !important'
      },
      preview: {
        'font-size': '1em !important'
      }
    },
    ReactComponent: {
      root: {
        'font-size': '1em !important'
      }
    },
    Section: {
      root: {
        'font-size': '1em !important'
      }
    },
    Sections: {
      root: {
        'font-size': '1em !important'
      }
    },
    StyleGuide: {
      root: {
        'font-size': '1em !important'
      },
      content: {
        'font-size': '1em !important'
      },
      hasSidebar: {
        'font-size': '1em !important'
      }
    },
```

sapegin:
Please use fenced blocks with language tags (for example ```` ```js````) to make code examples more readable.

sapegin:
Looks like we need to add font-size here:

https://github.com/styleguidist/react-styleguidist/blob/master/src/styles/setupjss.js#L21-L23

Feel free to send a pull request for that ;-)

PR #585, opened 2017-08-20
Fix: CSS EM unit is not isolate (#584)

papasmurf17:
I added font-size to jss configuration, I couldn't find any tests for jss conf.

I test the Fix with my react components library and it works fine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,7 +2,6 @@
   "name": "react-styleguidist",
   "version": "0.0.0-development",
   "lockfileVersion": 1,
-  "requires": true,
   "dependencies": {
     "@semantic-release/commit-analyzer": {
       "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -2,7 +2,6 @@
   "name": "react-styleguidist",
   "version": "0.0.0-development",
   "lockfileVersion": 1,
-  "requires": true,
```
sapegin: Hm, why would npm changes that? Could you please remove this file from the PR?

papasmurf17: The requires field has been added since npm [5.1.0](https://github.com/npm/npm/releases/tag/v5.1.0)
> This has a handful of fixes:
>- It introduces a new package-lock.json field, called requires, which tracks which modules a given module requires.
>- It fixes #16839 which was caused by not having this information available, particularly when git dependencies were involved.
>- It fixes #16866, allowing the package.json to trump the package-lock.json.
>- It stops calling fetch-package-metadata.addBundled on fake children (this was harmless, because bundleDependencies would never be set, but also pointless.)
>- npm ls now loads the shrinkwrap, which opens the door to showing a full tree of dependencies even when nothing is yet installed. (It doesn't do that yet though.)

sapegin: OK, leave it then but it’s even more weird now — I’m using npm 5.3 ;-)

PR #588, opened 2017-08-21
Omit component if propsParser returns an empty array

suevalov:
First of all, thanks for great project!

I started using `react-styleguidist` with Typescript and `react-typescript-docgen` and have found one really annoying bug. 

If there is no exported component in `.ts` file, `parse` from `react-typescript-docgen` return an empty array.

```ts
import * as React from 'react';

export interface MyComponentProps {
	foo: string;
	className?: string;
}

const MyComponent: React.SFC<MyComponentProps> = () => (...);

export default MyComponent;
```

It has to be ```export const MyComponent```, but during development you can forgot about this.

`props-loader.js` reaches line 45 and assumes that array is not empty. Then props[0], which is `undefined`, is passed to `getProps` function and `react-styleguidist` stops working. So, you have to add export to the component and start it again.

### DIFF
--- a/loaders/props-loader.js
+++ b/loaders/props-loader.js
@@ -10,6 +10,8 @@ const getExamples = require('./utils/getExamples');
 const getProps = require('./utils/getProps');
 const consts = require('../scripts/consts');
 
+const ERROR_MISSING_DEFINITION = 'No suitable component definition found.';
+
 module.exports = function(source) {
 	const file = this.request.split('!').pop();
 	const config = this._styleguidist;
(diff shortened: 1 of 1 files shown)


sapegin:
Shouldn’t we print a warning in this case?

suevalov:
@sapegin yep, it does make sense.

Technically, it's the same as `Error: No suitable component definition found` and the message should be the same as in `catch` block.

I can move `isArray` to `try` and throw an error if it's empty array. In this case `catch` block will handle warning printing. What do you think?

sapegin:
Let’s try it ;-)

suevalov:
Done!

sapegin:
Awesome, thanks! Merged ;-)

Issue #600, opened 2017-09-04
Deeply nested section titles are all h2


kidkuro:
#554 addressed the issue of nested sections being tagged as primary, but a section heading still only has the concept of primary or not.

These means that all section nested 2 or more levels deep all get `h2` for a heading. Sections headings should nest with `h1`, `h2`, `h3`, `h4`, etc.

For example, modifying the sections example component to have these sections:

```js
module.exports = {
  title: 'React Style Guide Example',
  sections: [
    {
      name: 'Principles',
      sections: [
        {
          name: 'Colour',
          sections: [
            {
              name: 'Colour palette',
            },
          ],
        },
      ],
    },
    {
      name: 'Components',
      sections: [
        {
          name: 'Buttons',
          components: () => [
            './src/components/Button/Button.js',
            './src/components/RandomButton/RandomButton.js',
          ],
        },
      ],
    },
  ],
}
```

Results in this layout:

![image](https://user-images.githubusercontent.com/1942003/30009096-b651325e-9179-11e7-8a3a-90f0597e4edd.png)

kidkuro:
A possible solution to this would be to replace the `primary` prop everywhere with a `depth` prop.

kidkuro:
Created PR #601 with a `depth` prop.

PR #601, opened 2017-09-04
Fix: Add depth to sections so heading levels are correct (#600)

kidkuro:
Added a `depth` prop to section heading and pass down through all sections.

Left in the `root` and `primary` props for now in case anyone is using them in a customised styleguide, but they should be deprecated and removed.

Also updated the sections example to demonstrate deeply nested sections.

### DIFF
--- a/examples/sections/styleguide.config.js
+++ b/examples/sections/styleguide.config.js
@@ -4,19 +4,39 @@ module.exports = {
 	title: 'React Style Guide Example',
 	sections: [
 		{
-			name: 'Components',
-			components: './src/components/**/[A-Z]*.js',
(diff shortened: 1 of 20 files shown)


sapegin:
> Left in the root and primary props for now in case anyone is using them in a customised styleguide, but they should be deprecated and removed.

We don’t really guarantee backwards compatibility here, otherwise it would make any changes in Styleguidist UI almost impossible. But for most common use cases we provide config options.

kidkuro:
Updated to remove primary, and reset depth to be 1-based

sapegin:
Thanks, merged!

PR #604, opened 2017-09-11
Update GettingStarted.md

gbhasha:
Removed the unnecessary '+' char, so that it's easy to copy paste the code.


sapegin:
It’s not unnecessary, it makes diffs easier to read. Copy paste issue should be fixed on the rendering part somehow.

PR #617, opened 2017-09-30
Fix navigating back from isolated / focus mode

BTMPL:
Current implementation ends up linking to the base url (`window.location.pathname`) which does not include the `#` sign so it will perform an HTTP request.

This PR changes the behavior - it will now navigate to `#/`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -75,6 +75,8 @@ function renderStyleguide() {
 	}
 	document.title = documentTitle;
 
+	// If the current hash location was set to just `/` (e.g. when navigating back from isolated view to overview)
+	// replace the URL with one without hash, to present the user with a single address of the overview screen
 	const hash = location.hash.slice(1);
 	if (hash === '/') {
 		const url = window.location.pathname + window.location.search;
--- a/src/index.js
+++ b/src/index.js
@@ -75,6 +75,12 @@ function renderStyleguide() {
 	}
 	document.title = documentTitle;
 
+	const hash = location.hash.slice(1);
+	if (hash === '/') {
+		const url = window.location.pathname + window.location.search;
+		history.replaceState('', document.title, url);
+	}
+
 	ReactDOM.render(
 		<StyleGuide
 			codeRevision={codeRevision}
--- a/src/rsg-components/slots/IsolateButton.js
+++ b/src/rsg-components/slots/IsolateButton.js
@@ -7,7 +7,7 @@ import { getUrl } from '../../utils/utils';
 
 const IsolateButton = ({ name, example, isolated }) =>
 	isolated ? (
-		<ToolbarButton href={getUrl()} title="Show all components">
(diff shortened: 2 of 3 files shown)


sapegin:
Thanks for the pull request! Is there a way to remove this `#/` from the address bar?

BTMPL:
No, I don't think so, that's basically how it is supposed to work when using hash routing (see react-router HashRouter).

You could try adding something like

```
window.history.replaceState(undefined, undefined, './')
```

as a callback or something to remove the extra `#/`, but not sure if this is really a big issue.

sapegin:
It is a UX issue because we now have two URLs for the home page: one is normal and one is ugly without any reason from the user point of view.

BTMPL:
Ok, I could get around that with

```diff
-               <ToolbarButton href={getUrl({ anchor: true, slug: '/' })} title="Show all components">
+               <ToolbarButton onClick={() => {
+                       window.location = './#/';
+                       window.history.replaceState(undefined, undefined, './');
+               }}title="Show all components">
```
but now I'm not sure if this is OK, as we're limiting accessibility (it will render as a `<button>`); there's no way to render an `<a>` with an callback / `onClick`

BTMPL:
Can of course use `getUrl` to grab the valid URLs, just wanted to quickly demonstrate the idea.

sapegin:
Thanks, I’ll checkout your branch and see what can we do here. I’m wondering how React Router works because they somehow prevent real navigation and just render another component…

BTMPL:
react-router uses the BrowserRouter which hooks into HTML history API and is effective only when you use their `Link` component (so a normal `a href` would still trigger an HTTP request).

As for HashRouter - I *think* they redirect you from `/` to `/#/` right away.

sapegin:
They still render `Link` as `a` but I assume it has `onClick` handler that prevent navigation.

Actually I’ve solved the same issue here: https://github.com/sapegin/hashnav/blob/master/hashnav.js:

1. Listen `hashchange`.
2. If URL is `#/` then do `history.replaceState('', document.title, '/')`.

BTMPL:
Sure, the same logic could be implemented here but it would require bigger changes to the one I proposed. 

And yes, react-router uses onClick handlers.

sapegin:
Would you mind trying something similar to what I did in HashNav? If not I’ll merge the PR as is and will try to finish it later.

BTMPL:
Sure, I'll try to check out how to resolve this in the next 24h

sapegin:
Awesome, thanks! 🍕

BTMPL:
Since there's already a hashChange listener in `index.js` (the one that re-renders the whole app) I added the logic in there.

Since using `history.replaceState` does not trigger another `hashChange` event to be emitted I've placed it right before the rendered.

sapegin:
Thanks! Could you please add a comment explaining what we’re doing here (the hash replacement thing) and I’ll merge it.

sapegin:
Awesome, thanks!

PR #618, opened 2017-09-30
Update Enzyme to 3.0

lutien:
Updating Enzyme and enzyme-to-json to the last versions with an adapter for react 15. 
Tests are updated to work with new versions.
Should help to migrate tests to react 16.


package.json
```diff
@@ -41,6 +41,7 @@
     "copy-webpack-plugin": "^4.0.1",
     "css-loader": "^0.28.7",
     "doctrine": "^2.0.0",
+    "enzyme-adapter-react-15": "^1.0.0",
```
sapegin: Should be in devDependencies.

src/rsg-components/Preview/Preview.spec.js
```diff
@@ -38,7 +38,10 @@ it('should not not fail when Wrapper wasn’t mounted', () => {
 });
 
 it('should render component renderer', () => {
-	const actual = shallow(<Preview code={code} evalInContext={evalInContext} />, options);
+	const actual = shallow(
+		<Preview code={code} evalInContext={evalInContext} />,
+		Object.assign({}, options, { disableLifecycleMethods: true })
```
sapegin: Why do we need `disableLifecycleMethods` here? What was changed in Enzyme?

lutien: Before in shallow rendering lifecycle methods didn`t work by default. Now they do

sapegin: 👍

package.json
lutien: Yes, true. Fixed.

lutien:
In a migration guide said that getElement in shallow rendering is getNode method from previous version.
http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#with-shallow-getnode-should-be-replaced-with-getelement
It means this is a documentation for getElement - http://airbnb.io/enzyme/docs/api/ShallowWrapper/getNode.md
And this is for getDOMNode http://airbnb.io/enzyme/docs/api/ReactWrapper/getDOMNode.md

sapegin:
I see, make sense!

src/rsg-components/Markdown/__snapshots__/Markdown.spec.js.snap
```diff
@@ -1,19 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Markdown should render Markdown in span in inline mode 1`] = `"<span class=\\"rsg--base-0\\">Hello <em class=\\"rsg--em-16 rsg--base-0\\">world</em>!</span>"`;
```
sapegin: Could you take a look why these wrappers have disappeared?

src/rsg-components/Playground/Playground.spec.js
```diff
@@ -78,7 +79,6 @@ it('should open a code editor', done => {
 
 	setTimeout(() => {
 		expect(actual.find(reactCodeMirrorSelector)).toHaveLength(1);
-		done();
```
sapegin: And why `done` was removed? This test is still async. Looks like I’m missing something here :-)

src/rsg-components/Markdown/__snapshots__/Markdown.spec.js.snap
lutien: So, I've investigated a bit. The reason for it you can find here - https://github.com/airbnb/enzyme/issues/1162
The short answer is that it happened because of breaking changes in cheerio which is used for render in enzyme. But we actually can get these wrappers back if instead of doing this
`expect(render(<Test />).html()).toMatchSnapshot();`
we`ll do this
`import cheerio from 'cheerio';  expect(cheerio.html(render(<Test />))).toMatchSnapshot();`

src/rsg-components/Playground/Playground.spec.js
lutien: I thought with `expect.assertions(1);` it would work as with promises, but no :(
Test failed before but the reason was not here, now it should work properly.

src/rsg-components/Markdown/__snapshots__/Markdown.spec.js.snap
sapegin: Doesn’t look very nice but I think it would make sense to use this because we want to change the wrapper too.

src/rsg-components/Playground/Playground.spec.js
sapegin: `expect.assertions` just tells Jest how many assertions to expect and it would fail the test after a timeout if there were less assertions.

src/rsg-components/Markdown/__snapshots__/Markdown.spec.js.snap
lutien: And done

sapegin:
Awesome, thanks! 🍕

Issue #613, opened 2017-09-25
Custom React components for Markdown elements


ryanoglesby08:
I'd like to be able to configure custom components that get used when rendering Markdown documentation blocks.

For example, I have a custom `<Link>` component and would like to render that for all hyperlinks in my documentation.

Raw Markdown: MyComponent.md
"You can find more information about the inspiration for this component \[here\](http://example.com)."

Compiled
"You can find more information about the inspiration for this component &lt;Link href="http://example.com"&gt; here &lt;/Link&gt;.

I see that this is possible by completely replacing the [Markdown.js](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Markdown/Markdown.js#L73) component as described in the [Cookbook](https://react-styleguidist.js.org/docs/cookbook.html#how-to-change-the-layout-of-a-style-guide) and configuring HTML "a" tags to use my custom Link component.

Replacing the entire Markdown.js component feels dirty and dangerous and it is a very large component that has a lot of responsibilities.

I would prefer to be able to provide a custom mapping as configuration to react-styleguidist. Or if each element in Markdown.js is exposed as a separate component, I could override just the ones I want.

Have you considered these possibilities? Is replacing the Markdown.js file the way to go? I could also look at making a PR to add this capability.

Thanks for the great tool btw! I've used it on multiple projects now. :)

sapegin:
> Have you considered these possibilities? Is replacing the Markdown.js file the way to go? I could also look at making a PR to add this capability.

No, but it looks like a good idea and shouldn’t add a lot of complexity, so feel free to make a PR for that! ;-)

ryanoglesby08:
Awesome! Will work on it.

SaraVieira:
@sapegin Would be allowing a markdown parser the way to go ?

sapegin:
@SaraVieira Not sure I understand what you mean here ;-|

probablyup:
Not sure if @sapegin has it exposed, but `markdown-to-JSX` has this capability: https://github.com/probablyup/markdown-to-jsx

ryanoglesby08:
Yep @probablyup the capability is not directly exposed to the consumer of react-styleguidist quite yet. I'm working through some PRs (#650 and #651 already merged) to introduce more components into the markdown parsing so that the consumer can specify their own.

The entire list of components to create is in this PR --> #619, if anyone else wants to help out!

[...]

PR #619, opened 2017-10-01
POC: Extract Markdown links and headers as components. #613

ryanoglesby08:
... so that they can be overridden in styleguide.config.js.

### DIFF
--- a/examples/customised/src/components/Button/Readme.md
+++ b/examples/customised/src/components/Button/Readme.md
@@ -1,10 +1,10 @@
-Basic button:
+## Basic button:
 
 ```jsx
 <Button>Push Me</Button>
 ```
 
-Big pink button:
+### Big pink button:
 
 ```jsx
 <Button size="large" color="deeppink">Lick Me</Button>
--- a/examples/customised/src/link.css
+++ b/examples/customised/src/link.css
@@ -1,0 +1,7 @@
+:global .link {
+	color: salmon;
+	text-decoration: underline;
+}
+:global :link:hover {
+	text-decoration: none;
+}
--- a/examples/customised/styleguide.config.js
+++ b/examples/customised/styleguide.config.js
@@ -36,9 +36,12 @@ module.exports = {
 	},
 	// Override Styleguidist components
 	styleguideComponents: {
+		'Markdown/Link': path.join(__dirname, 'styleguide/components/Markdown/Link.js'),
+		'Markdown/Header': path.join(__dirname, 'styleguide/components/Markdown/Header.js'),
 		Logo: path.join(__dirname, 'styleguide/components/Logo'),
 		StyleGuideRenderer: path.join(__dirname, 'styleguide/components/StyleGuide'),
 	},
+	require: [path.join(__dirname, 'src/link.css')],
 	webpackConfig: {
 		module: {
 			rules: [
--- a/examples/customised/styleguide/components/Markdown/Header.css
+++ b/examples/customised/styleguide/components/Markdown/Header.css
@@ -1,0 +1,26 @@
+.h1 {
+	font-size: 2rem;
+	color: gold;
+}
+
+.h2 {
+	font-size: 1.5rem;
+	color: red;
+}
+
+.h3 {
+	font-size: 1.25rem;
+	color: cadetblue;
+}
+
+.h4 {
+	font-size: 1rem;
+}
+
+.h5 {
+	font-size: 1rem;
+}
+
+.h6 {
+	font-size: 1rem;
+}
--- a/examples/customised/styleguide/components/Markdown/Header.js
+++ b/examples/customised/styleguide/components/Markdown/Header.js
@@ -1,0 +1,59 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import s from './Header.css'
+
+function Header({ level, children }) {
+	return React.createElement(
+		`h${level}`,
+		{
+			className: s[`h${level}`],
+		},
+		children
+	);
+}
+
+Header.propTypes = {
+	level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
+	children: PropTypes.node,
+};
+
+export default {
+	h1: {
+		component: Header,
+		props: {
+			level: 1,
+		},
+	},
+	h2: {
+		component: Header,
+		props: {
+			level: 2,
+		},
+	},
+	h3: {
+		component: Header,
+		props: {
+			level: 3,
+		},
+	},
+	h4: {
+		component: Header,
+		props: {
+			level: 4,
+		},
+	},
+	h5: {
+		component: Header,
+		props: {
+			level: 5,
+		},
+	},
+	h6: {
+		component: Header,
+		props: {
+			level: 6,
+		},
+	},
+
+};
--- a/examples/customised/styleguide/components/Markdown/Link.js
+++ b/examples/customised/styleguide/components/Markdown/Link.js
@@ -1,0 +1,7 @@
+export default {
+	a: {
+		props: {
+			className: 'link',
+		},
+	},
+};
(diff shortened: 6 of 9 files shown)


ryanoglesby08:
Re: Issue #613

ryanoglesby08:
Hi @sapegin 

This is a WIP and more of a proof of concept right now. I wanted to go ahead and get a PR out there so that I can get some feedback before continuing further down this path.

My strategy is to break up the large `styles` object in "rsg-components/Markdown" and extract components for each element. I've done that for `a` and `h1`-`h6`. This will allow each Markdown element/component to be overridden using the `styleguideComponents` config option. When overriding the Markdown components, you will export a `markdown-to-jsx` config object containing the new component and any props.

I've overridden links and headers in this manner in the customised example.

This seems to be consistent with the patterns already established, flexible, allow custom theming and styles to still work, and does not introduce much extra complexity.

Let me know what you think!

sapegin:
Hey @ryanoglesby08, thank you very much for your pull request! I’m just back from React Alicante (it was amazing!) and will have a look soon ;-)

ryanoglesby08:
Re: I’d extract Header component the same way so it could be reused not only in Markdown.

Sure. I can do that eventually.

------------

Re: And I’m not sure we need to store these configs in separate files.

By "configs" I assume you mean the mappings of component + props to the corresponding tag name that gets translated by markdown-to-jsx.

I think storing the configs in separate files enables modularity and creates a more flexible solution. By storing the configs in separate files, it allows the react component library to define its own mappings per component without having to override a giant Markdown.js file.

Overriding any Markdown element with a new component requires 2 things: 1) the new component and 2) a new mapping of the Markdown element to the component and props. If I did not keep the configs separate, it means more files need to change to override a single Markdown element.

Do you have other ideas for how I might approach this?

sapegin:
> I think storing the configs in separate files enables modularity and creates a more flexible solution. By storing the configs in separate files, it allows the react component library to define its own mappings per component without having to override a giant Markdown.js file.

Do you see many use cases when people would want to define separate components for each heading level? Do you think it’s important to differentiate components (link, header, etc.) that are used in Markdown and the same components that are used in Styleguidist UI? I think using normal component (when you export a real component) increases modularity and much less surprising.

> Overriding any Markdown element with a new component requires 2 things: 1) the new component and 2) a new mapping of the Markdown element to the component and props. If I did not keep the configs separate, it means more files need to change to override a single Markdown element.

In both cases you would need to override one file: file that exports component or file that exports mapping and hides component inside.

ryanoglesby08:
Hi @sapegin. I've finally been able to work a bit more on this. I've changed the approach a little bit based on your feedback. This feels like a good approach to me. Here is the structure I've come out with so far. Let me know of your feedback.

rsg-components
|
| ->**Heading/HeadingRenderer.js**: a h1 - h6 component used throughout styleguidist, including Markdown files (Override me to replace all headings)
|
|
|-> **Markdown/Heading.js**: An adapter on top of HeadingRenderer that adds bottom margin. This is useful for the prose in Markdown, but not wanted on all headings throughout the style guide, such as the SectionHeadings. (Override me to replace how much spacing is added around headings in Markdown)
|
|-> **Markdown/Markdown.js**: Contains the mapping of all components to elements in Markdown


------- 

examples/customised

src/components
| -> **Header.js**: A custom h1 - h6 component being showcased

styleguide/components
| -> **Heading.js**: An adapter that adapts the interface of "rsg-components/Heading/HeadingRenderer" to the Header that is being showcased. Namely adapting the `depth` prop to the `level` prop, which have different types.
| -> **Link.js**: A Link component to be used in the styleguide with custom link styling. This is just an example of how to incorporate plain styles and another overridden component.

**styleguide.config.js**
—> alias `HeadingRenderer` with "styleguide/components/Heading"
—> alias `Link` with "styleguide/components/Link"

src/rsg-components/Heading/HeadingRenderer.js
```diff
@@ -0,0 +1,63 @@
...
+
+	return (
+		<Tag id={id} className={headingClasses}>
+			{children}
+		</Tag>
+	);
+}
+
+export const styles = ({ color, fontSize, fontFamily }) => ({
+	heading: {
```
sapegin: I’d compose using cx() function — would make code simpler and shorter.

src/rsg-components/Markdown/Heading.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Styled from 'rsg-components/Styled';
+
+import HeadingRenderer from 'rsg-components/Heading/HeadingRenderer';
```
sapegin: Should be `import HeadingRenderer from 'rsg-components/Heading'` — `*Render` components are implementation detail, we should’t leak to the caller code. See for details: https://react-styleguidist.js.org/docs/development.html

src/rsg-components/Heading/HeadingRenderer.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import Styled from 'rsg-components/Styled';
+
+function HeadingRenderer({ classes, depth, id, deprecated, children }) {
```
sapegin: `level` sounds better than `depth` for me.

src/rsg-components/Markdown/Heading.js
```diff
@@ -0,0 +1,29 @@
...
+
+function MarkdownHeading({ classes, depth, children }) {
+	return (
+		<div className={classes.spacing}>
```
sapegin: Maybe `Heading` component should accept `className`? And we could get rid of the div here.

src/rsg-components/SectionHeading/SectionHeadingRenderer.js
```diff
@@ -1,39 +1,36 @@
...
-	});
 	return (
-		<Tag id={id} className={classes.root}>
-			<a href={href} className={headingClasses}>
-				{children}
-			</a>
-			<div className={classes.toolbar}>{toolbar}</div>
-		</Tag>
+		<HeadingRenderer id={id} depth={depth} deprecated={deprecated}>
+			<div className={classes.root}>
```
sapegin: And get rid or this div as well ;-)

src/rsg-components/Heading/HeadingRenderer.js
ryanoglesby08: Sure can use `level`. I used `depth` because that prop name was already in place from the `SectionHeadingRenderer`, which is where this component was extracted from.

ryanoglesby08:
Also, I noticed that the Markdown.js file has styles for "input" elements. Is this necessary? There is no Markdown syntax for inputs, and I'm not sure the purpose of trying to embed inputs in Markdown documentation. 

https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Markdown/Markdown.js#L121-L125

sapegin:
Maybe better to split into two (or more PRs) if that’s convenient for you.

> Also, I noticed that the Markdown.js file has styles for "input" elements. Is this necessary?

They are for todos like in your previous comment ;-)

src/rsg-components/Heading/HeadingRenderer.js
sapegin: I see ;-) This line is funny:

```
const headingLevel = Math.min(6, depth);
```

Let’s make it `level` everywhere.

ryanoglesby08:
Cool, will break it up into multiple PRs. Will leave this one here to keep the todo list if you don't mind.

sapegin:
I’d finish this one because it looks very close to done and then continue when you have time.

probablyup:
@ryanoglesby08 actually GFM task lists will render inputs: https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments

GFM is supported in markdown-to-jsx

ryanoglesby08:
All elements within Markdown now have associated components, making them easy to override consumers. Closing this PR. Thanks for all the suggestions and help @sapegin

Issue #513, opened 2017-06-26
Document usage with styling libraries


sapegin:
I’m sue Styleguidist works with most of CSS in JS and styling libraries our of the box but set up could be not obvious. Would be nice to try all of them, find a working set up and document it.

* [x] [Styled Components](https://www.styled-components.com/) (already documented but could be improved, see also https://github.com/reactjs/react-docgen/issues/187 and https://github.com/styled-components/styled-components/issues/945)
* [ ] [react-jss](https://github.com/cssinjs/react-jss)
* [x] [Fela](https://github.com/rofrischmann/fela) (should be similar to SC, see also #478)
* [x] [react-css-modules](https://github.com/gajus/react-css-modules) (named export of the base component works, see #501)
* [ ] [Glamorous](https://github.com/paypal/glamorous)
* [ ] [Aphrodite](https://github.com/Khan/aphrodite)
* [ ] [Radium](https://github.com/FormidableLabs/radium)
* [x] [Styletron](https://github.com/rtsao/styletron)

n1313:
I'm using CSS Modules together with [themr](https://github.com/javivelasco/react-css-themr), I'll try to find time to write a few words about my setup.

sapegin:
I’ve added a new guide to the docs to document usage with third-party libraries. Feel free to submit a pull request if you have experience with libraries not described in the guide yet.

https://react-styleguidist.js.org/docs/thirdparties.html

a-ignatov-parc:
We are using react-jss and there is no extra setup needed. At least for current release. There is refactoring in progress to fix SSR issues in JSS but I don't think that it will change something in configuration.

sapegin:
Fela has an official example now: https://react-styleguidist.js.org/docs/thirdparties.html#fela

aaronjensen:
@sapegin I don't think that the example for styled-components is actually a great way of doing it. It adds another component to the tree that does nothing but pass things on to the styled component. It also adds quite a bit of verbosity to every file.

Is there a way to convince react-docgen that it's actually a component?

sapegin:
@aaronjensen Sure, it can be improved. But that’s the best possible solution for now as much as I know.

There are couple of issues in react-docgen repo but no real work has been done in the core as I know.

Also there’s some unfinished work here: https://github.com/Jmeyering/react-docgen-annotation-resolver

daphnesmit:
[...]

PR #623, opened 2017-10-04
Docs: Add Styletron example

Zhangdroid:
#513

### DIFF
--- a/docs/Thirdparties.md
+++ b/docs/Thirdparties.md
@@ -209,7 +209,7 @@ const { StyletronProvider } = require('styletron-react');
 </StyletronProvider>
 ```
 
-To use Styletron in every component, which is the same way as [Redux](#redux), redefine the Wrapper component:
+To use Styletron in every component, which is similar to [Redux](#redux), redefine the Wrapper component:
 
 ```javascript
 // styleguide.config.js
--- a/docs/Thirdparties.md
+++ b/docs/Thirdparties.md
@@ -225,8 +225,8 @@ module.exports = {
 // lib/styleguide/Wrapper.js
 import React, { Component } from 'react';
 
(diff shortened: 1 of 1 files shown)


docs/Thirdparties.md
```diff
@@ -194,3 +195,46 @@ export function Button({ color, size, children }) { /* ... */ }
...
+```jsx
+const Styletron = require('styletron-client');
+const { StyletronProvider } = require('styletron-react');
+
+<StyletronProvider styletron={new Styletron()}>
+  <App greeting="Choose your pizza!"/>
+</StyletronProvider>
+```
+
+To use Styletron store in every component, which is the same way as [Redux](#redux), redefine the Wrapper component:
```
sapegin: “store“ → ””?
“which is the same way as [Redux](#redux)” → “similar to [Redux](#redux)”?

docs/Thirdparties.md
```diff
@@ -194,3 +195,46 @@ export function Button({ color, size, children }) { /* ... */ }
...
+    Wrapper: path.join(__dirname, 'lib/styleguide/Wrapper')
+  }
+};
+```
+
+```jsx
+// lib/styleguide/Wrapper.js
+import React, { Component } from 'react';
+
+const Styletron = require('styletron-client');
```
sapegin: I’d use imports for consistency here.

docs/Thirdparties.md
Zhangdroid: Sorry, it's copy from redux part and I didn't look carefully, fixed :)

docs/Thirdparties.md
Zhangdroid: fixed

sapegin:
Thanks!

Issue #614, opened 2017-09-26
React 16


sapegin:
This issue will track React 16 compatibility. It seems to work fine but there are several peer dependency warnings.

Dependencies compatibility:

- [x] [React 16 final release](https://code.facebook.com/posts/1716776591680069/react-16-a-look-inside-an-api-compatible-rewrite-of-our-frontend-ui-library/)
- [x] [Enzyme 3.0](https://github.com/airbnb/enzyme/blob/master/CHANGELOG.md#300)
- [x] [markdown-to-jsx](https://github.com/probablyup/markdown-to-jsx) — N/A (seems to work)
- [x] [react-codemirror2](https://github.com/scniro/react-codemirror2) — ~~no support~~
- [x] [react-group](https://github.com/sapegin/react-group) — ~~N/A~~
- [x] [react-icons](https://github.com/gorangajic/react-icons) — ~~[no support]~~(https://github.com/gorangajic/react-icons/issues/103)
- [x] ~~[react-modal](https://github.com/reactjs/react-modal) — [some issues](https://github.com/reactjs/react-modal/issues/498) (we use it only in examples)~~

* **N/A** means no reported issues.
* **No support** means `peerDependencies` don’t allow React 16, may work otherwise.

Styleguidist upgrade status:

- [ ] Upgrade dev dependencies to React 16 — #631
- [x] Migrate to Enzyme 3.0 — #618
- [x] Remove react-addons-test-utils (not used anymore)
- [x] Test or update markdown-to-jsx
- [x] Update react-codemirror2
- [x] Test or update react-group — tested, works fine
- [x] Update react-icons
- [x] Update or remove react-modal
- [x] Add global [componentDidCatch handler](https://blog.bugsnag.com/react-16-error-handling/) — #627
- [x] Test Styleguidist with React 16: there should be no warning in browser console or npm warning on `npm install`

vlad-zhukov:
Just wanted to note that this issue is about using React 16 internally. Examples (can?) use the version of React you (as a developer) provide. What it means is you can use React 16 in examples right now, or use an older React version when the `styleguidist` will use v16 internally.

sapegin:
Yes, it’s about making sure that Styleguidist itself will work with React 16 and will not print peer dependencies warning.

smashercosmo:
react-codemirror2 has got support for react 16

nwwells:
[...]

PR #627, opened 2017-10-06
(feat): Adds componentDidCatch() to top-level component

tizmagik:
Adds `componentDidCatch()` to top-level `<StyleGuide />` component. We could arguably have added an additional `<ErrorBoundary />` layer in case there's an error in the `<StyleGuide />` component itself, but that seemed unlikely.

To test this, you would need to update to React 16 locally (I left that out of this PR because there's still some other pending tasks in #614 ) and then access a property on an undefined object or something in the render method of a component deep in the tree. Here's a screenshot of what it looks like:

image

Note that the console also logs the React-generated error, which is nice.

### DIFF
--- a/src/rsg-components/PlaygroundError/PlaygroundErrorRenderer.js
+++ b/src/rsg-components/PlaygroundError/PlaygroundErrorRenderer.js
@@ -21,7 +21,7 @@ export function PlaygroundErrorRenderer({ classes, message }) {
 
 PlaygroundErrorRenderer.propTypes = {
 	classes: PropTypes.object.isRequired,
-	message: PropTypes.string.isRequired,
+	message: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
 };
 
 export default Styled(styles)(PlaygroundErrorRenderer);
--- a/src/rsg-components/PlaygroundError/PlaygroundErrorRenderer.js
+++ b/src/rsg-components/PlaygroundError/PlaygroundErrorRenderer.js
@@ -21,7 +21,7 @@ export function PlaygroundErrorRenderer({ classes, message }) {
 
 PlaygroundErrorRenderer.propTypes = {
 	classes: PropTypes.object.isRequired,
-	message: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+	message: PropTypes.string.isRequired,
 };
 
 export default Styled(styles)(PlaygroundErrorRenderer);
--- a/src/rsg-components/StyleGuide/StyleGuide.js
+++ b/src/rsg-components/StyleGuide/StyleGuide.js
@@ -70,7 +70,7 @@ export default class StyleGuide extends Component {
 								This may be due to an error in a component you are overriding, or a bug in
 								react-styleguidist.<br />If you believe this is a bug, please submit an issue:&nbsp;
 								<a
-									style={{ color: 'white' }}
+									style={{ color: 'inherit' }}
 									href="https://github.com/styleguidist/react-styleguidist/issues"
 								>
 									https://github.com/styleguidist/react-styleguidist/issues
--- a/src/rsg-components/StyleGuide/StyleGuide.js
+++ b/src/rsg-components/StyleGuide/StyleGuide.js
@@ -4,6 +4,7 @@ import TableOfContents from 'rsg-components/TableOfContents';
 import StyleGuideRenderer from 'rsg-components/StyleGuide/StyleGuideRenderer';
 import Sections from 'rsg-components/Sections';
 import Welcome from 'rsg-components/Welcome';
+import PlaygroundError from 'rsg-components/PlaygroundError';
 import { HOMEPAGE } from '../../../scripts/consts';
 
 export default class StyleGuide extends Component {
@@ -32,6 +33,11 @@ export default class StyleGuide extends Component {
 		isolatedComponent: false,
 	};
 
+	state = {
+		error: false,
+		info: null,
+	};
+
 	getChildContext() {
 		return {
 			codeRevision: this.props.codeRevision,
@@ -43,8 +49,38 @@ export default class StyleGuide extends Component {
 		};
 	}
 
+	componentDidCatch(error, info) {
+		this.setState({
+			error,
+			info,
+		});
+	}
+
 	render() {
 		const { config, sections, welcomeScreen, patterns, isolatedComponent } = this.props;
+
+		if (this.state.error) {
+			return (
+				<PlaygroundError
+					classes={{}}
+					message={
+						<div style={{ margin: 10 }}>
+							{`${this.state.error} ${this.state.info.componentStack}`}
+							<p>
+								This may be due to an error in a component you are overriding, or a bug in
+								react-styleguidist.<br />If you believe this is a bug, please submit an issue:&nbsp;
+								<a
+									style={{ color: 'white' }}
+									href="https://github.com/styleguidist/react-styleguidist/issues"
+								>
+									https://github.com/styleguidist/react-styleguidist/issues
+								</a>
+							</p>
+						</div>
+					}
+				/>
+			);
+		}
 
 		if (welcomeScreen) {
 			return <Welcome patterns={patterns} />;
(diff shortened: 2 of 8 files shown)


src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -43,9 +49,39 @@ export default class StyleGuide extends Component {
...
+
 	render() {
 		const { config, sections, welcomeScreen, patterns, isolatedComponent } = this.props;
 
+		if (this.state.error) {
+			return (
+				<PlaygroundError
+					classes={{}}
+					message={
+						<div style={{ margin: 10 }}>
```
tizmagik: Wasn't sure how best to style, happy to change this, but it seemed fairly innocuous enough here 😄

src/rsg-components/PlaygroundError/PlaygroundErrorRenderer.js
```diff
@@ -21,7 +21,7 @@ export function PlaygroundErrorRenderer({ classes, message }) {
 
 PlaygroundErrorRenderer.propTypes = {
 	classes: PropTypes.object.isRequired,
-	message: PropTypes.string.isRequired,
+	message: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
```
tizmagik: I changed the PropType here to also allow a Component to be passed in for better control over spacing

src/rsg-components/StyleGuide/StyleGuide.js
sapegin: I think we shouldn’t use PlaygroundError component here — full screen of ~~blood~~ would be an overkill, so just red text on white background would be enough. I think we can extract the layout from Welcome component to a Page (don’t know a better name for that) component and use it to show an error. Also please take a look at our developer guide: https://react-styleguidist.js.org/docs/development.html

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -43,9 +49,39 @@ export default class StyleGuide extends Component {
...
+							{`${this.state.error} ${this.state.info.componentStack}`}
+							<p>
+								This may be due to an error in a component you are overriding, or a bug in
+								react-styleguidist.<br />If you believe this is a bug, please submit an issue:&nbsp;
```
sapegin: It is React Styleguidist ;-)

I’d make [submit an issue] a link, no need to show the URL in the UI.

And you can use Markdown component to render text, see WelcomeRenderer component.

sapegin:
And we’ll need some new tests ;-)

src/rsg-components/StyleGuide/StyleGuide.js
tizmagik: Yea I wanted to minimize how much code is used when rendering an error; I even hesitated to use PlaygroundError since it could also contain errors itself. Do you think just straight text would work? I'd rather not introduce another component and avoid another potential failure point by doing sol

tizmagik: Let me push that up and we can talk through it more

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -61,24 +60,20 @@ export default class StyleGuide extends Component {
...
-									style={{ color: 'inherit' }}
-									href="https://github.com/styleguidist/react-styleguidist/issues"
-								>
-									https://github.com/styleguidist/react-styleguidist/issues
-								</a>
-							</p>
-						</div>
-					}
-				/>
+				<div style={{ margin: 10, color: 'crimson' }}>
```
sapegin: * We still need to extract this — non *Render components shouldn’t print any HTML ;-)
* Margin should be 8 or better 16 (we use 8px grid by default).

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -61,24 +60,20 @@ export default class StyleGuide extends Component {
...
+					<code style={{ whiteSpace: 'pre' }}>{`${this.state.error} ${this.state.info
```
sapegin: * Any reason not to user `<pre>` tag? ;-)
* I don’t think you need to wrap error message into another string.

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -61,24 +60,20 @@ export default class StyleGuide extends Component {
...
+						.componentStack}`}</code>
+					<p>
+						This may be due to an error in a component you are overriding, or a bug in React
+						Styleguidist.<br />If you believe this is a bug,&nbsp;
```
sapegin: Probably would be better to print this in another paragraph.

src/rsg-components/StyleGuide/StyleGuide.js
tizmagik: Heh okay if you insist. I can create a new overridable `Error` component to render the error message, but IMO that introduces an increased chance of situations where it's possible to hit an error condition while trying to render the error message itself. I guess we always have the `console.error` to fallback to? 😁

sapegin: I’m not insisting on using JSS, etc. but I’m trying to explain why it’s not that easy as it look ;-)

tizmagik: So is it a matter of also setting a background color here? I'm not sure what approach you'd like to take here...

sapegin: Yeah, I think that would be enough to ensure readability. I still hope it will be a rare case to see this thing in real life ;-)

tizmagik: Okay cool, I could do that 👍 

Yea, hopefully this never even hits, but you never know 🤷‍♂️

sapegin: I also think it’s safe to import styles from https://github.com/styleguidist/react-styleguidist/blob/master/src/styles/theme.js because its just const values. It won’t be overridable by the use but we’ll have less hardcoded styles.

tizmagik: Yes, agreed, I actually started doing that, will push that up shortly

src/rsg-components/StyleGuide/StyleGuide.js
tizmagik: - `<pre>` makes sense 😁 
- Error needs to be wrapped because it's actually an object (stack trace), so this coerces it to a string.

sapegin: Then I’d to `.toString()` to clearly show the intent.

tizmagik:
Okay pushed up the changes we talked about. I'll add tests once we agree on the approach 👍

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -43,9 +49,48 @@ export default class StyleGuide extends Component {
...
+							If you believe this is a bug,&nbsp;
+							<a
+								style={{ color: 'inherit' }}
+								href="https://github.com/styleguidist/react-styleguidist/issues"
+							>
+								please submit an issue
+							</a>.
+						</p>
+					</div>
+				</div>
```
tizmagik: We can move this "component" to a separate file, but I don't think it should be overridable by the user.

sapegin: We can’t enforce that but it’s not a reason to move it ;-) Consistency is one reason and it’s just a bit chunk of HTML that I’d like to hide from this component.

sapegin: Now I like it! 🍕

src/rsg-components/StyleGuide/StyleGuide.spec.js
```diff
@@ -46,6 +46,15 @@ it('should render welcome screen', () => {
 	expect(actual).toMatchSnapshot();
 });
 
+it('should render an error when componentDidCatch() is triggered', () => {
+	const wrapper = shallow(<StyleGuide codeRevision={1} config={config} sections={[]} slots={{}} />);
+	wrapper
+		.instance()
+		.componentDidCatch({ toString: () => 'error' }, { componentStack: { toString: () => 'info' } });
```
tizmagik: Note the bit of complexity introduced here by explicitly using `.toString()` in the component 😞

sapegin: Looks good to me ;-)

tizmagik:
Good to merge?

sapegin:
As soon as we extract the HTML to a separate component ;-)

tizmagik:
Oh, not sure what you mean. I suggested a separate file but you said you liked it as it was? I even got a 🍕 emoji! Haha

Could you clarify, or better yet, commit to the branch 😁

sapegin:
I like the code but I still want it in a separate file ;-)

sapegin:
Merged, thanks!

Issue #424, opened 2017-05-02
Improve Responsive navigation 


SaraVieira:
When on mobile navigation should be a pop up menu on the side but in this case it just gets pushed to the bottom in a somewhat weird way. 

![screenshot_20170502-110514](https://cloud.githubusercontent.com/assets/1051509/25613856/abb5d354-2f27-11e7-98fb-ca8c4394fadd.png)

Open to suggestions here but I think having slide out menu when there is no more room would be the way to go, thoughts? 
.I will work on this if it's not a problem

sapegin:
There are two things:

1. We’ve never spend much time on mobile experience so it’s very simplistic now.

2. There’s more generic UI issue: we’re keep adding things to the UI and right now it’s a bit cluttered and inconsistent.

So any improvements in both are very appreciated but try not to fix just one issue but think a bit broader ;-)

SaraVieira:
I'll try add a pixel grid to this to make the elements more evenly spaced. 
Also put the logo in the header would be awesome :) 

Should I just work with freedom and put the screenshots here?

sapegin:
I was thinking about adding something [like this](https://medium.freecodecamp.com/css-in-javascript-the-future-of-component-based-styling-70b161a79a32) (see *Define spacing constants*).

> Should I just work with freedom and put the screenshots here?

Yeah! 🍕

SaraVieira:
That is a great idea , that's exactly what I was thinking about, using a 8px grid and everything 😄 

Thank you for the push acess 🎉

SaraVieira:
@sapegin The designer in my team sent me this and I think this would be awesome:

![kapture_2017-05-02_at_18 24 48](https://cloud.githubusercontent.com/assets/1051509/25657229/bfb5dcdc-2ff4-11e7-9108-c50eecc0ab13.gif)

sapegin:
But we already have almost the same, but without animation ;-)

I have a proposal for UI improvements but I need some time to write it, will try to open an issue today.

SaraVieira:
@sapegin If you create it today I can just close this ticket and just work in the other one 👍

SaraVieira:
@sapegin It's the theme that makes it look awesome

sapegin:
If you mean dark editor theme, you can change theme for your style guide ;-)

SaraVieira:
I know you can change it but I think it would be a better default

sapegin:
[...]

PR #629, opened 2017-10-06
WIP: Small UI Improvements Phase 2

### LINKED ISSUE #426: UI improvements proposal (UI 2.0)
We keep adding new function and new “buttons” to the UI and right now it’s inconsistent and far from perfect. Adding more features (and there are plans and wishes to do that) will most likely destroy user experience completely. I was thinking about that for a long time and here are my thoughts.

Goals I want to solve in this issue:

* Improve developer experience (component development process);
* Simplify and streamline the UI;
* Add UI extension points for plugins.

Styleguidist UI consists of these three “levels” (and actions you can do on each level, with some future plans):

* Global (toggle sidebar, toggle props/description, switch background, close isolation = go home);
* Component / section (get anchor link, open in isolation);
* Example (open in isolation, toggle code visibility, show snapshots).

As a possible solution I see something like this:

ui2

Sorry for my drawing skills but basically we’ll have toolbars for each “level” that could be possible to extend from plugins plus (also extendable) tab bar for each example (this was already discussed in a relation to the plugin API).

Related issues: #354, #333, #327, #424, #421, #235.

Any feedback? Some design help would be very useful here ;-)

SaraVieira:
**The latest demo: https://styleguide-vwetycents.now.sh/**

What is done:

* Carret added on Props to help people understand that it's a clickable drop-down
* Sidebar now closes and opens with the click of a button 🎉
* Toggle props button is in place
* Responsive Navigation, when the user is on a phone the navigation will hide and in his place the user will only see the show sidebar button to toggle it.

What I need help:

* Understand the best way to make toggle props master button work
* Fix commented test
* Feedback 

Closes #426 & #424


SaraVieira:
I know the build is failing , any idea what values used to be ?

sapegin:
> I know the build is failing

It says the build is now two times bigger than it used to be.

SaraVieira:
- [ ] - Code tab should look the same way and the arrow icons should be inside the tab button component. 
- [ ] - remove yarn lock file.
- [ ] - Toggle props is hidden in isolated mode

On it !

ankri:
I really like the new looks! Could you please also add a `title` attribute to all the toggles and buttons? I don't think there is a need for a separate tooltip dependency but `title`s for example would have helped me to understand what this button is for:

image

sapegin:
Yeah, we just use `title` for all other toolbar buttons.

sapegin:
I’m not giving much feedback about the code because we’re not yet sure how everything should look. But I’ll read it carefully later ;-)

> I know but why ? I didn't even install anything

That’s something weird. I only see two new icons, something must be wrong here. I’ll take a look later.

> You think the click should make changes to the theme ?

I don’t think so. Maybe we should hide the header and show component name somewhere else, like on a fixed toolbar near buttons to toggle sidebar and toggle uberisolated mode. I’m open for ideas ;-)

/cc @okonet @kof @tizmagik @n1313 @stepancar

SaraVieira:
@sapegin I could do a tooltip in just CSS and the content in the before, what you think?

SaraVieira:
I'll do some iterations of that and send some screenshots :)

sapegin:
Maybe a good idea but as a separate PR ;-)

sapegin:
Also use `ToolbarButton` bur buttons — right now it’s just an image.

SaraVieira:
What do you mean by the last comment?

sapegin:
You’re attaching an `onClick` handler to an image which make them inaccessible, inconsistent with other toolbar buttons, you have to write a lot of extra styles, and we already have a component for toolbar buttons.

SaraVieira:
Oh! Okay, cool 😃

SaraVieira:
@ankri @sapegin Added the toolbarButton and it looks waaaay better actually :p 

thank you for reminding me 👍 

https://styleguide-vwetycents.now.sh/

Will take a lot at some possible mods tomorrow for focus mode :)

okonet:
@SaraVieira here is my thoughts on the UI:

1. I think the location of the buttons in the sidebar is odd. It was merely an accident I noticed them.
1. I'm wondering why do we even need the hide sidebar button?
1. The second button doesn't do anything for me...

SaraVieira:
@okonet Thanks for the feedback! 🙌

Regarding the second button it's true it doesn't do anything yet :(

Where do you think it would be a better position for them?

About the hiding the sidebar it has come in handy for me a couple of times but yes it's a niche problem

sapegin:
@SaraVieira Any particular use cases for manually hiding the sidebar? I really want the uberisolated mode ;-)

SaraVieira:
The main use case is mobile like it does right now, besides that it could be useful if you want to work in focus mode.

Now that I think about its more of a pretty thing then an actually necessity on desktop 🙁

sapegin:
But do you really need to see all components in that mode?

SaraVieira:
Mostly not 🙁

Maybe the sidebar removal in desktop should enable focus mode instead of having two buttons

SaraVieira:
Btw, @sapegin I pinged you on slack

sapegin:
I think the main question here is what would be the best experience for developing components?

For me it would be uberisolation mode: 

* one component
* all examples visible at the same time
* No distractions like a huge heading, pathline, etc.
* Togglable side paddings in example playgrounds (for responsive testing)

okonet:
Yes, let's approach the problem with real use cases and not with the UI first. I'll write down my thoughts on the DX as soon as I get some more time. Sorry for delays :-/

tizmagik:
Yes, for me the most important DX consideration is lack of chrome (padding, etc) for responsive testing.

SaraVieira:
@tizmagik I can tackle that and just get a better mobile experience for testing

@okonet Not a problem man! Get back to this when you have time :)

sapegin:
I’d suggest to split this PR otherwise we’ll never finish it because there are so many things that needs further discussion.

> Carret added on Props to help people understand that it's a clickable drop-down

+1 from me.

> Sidebar now closes and opens with the click of a button 🎉

Not sure we really need that.

> Toggle props button is in place

That’s the whole discussion of the uberisolated mode.

> Responsive Navigation, when the user is on a phone the navigation will hide and in his place the user will only see the show sidebar button to toggle it.

This is useful I think but since we’re not sure about the toggle sidebar button I’m not sure how it should work.

What do y’ll think?

My thought and wishes about uberisolation:

image 2017-11-17 at 1 14 49 pm

Any feedback?

Issue #630, opened 2017-10-07
Errors running CRA Example


tizmagik:
Stumbled on this when working on React 16 upgrade

- fresh checkout of `react-styleguidist`
- `npm install`
- `npm run start:cra`

```
➜  crars git:(master) npm run start:cra

> react-styleguidist@0.0.0-development start:cra /Users/206567/Documents/workspace/scratch/crars
> cd examples/cra && ../../bin/styleguidist.js server

You can now view style guide in the browser.

Local:            http://localhost:6060/

On your network:  http://192.168.1.12:6060/

./src/components/Button.js
Module parse failed: /Users/206567/Documents/workspace/scratch/crars/examples/cra/src/components/Button.js Unexpected token (16:2)
You may need an appropriate loader to handle this file type.
| 
| 	return (
| 		<button className="button" style={styles}>
| 			{children}
| 		</button>

```

sapegin:
Does it work when you cd to that folder and run it from there?

tizmagik:
Sort of, I get these:

![image](https://user-images.githubusercontent.com/244704/31310948-f6694f6a-ab6f-11e7-9632-048a78c6b48a.png)

Is it supposed to work from running `npm run start:cra` from top-level?

sapegin:
Yes, it is. I’m not sure why it’s not working now.

PR #631, opened 2017-10-07
React 16 Upgrade

### LINKED ISSUE #614: React 16

tizmagik:
Everything looks like it's working with no warnings or errors that I could see, except:

 - CRA example not working (also an issue on `master` I think?), filed a separate issue here: #630 
 - Test failures with Preview error handling test -- I would like to discuss this further...

For the Preview component, I think we might need to change up how we do error handling in React 16? I'm wondering if instead of a try/catch if we need to add a `componentDidCatch()` to the preview renderer and catch/report errors there. Since React is a peerDep, this would have the nice side affect of not triggering when users are running with React 15 and the old try/catch behavior would still work, I think.

Please have a look and let me know what you think we should do...

Closes #614


package.json
```diff
@@ -1,213 +1,215 @@
 {
```
sapegin: Indentation was changed for some reason. There should be two spaces.

sapegin:
Yeah, sounds good, if it will work in React 15 and 16.

test/raf-polyfill.js
```diff
@@ -0,0 +1,3 @@
+// requestAnimationFrame “polyfill”
+window.requestAnimationFrame = a => a();
+global.requestAnimationFrame = window.requestAnimationFrame;
```
tizmagik: Note that this is technically an incorrect shim, it should be a `setTimeout()` or something to properly simulate async nature of rAF.

sapegin: I don’t care about async nature ;-)

tizmagik: Hm yea, that might be a slippery slope though, especially when async rendering is enabled by default in the future.

tizmagik:
Does VSCode not respect editorconfig? Also not sure why it only does it for that package.json

tizmagik:
Editorconfig was on. I don't know, I give up, I ran prettier manually via CLI and won't touch that file anymore 🤞

sapegin:
Please don’t run Prettier on package.json — it’s all messed up now ;-|

sapegin:
> Does VSCode not respect editorconfig? 

It has a plugin for that. But I have a feeling that it doesn’t always work.

tizmagik:
Yea I had the plugin and it was enabled, not sure why it messed up the spacing... ah well that's not the important diff for right now, need to get the tests working

tizmagik:
This upgrade might not be as straight forward actually, need to give how we're error handling some more thought...

image

Full message

```
Error: A cross-origin error was thrown. React doesn't have access to the actual error object in development. See https://fb.me/react-crossorigin-error for more information.
```
More info, https://fb.me/react-crossorigin-error

---

If anyone has any ideas, feel free to commit to this branch or branch off of it and run with it 👍

sapegin:
As I understand we’re trying to print an error from another React root:

https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Preview/Preview.js#L99-L103

Probably we should use componentDidCatch and rethrow it or something like that.

sapegin:
@tizmagik Did you have time to investigate the issue? Do you need any help with this?

tizmagik:
Haven’t had time to look at this since my last commit but hope to pick this up again this weekend. But, as always, feel free to jump in if you have any ideas!

tizmagik:
Cool I think we're good now @sapegin , all synced up with master too

examples/basic/package.json
```diff
@@ -25,9 +25,10 @@
     "css-loader": "^0.28.4",
     "dog-names": "^1.0.2",
     "lodash": "^4.17.4",
-    "prop-types": "^15.5.10",
-    "react": "^15.4.2",
-    "react-dom": "^15.4.2",
+    "prop-types": "^15.6.0",
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0",
+    "react-modal": "^3.0.0",
```
sapegin: We’re not using react-modal anymore.

src/rsg-components/Preview/Preview.js
```diff
@@ -90,7 +90,7 @@ export default class Preview extends Component {
 
 		const exampleComponent = this.evalInContext(compiledCode);
 		const wrappedComponent = (
-			<Wrapper>
+			<Wrapper onError={err => this.handleError(err)}>
```
sapegin: We’re using `bind` in constructor in other components.

tizmagik: Hm, I dunno why but the tests failed with the bind. Thoughts?

sapegin: You were adding it to another component — I’ve fixed it and it seems to work fine now ;-)

sapegin:
Thanks, merged! 🍕

src/rsg-components/Preview/Preview.js
tizmagik: Oh haha thanks, was context switching too much I guess 😁

PR #632, opened 2017-10-09
Update react-styleguidist in CRA example

adityavohra7:
Related to #630. I updated `react-styleguidist`, and regenerated `package-lock.json`. `cd`ing into the example's directory and running `npm run styleguide` now works as expected. I can look into why `npm run start:cra` doesn't work from the root dir next!

screen shot 2017-10-08 at 8 24 11 pm

### DIFF
--- a/examples/cra/package-lock.json
+++ b/examples/cra/package-lock.json
@@ -4,15 +4,3185 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "abab": {
+      "version": "1.0.4",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.4.tgz",
+      "integrity": "sha1-X6rZwsB/YN12dw9xzwJbYqY8/U4=",
+      "dev": true
+    },
+    "accepts": {
+      "version": "1.3.4",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.4.tgz",
+      "integrity": "sha1-hiRnWMfdbSGmR0/whKR0DsBesh8=",
+      "dev": true,
+      "requires": {
+        "mime-types": "2.1.17",
+        "negotiator": "0.6.1"
+      }
+    },
+    "acorn": {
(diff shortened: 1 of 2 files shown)


adityavohra7:
Update: `npm run start:cra -- --verbose` yields [this](https://gist.github.com/adityavohra7/5a3de829ed07a389c7d3409592b1ddbe#file-gistfile1-txt-L59). I can't seem to find `babel-loader` anywhere in there. `babel-loader` _does_ show up for the basic example. Is the CRA example missing a `styleguide.config.js` that specifies `babel-loader`, or am I missing some magic 😅

sapegin:
Looks like this line doesn’t work:

https://github.com/styleguidist/react-styleguidist/blob/da644226cd990c0cb92e320e4a420dadcd5bd535/scripts/utils/findUserWebpackConfig.js#L22-L23

sapegin:
Thanks, merged!

adityavohra7:
@sapegin yeah I think that makes sense, `resolve(CREATE_REACT_APP_WEBPACK_CONFIG);` will try to find `react-scripts` in the root level `node_modules`, where it isn't installed.

sapegin:
Yeah, probably we should just remove it from the main package.json and run from it’s own folder.

adityavohra7:
Either that or change the `start:cra` run script to something like `cd examples/cra && npm install && npm run styleguide`. Though I personally wouldn't mind having to manually `cd` into the examples dir to start  styleguide 😅  It also makes the root `package.json` a little simpler.

Though if we remove the `start:cra` run script from the root `package.json`, should we do the same for the other examples as well? So running all examples is consistent? Also, are `preact` related things in the root `package.json` specifically for the `preact` example?

sapegin:
I’d leave these three because I use them very often to test new features and we run them on CI:

https://github.com/styleguidist/react-styleguidist/blob/583bd3eb1882e73c7a7ddc5336d8a84a40f5d671/package.json#L142-L144

PR #635, opened 2017-10-11
Fix: Update dependency react-docgen

odinuge:
The new verison of `react-docgen@next` adds support for parsing flow
function parameters without names. Eg. `value: string => void`.

Previously react-docgen was unable to parse files containing flow
annotations like these, and therefore resulting in styleguidist being
unable to render them.


sapegin:
Thanks, merged!

PR #638, opened 2017-10-13
clear console on reload

### LINKED ISSUE #519: Run console.clear() when components are reloaded

xjlim:
fix #519

### DIFF
--- a/src/rsg-components/Preview/Preview.js
+++ b/src/rsg-components/Preview/Preview.js
@@ -49,6 +49,7 @@ export default class Preview extends Component {
 	};
 
 	componentDidMount() {
+		console.clear(); // eslint-disable-line no-console
 		this.executeCode();
 	}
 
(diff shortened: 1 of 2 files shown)


sapegin:
Thanks, merged!

BruceHubbard:
What was the logic behind clearing the console?  I recently updated my styleguidist version and I've noticed all of my React warnings are now hidden because the console gets cleared here.

sapegin:
@BruceHubbard I guess it shouldn’t be cleared on hard page reload, only on HMR updates. If it’s not the case, then it’s a bug.

BruceHubbard:
@sapegin I'll open a new issue and see if others are having the same problem.  On a hard page reload in Chrome I end up with zero React warnings when I know I should have some.

PR #640, opened 2017-10-14
Run console.clear when Preview component is mounted. 

soul-wish:
Hi. Here is a fix for #519. Also, I created a test to check if `console.clear` was called (thanks to @tizmagik for his comments in other PR). Let me know if I've missed something 😉


soul-wish:
Hm.. It seems like I'm a bit late with these changes :) (wrote a code yesterday but finished with a test case only today). It seems like we should reject this PR, right @sapegin ?

sapegin:
Yeah, thanks anyway ;-)

Issue #641, opened 2017-10-16
Styleguidist tries to render code blocks without language tag


wkillerud:
> Any code block with a language tag of `js`, `jsx` or `javascript` will be rendered as a React component with an interactive playground

From the [documentation](https://github.com/styleguidist/react-styleguidist/blob/master/docs/Documenting.md#usage-examples-and-readme-files) it looks like only code blocks with the language tab `js`, `jsx`, or `javascript` should render React components. However React Styleguidist even tries to render code blocks without a language tag.

A workaround is to lie and mark the block as a valid but non-JavaScript/JSX language, for instance Java.

A repo which demonstrates the issue and workaround with a component called `Dots`: https://github.com/wkillerud/example

I would expect code blocks with no language set to be rendered as a code block and not an interactive version.

sapegin:
This is expected behavior for backwards compatibility, use`static` modifier.

wkillerud:
Allright. Could the documentation perhaps be updated to reflect this then? :) I can update the quoted line and add an entry to the cookbook if you'd like.

sapegin:
We don’t recommend it but I agree it would be nice to mention that it still works ;-) 

I think this is the best place for it: https://github.com/styleguidist/react-styleguidist/blob/master/docs/Documenting.md (and it should be mentioned that this is backwards compatible mode, not recommended one).

PR #642, opened 2017-10-16
Explain behavior for code blocks w/o language tag

wkillerud:
In issue #641 I was confused that code blocks without language tags were
being rendered as a React component, and not a code documentation block.
It turned out this was for backward compatibility reasons. This commit
updates the documentation to hopefully avoid such confusion in the
future. The docs now also suggest a workaround.

### DIFF
--- a/docs/Documenting.md
+++ b/docs/Documenting.md
@@ -58,7 +58,7 @@ export default class Button extends React.Component {
 
 ## Usage examples and Readme files
 
-Styleguidist will look for any `Readme.md` or `ComponentName.md` files in the component’s folder and display them. Any code block with a language tag of `js`, `jsx` or `javascript` will be rendered as a React component with an interactive playground.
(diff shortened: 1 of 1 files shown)


sapegin:
Thanks!

wkillerud:
So sorry, I should have tested this in the repo I created before I made a PR. Creating blocks using "static" instead of "javascript" just gives an error:

```
```static
```

> Unknown language: "static"

Is there a different way this modifier should be set?

I tried digging down in the dependencies to find other documentation but came up short when I reached [Github Flavored Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) and [Linguist](https://github.com/github/linguist) without finding any references to "static".

sapegin:
It should be `js static`, for example, see examples here: https://react-styleguidist.js.org/examples/basic/

sapegin:
Yeah, I’ve missed this too:

> `static` in place of a language tag

sholud be

> `static` modifier with a language tag

wkillerud:
Oh, OK. I need to read the examples more thoroughly it seems. Sorry for the confusion!

sapegin:
No worries, and if you send another PR, it would half of a Hacktoberfest t-shirt which is cool ;-)

PR #647, opened 2017-10-19
Feat: Forkme ribbon option

wkwiatek:
Adds 'Fork me' ribbon from #195.

### DIFF
--- a/src/rsg-components/Ribbon/Ribbon.js
+++ b/src/rsg-components/Ribbon/Ribbon.js
@@ -1,0 +1,11 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import RibbonRenderer from 'rsg-components/Ribbon/RibbonRenderer';
+
+export default function Ribbon({}, { config }) {
+	return <RibbonRenderer {...config.ribbon} />;
+}
+
+Ribbon.contextTypes = {
+	config: PropTypes.object,
+};
--- a/src/rsg-components/StyleGuide/StyleGuide.js
+++ b/src/rsg-components/StyleGuide/StyleGuide.js
@@ -73,6 +73,7 @@ export default class StyleGuide extends Component {
 				homepageUrl={HOMEPAGE}
 				toc={<TableOfContents sections={sections} />}
 				hasSidebar={config.showSidebar && !isolatedComponent}
+				ribbon={config.ribbon}
 			>
 				<Sections sections={sections} depth={1} />
 			</StyleGuideRenderer>
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -24,6 +24,7 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 - [`propsParser`](#propsparser)
 - [`require`](#require)
 - [`resolver`](#resolver)
+- [`ribbon`](#ribbon)
 - [`sections`](#sections)
 - [`serverHost`](#serverhost)
 - [`serverPort`](#serverport)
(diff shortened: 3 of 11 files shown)


src/rsg-components/Ribbon/Ribbon.js
```diff
@@ -0,0 +1,50 @@
...
+		fontFamily: fontFamily.base,
+		position: 'relative',
+		right: -37,
+		top: -22,
+		display: 'block',
+		width: 190,
+		padding: '5px 15px',
+		textAlign: 'center',
+		fontSize: 14,
+		background: '#f9970d',
```
sapegin: It would be better to define default colors and text in config schema. Probably you’ll need to use `process` function for that.

src/rsg-components/Ribbon/Ribbon.js
```diff
@@ -0,0 +1,50 @@
...
+export const styles = ({ fontFamily }) => ({
+	forkme: {
+		position: 'fixed',
+		top: 0,
+		right: 0,
+		width: '149px',
+		height: '149px',
+		zIndex: 999,
+	},
+	forkmeIcon: {
```
sapegin: It’s not really an icon, I’d call it `link`.

src/rsg-components/Ribbon/Ribbon.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Styled from 'rsg-components/Styled';
+
+export const styles = ({ fontFamily }) => ({
+	forkme: {
```
sapegin: `root` or `ribbon` would be more in line with other components.

src/rsg-components/Ribbon/Ribbon.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
```
sapegin: Please take a look at https://react-styleguidist.js.org/docs/development.html This component should be split into two: Ribbon that will read config for a React context and RibbonRenderer — this component that prints actual HTML.

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -73,6 +73,7 @@ export default class StyleGuide extends Component {
 				homepageUrl={HOMEPAGE}
 				toc={<TableOfContents sections={sections} />}
 				hasSidebar={config.showSidebar && !isolatedComponent}
+				ribbon={config.ribbon}
```
sapegin: You don’t need this because all config is accessible in the context.

sapegin:
And please revert lock file changes, you don’t have any changes in `package.json`.

src/rsg-components/Ribbon/Ribbon.js
wkwiatek: Ok, please take a look now and see if it's correct now.

src/rsg-components/StyleGuide/StyleGuide.js
wkwiatek: I think we still need this here as we're passing it down to the component to check if we need to render a ribbon (just like `hasSidebar`).

sapegin:
It’s closer now but a few things are still missed:

> Ribbon [component] that will read config for a React context

You’re still passing it down via props.

> It would be better to define default colors and text in config schema. Probably you’ll need to use process function for that.

You’re using a theme now. It may be a, actually, good idea but in that case you don’t need colors inside the `ribbon` option.

src/rsg-components/StyleGuide/StyleGuide.js
sapegin: No if you’ll read config from the context as I’m suggesting ;-)

PR #649, opened 2017-10-20
Passing codemirrorOptions as a configuration

4mgad:


scripts/schemas/config.js
```diff
@@ -39,6 +39,18 @@ module.exports = {
 			map: 'lodash/map',
 		},
 	},
+	codemirrorOptions: {
+		type: 'object',
+		default: {
```
sapegin: This will be overwritten if `codemirrorOptions` isn’t `undefined`, we need to merge objects manually using `process` function.

loaders/styleguide-loader.js
```diff
@@ -15,7 +15,7 @@ const filterComponentsWithExample = require('./utils/filterComponentsWithExample
 // Config options that should be passed to the client
 const CLIENT_CONFIG_OPTIONS = [
 	'title',
-	'highlightTheme',
+	'codemirrorOptions',
```
sapegin: This is a breaking change, we should keep the old option until the next major release. See `updateExample` option in the config schema.

`editorConfig` would be a more generic name, and better match existing options like `compilerConfig`.

4mgad:
Hi Artem, unfortunately I can't spend more time on this issue for the time being.

sapegin:
Closing in favor of #662.

PR #657, opened 2017-10-31
feat: Add JSDoc @component annotation support

okonet:
- [x] Enables documenting styled-components without additional wrappers
- [x] Update documentation
- [x] Add tests

### DIFF
--- a/docs/Thirdparties.md
+++ b/docs/Thirdparties.md
@@ -161,7 +161,7 @@ module.exports = {
 
 ### Styled Components
 
-The [recommended way](https://github.com/styleguidist/react-styleguidist/issues/37#issuecomment-263502454) of using [Styled Components](https://www.styled-components.com/) is like this:
+The recommended way of using [Styled Components](https://www.styled-components.com/) is by using a special `@component` JSDoc annotation:
 
 ```jsx
 import React from 'react';
@@ -173,9 +173,9 @@ const SalmonButton = styled.button`
   color: snow;
 `;
 
-export default function Button({ children, ...props }) {
-  return <SalmonButton {...props}>{children}</SalmonButton>;
-}
+/** @component */
+export default SalmonButton;
+
 ```
 
 ### Fela
--- a/scripts/schemas/config.js
+++ b/scripts/schemas/config.js
@@ -10,6 +10,7 @@ const startCase = require('lodash/startCase');
 const reactDocgen = require('react-docgen');
 const createDisplayNameHandler = require('react-docgen-displayname-handler')
 	.createDisplayNameHandler;
+const annotationResolver = require('react-docgen-annotation-resolver').default;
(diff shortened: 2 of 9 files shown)


okonet:
Related to https://github.com/styled-components/styled-components/issues/945

docs/Thirdparties.md
```diff
@@ -173,9 +173,11 @@ const SalmonButton = styled.button`
   color: snow;
 `;
 
-export default function Button({ children, ...props }) {
-  return <SalmonButton {...props}>{children}</SalmonButton>;
-}
+/**
+ * @component
+ */
+export default SalmonButton;
```
sapegin: Will `/** @component */` work too?

scripts/schemas/config.js
```diff
@@ -107,7 +108,12 @@ module.exports = {
 	},
 	resolver: {
 		type: 'function',
-		default: reactDocgen.resolver.findAllExportedComponentDefinitions,
+		default: (ast, recast) => {
+			const defaultResolver = reactDocgen.resolver.findAllExportedComponentDefinitions;
```
sapegin: I’d not call it a default resolver, it’s not default for react-docgen and not default for Styleguidist anymore ;-)

sapegin:
Would be nice to have some tests too ;-)

And thanks for the pull request!

scripts/schemas/config.js
okonet: So what would you use instead?

docs/Thirdparties.md
okonet: Yes, as long it's a valid JSDoc it will work. Tests are following.

scripts/schemas/config.js
sapegin: I’d use `findAllExportedComponentDefinitions` probably, it’s clear at least ;-)

docs/Thirdparties.md
sapegin: I’d use shorter form in the docs then ;-)

MoeSattler:
@sapegin @okonet this is really important. Can I help to make this happen?

sapegin:
@MoeSattler yes you can and it will be amazing! We need some basic tests to make sure we won’t break this.

sapegin:
Probably something like this would work:

https://github.com/styleguidist/react-styleguidist/blob/12b8db1f5267fb147a9216b5e8cd8dc41a74447c/loaders/__tests__/props-loader.spec.js#L27-L44

okonet:
@sapegin added tests and rebased (and renamed) commits on top of latest `master`

sapegin:
Thanks, merged!

MoeSattler:
Thanks a lot @okonet !
@sapegin any idea already when you will release?

sapegin:
@MoeSattler Soonish, next week probably ;-)

PR #662, opened 2017-10-31
feat(editor): Passes all codeMirror options to editor

### LINKED ISSUE #648: Line numbers enable configuration needed
Is it possible provide a showLineNumbers configuration flag in styleguide.config.js to enable/disable showing line numbers in the code editor?

SaraVieira:
All you now pass inside codeMirrorOptions in your styleguide config will be passed to codeMirror.
I left highlight theme out because I didn't want to cause a breaking change

closes #648

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -14,10 +14,10 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 * [`configureServer`](#configureserver)
 * [`dangerouslyUpdateWebpackConfig`](#dangerouslyupdatewebpackconfig)
 * [`defaultExample`](#defaultexample)
+* [`editorConfig`](#editorConfig)
 * [`getComponentPathLine`](#getcomponentpathline)
 * [`getExampleFilename`](#getexamplefilename)
 * [`handlers`](#handlers)
-* [`editorConfig`](#editorConfig)
 * [`ignore`](#ignore)
 * [`logger`](#logger)
 * [`previewDelay`](#previewdelay)
@@ -168,6 +168,12 @@ module.exports = {
 }
 ```
 
+#### `editorConfig`
+
+Type: `Object`, default: [scripts/schemas/config.js](https://github.com/styleguidist/react-styleguidist/blob/87efdde2d52209e9c4072368402d11030943d69f/scripts/schemas/config.js#L101)
+
+Source code editor options, see [CodeMirror docs](https://codemirror.net/doc/manual.html#config) for all available options.
+
 #### `getExampleFilename`
 
 Type: `Function`, default: finds `Readme.md` or `ComponentName.md` in the component folder
@@ -226,23 +232,6 @@ module.exports = {
     )
 }
 ```
-
-#### `editorConfig`
-
-Type: `Object`, default:
-```
-{
-  theme: 'base16-light',
-  mode: 'jsx',
-  lineWrapping: true,
-  smartIndent: false,
-  matchBrackets: true,
-  viewportMargin: Infinity,
-  lineNumbers: false,
-}
-```
-
-This will determine how your editor looks, you can check [here](https://codemirror.net/doc/manual.html#config) all the available options
 
 #### `ignore`
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -17,7 +17,7 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 * [`getComponentPathLine`](#getcomponentpathline)
 * [`getExampleFilename`](#getexamplefilename)
 * [`handlers`](#handlers)
-* [`highlightTheme`](#highlighttheme)
+* [`editorConfig`](#editorConfig)
 * [`ignore`](#ignore)
 * [`logger`](#logger)
 * [`previewDelay`](#previewdelay)
@@ -227,11 +227,22 @@ module.exports = {
 }
 ```
 
-#### `highlightTheme`
+#### `editorConfig`
 
-Type: `String`, default: `base16-light`
+Type: `Object`, default:
+```
+{
+  theme: 'base16-light',
+  mode: 'jsx',
+  lineWrapping: true,
+  smartIndent: false,
+  matchBrackets: true,
+  viewportMargin: Infinity,
+  lineNumbers: false,
+}
+```
 
-[CodeMirror theme](http://codemirror.net/demo/theme.html) name to use for syntax highlighting in the editor.
+This will determine how your editor looks, you can check [here](https://codemirror.net/doc/manual.html#config) all the available options
 
 #### `ignore`
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -170,7 +170,7 @@ module.exports = {
 
 #### `editorConfig`
 
-Type: `Object`, default: [scripts/schemas/config.js](https://github.com/styleguidist/react-styleguidist/blob/87efdde2d52209e9c4072368402d11030943d69f/scripts/schemas/config.js#L101)
+Type: `Object`, default: [scripts/schemas/config.js](https://github.com/styleguidist/react-styleguidist/tree/master/scripts/schemas/config.js#L101)
 
 Source code editor options, see [CodeMirror docs](https://codemirror.net/doc/manual.html#config) for all available options.
 
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "html-webpack-plugin": "^2.30.1",
     "is-directory": "^0.3.1",
     "javascript-stringify": "^1.6.0",
+    "jest-cli": "^22.0.0",
     "jss": "^9.3.3",
     "jss-camel-case": "^6.0.0",
     "jss-compose": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "html-webpack-plugin": "^2.30.1",
     "is-directory": "^0.3.1",
     "javascript-stringify": "^1.6.0",
-    "jest-cli": "^22.0.0",
     "jss": "^9.3.3",
     "jss-camel-case": "^6.0.0",
     "jss-compose": "^5.0.0",
--- a/scripts/__tests__/config.spec.js
+++ b/scripts/__tests__/config.spec.js
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import path from 'path';
 import getConfig from '../config';
-import config from '../schemas/config';
+import consts from '../consts';
 
 const cwd = process.cwd();
 afterEach(() => {
@@ -199,9 +199,9 @@ it('should allow no webpack config', () => {
 	expect(fn).not.toThrow();
 });
 
-it('should return defaultEditorConfig if nothing passed', () => {
+it('should return EDITOR_CONFIG if nothing passed', () => {
 	const result = getConfig();
-	expect(result.editorConfig).toEqual(config.defaultEditorConfig);
+	expect(result.editorConfig).toEqual(consts.EDITOR_CONFIG);
 });
 
 it('should return apply user preferences', () => {
@@ -211,7 +211,7 @@ it('should return apply user preferences', () => {
 		},
 	});
 	expect(result.editorConfig).toEqual({
-		...config.defaultEditorConfig,
+		...consts.EDITOR_CONFIG,
 		mode: 'js',
 	});
 });
@@ -222,7 +222,7 @@ it('should allow user to pass highlightTheme', () => {
 		editorConfig: {},
 	});
 	expect(result.editorConfig).toEqual({
-		...config.defaultEditorConfig,
+		...consts.EDITOR_CONFIG,
 		theme: 'dracula',
 	});
 });
@@ -235,7 +235,7 @@ it('should allow override highlightTheme', () => {
 		},
 	});
 	expect(result.editorConfig).toEqual({
-		...config.defaultEditorConfig,
+		...consts.EDITOR_CONFIG,
 		theme: 'base-16-dark',
 	});
 });
--- a/scripts/__tests__/config.spec.js
+++ b/scripts/__tests__/config.spec.js
@@ -1,7 +1,6 @@
 import fs from 'fs';
 import path from 'path';
 import getConfig from '../config';
-import consts from '../consts';
 
 const cwd = process.cwd();
 afterEach(() => {
@@ -199,9 +198,10 @@ it('should allow no webpack config', () => {
 	expect(fn).not.toThrow();
 });
 
-it('should return EDITOR_CONFIG if nothing passed', () => {
+it('should return editorConfig if nothing passed', () => {
 	const result = getConfig();
-	expect(result.editorConfig).toEqual(consts.EDITOR_CONFIG);
+	expect(result.editorConfig).toHaveProperty('mode', 'jsx');
+	expect(result.editorConfig).toHaveProperty('theme', 'base16-light');
 });
 
 it('should return apply user preferences', () => {
@@ -210,32 +210,7 @@ it('should return apply user preferences', () => {
 			mode: 'js',
 		},
 	});
-	expect(result.editorConfig).toEqual({
-		...consts.EDITOR_CONFIG,
-		mode: 'js',
-	});
-});
 
-it('should allow user to pass highlightTheme', () => {
-	const result = getConfig({
-		highlightTheme: 'dracula',
-		editorConfig: {},
-	});
-	expect(result.editorConfig).toEqual({
-		...consts.EDITOR_CONFIG,
-		theme: 'dracula',
-	});
-});
-
-it('should allow override highlightTheme', () => {
-	const result = getConfig({
-		highlightTheme: 'dracula',
-		editorConfig: {
-			theme: 'base-16-dark',
-		},
-	});
-	expect(result.editorConfig).toEqual({
-		...consts.EDITOR_CONFIG,
-		theme: 'base-16-dark',
-	});
+	expect(result.editorConfig).toHaveProperty('mode', 'js');
+	expect(result.editorConfig).toHaveProperty('lineWrapping', true);
 });
--- a/scripts/__tests__/config.spec.js
+++ b/scripts/__tests__/config.spec.js
@@ -198,13 +198,13 @@ it('should allow no webpack config', () => {
 	expect(fn).not.toThrow();
 });
 
-it('should return editorConfig if nothing passed', () => {
+it('editorConfig option should have default values', () => {
 	const result = getConfig();
 	expect(result.editorConfig).toHaveProperty('mode', 'jsx');
 	expect(result.editorConfig).toHaveProperty('theme', 'base16-light');
 });
 
-it('should return apply user preferences', () => {
+it('should merge default editorConfig with options provided by the user', () => {
 	const result = getConfig({
 		editorConfig: {
 			mode: 'js',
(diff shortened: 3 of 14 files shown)


SaraVieira:
> This is a breaking change, we should keep the old option until the next major release. See updateExample option in the config schema.

I left the option there :)

> editorConfig would be a more generic name, and better match existing options like compilerConfig.

Okay , makes sense

sapegin:
> I left the option there :)

I still think we should add a deprecation warning, it’s not good to have two valid ways of changing one option.

SaraVieira:
Name has been changed sir !

How should I add that warning ? :)

sapegin:
You need to add `deprecated` field with a recommendation for a new option:

https://github.com/styleguidist/react-styleguidist/blob/043ba230d5238e3e859a4c1521a42d42f3d06fe6/scripts/utils/sanitizeConfig.js#L91-L92

sapegin:
You also need to merge the old option with the new object in the config, we shouldn’t access the old option anywhere else.

SaraVieira:
Yellow !

There was a problem with the logger so I ended up using chalk with a yellow color it looks like this:

screen shot 2017-11-12 at 15 40 11

scripts/utils/sanitizeConfig.js
```diff
@@ -89,7 +88,8 @@ module.exports = function sanitizeConfig(config, schema, rootDir) {
 				throw new StyleguidistError(message);
 			}
 		} else if (props.deprecated) {
-			logger.warn(`${chalk.bold(key)} config option is deprecated. ${props.deprecated}`);
+			// eslint-disable-next-line no-console
+			console.log(chalk.yellow.bold(`${key} config option is deprecated. ${props.deprecated}`));
```
sapegin: Any reason to change this?

src/rsg-components/Editor/Editor.js
```diff
@@ -11,17 +11,18 @@ require('!!../../../loaders/style-loader!../../../loaders/css-loader!codemirror/
...
+
+const codeMirrorDefault = {
 	mode: 'jsx',
-	lineNumbers: false,
 	lineWrapping: true,
 	smartIndent: false,
 	matchBrackets: true,
 	viewportMargin: Infinity,
+	lineNumbers: false,
+	theme: 'base16-light',
```
sapegin: Now we have the same defaults object two times. Keep it in the config only and merge using `process` method.

src/rsg-components/Editor/Editor.js
```diff
@@ -46,9 +47,10 @@ export default class Editor extends Component {
 
 	render() {
 		const { code } = this.props;
-		const { highlightTheme } = this.context.config;
+		const { highlightTheme, editorConfig } = this.context.config;
```
sapegin: We should not use `highlightTheme` anywhere except the config schema. Now you’re overwriting the right option with the wrong one anyway.

scripts/utils/sanitizeConfig.js
SaraVieira: It simply did not log anything

src/rsg-components/Editor/Editor.js
SaraVieira: I'm sorry , what do you mean ? :(

Too many words

src/rsg-components/Editor/Editor.js
SaraVieira: Because I thought this is where you meant fpr me to merge

src/rsg-components/Editor/Editor.js
sapegin: 1. You have this object twice: here and in the config schema. We need only one in the schema.
2. Check how `process` method is used in the schema, you need something similar. You also need it to merge the old option with the new one.

scripts/utils/sanitizeConfig.js
sapegin: We need to find what’s the issue.

src/rsg-components/Editor/Editor.js
sapegin: > You also need to merge the old option with the new object **in the config**, we shouldn’t access the old option anywhere else.

scripts/utils/sanitizeConfig.js
sapegin: It was actually a bug, I’ve fixed it in master and updated your branch — should work now.

sapegin:
@SaraVieira Hey Sara, any plans to finish? It’s almost done 🦄

SaraVieira:
Hello ! Sorry for the time it passed :(

I was in london and super busy :(

It's not repeated anymore :)

About the process .... should it get the value i+and if none is return the hightlight theme one and the defaults ?

sapegin:
1. Merge value of `highlightTheme` option to theme field.
2. Merge value of `editorConfig` with default values object (it shouldn’t be in `default` field because we want to merge user options with defaults, not overwrite defaults).

Something like this:

```
editorConfig: {
  // ...
  process: (value, config) => {
    return Object.assign({}, { theme: config.highlightTheme }, value);
  },
},
```

SaraVieira:
Like this ?

sapegin:
Almost: defaults should be in the schema.

SaraVieira:
About now ? :p

sapegin:
One more time: 

> it shouldn’t be in default field because we want to merge user options with defaults, not overwrite defaults

You need to define a variable in the `schemas/config.js` file and merge user options with it.

SaraVieira:
screen shot 2017-12-18 at 16 44 39

sapegin:
Where to look?

SaraVieira:
Just wanted to make sure this is what you wanted:
https://github.com/styleguidist/react-styleguidist/pull/662/commits/e1496dcf8a0d3fbb3fbe71c53a87cd57695a4715

sapegin:
Nope ;-( You’re trying to store some data in the schema. There’s shouldn’t be a config option `defaultEditorConfig`, just a variable will be enough.

sapegin:
Probably will be easier to start from writing tests (you’ll need to do that anyway) ;-)

scripts/__tests__/config.spec.js
```diff
@@ -7,6 +7,16 @@ afterEach(() => {
 	process.chdir(cwd);
 });
```
SaraVieira: Tests are in this file

scripts/__tests__/config.spec.js
```diff
@@ -7,6 +7,16 @@ afterEach(() => {
...
 
+const defaultEditorConfig = {
```
sapegin: You don’t need to repeat it in tests. You can export it from the schema.

scripts/__tests__/config.spec.js
```diff
@@ -197,3 +207,46 @@ it('should allow no webpack config', () => {
 	const fn = () => getConfig();
 	expect(fn).not.toThrow();
 });
+
+it('should return defaultEditorConfig if nothing passed', () => {
+	const result = getConfig({
+		editorConfig: {},
```
sapegin: Nothing would be undefined, this is explicit empty object passed by the user, whatever the reason ;-)

scripts/schemas/config.js
```diff
@@ -95,19 +95,25 @@ module.exports = {
...
+				smartIndent: false,
+				matchBrackets: true,
+				viewportMargin: Infinity,
+				lineNumbers: false,
+			};
+
+			return Object.assign(
+				{},
+				defaultEditorConfig,
+				{ theme: config.highlightTheme || defaultEditorConfig.theme },
```
sapegin: `config.highlightTheme && { theme: config.highlightTheme }`

SaraVieira: Done :)

scripts/__tests__/config.spec.js
SaraVieira: Duh ! Fixed :)

sapegin:
Did you forget to push?

SaraVieira:
Pushed , sorry

scripts/schemas/config.js
```diff
@@ -17,7 +17,18 @@ const getUserPackageJson = require('../utils/getUserPackageJson');
...
+	mode: 'jsx',
+	lineWrapping: true,
+	smartIndent: false,
+	matchBrackets: true,
+	viewportMargin: Infinity,
+	lineNumbers: false,
+};
+
 module.exports = {
+	defaultEditorConfig,
```
sapegin: Sorry, it was a bad idea — now we have the same issue again with storing random data as config options. Let’s try to put it inside `editorConfig` with a name like `_default` ;-|

SaraVieira: I moved it to the consts file , is that okay ?

package.json
```diff
@@ -53,6 +53,7 @@
     "html-webpack-plugin": "^2.30.1",
     "is-directory": "^0.3.1",
     "javascript-stringify": "^1.6.0",
+    "jest-cli": "^22.0.0",
```
sapegin: `jest` already includes CLI, and we’re using Jest 21 because we need to run tests in Node 4.

src/rsg-components/Editor/Editor.spec.js
```diff
@@ -32,6 +35,33 @@ describe('Editor', () => {
 		expect(actual).toMatchSnapshot();
 	});
 
+	it('should allow user to use highlightTheme in editor', () => {
```
sapegin: At this point you don’t have `highlightTheme` option.

src/rsg-components/Editor/Editor.spec.js
```diff
@@ -32,6 +35,33 @@ describe('Editor', () => {
...
+		const actual = shallow(<Editor code={code} onChange={() => {}} />, options);
+
+		expect(actual).toMatchSnapshot();
+	});
+
+	it('should allow override user with editorConfig', () => {
```
sapegin: At this point you don’t know which options are provided by the user and which are default.

scripts/consts.js
```diff
@@ -8,4 +8,13 @@ module.exports = {
 	DOCS_WEBPACK: 'https://react-styleguidist.js.org/docs/webpack.html',
 	DOCS_DOCUMENTING: 'https://react-styleguidist.js.org/docs/documenting.html',
 	DOCS_THIRDPARTIES: 'https://react-styleguidist.js.org/docs/thirdparties.html',
+	EDITOR_CONFIG: {
```
sapegin: Why do you think it’s better than `editorConfig._default`?

package.json
SaraVieira: The problem is that I cant run the tests :(

scripts/consts.js
SaraVieira: Mainly because it's being used same place as all consts and it is a constant but I will change it if you think its better

sapegin: My reasoning: it’s actually used only in the config schema and the only reason to export it is to reuse it in the test for the config schema. That’s why I think it’s better to keep it close to the schema. But we can’t export it directly, so we’ll have to put it inside the `editorConfig` object.

src/rsg-components/Editor/Editor.spec.js
SaraVieira: yes I do , it's still being used if the user passes it :o

src/rsg-components/Editor/Editor.spec.js
SaraVieira: what do you mean ?

package.json
sapegin: Then we need to find how to fit *that* problem, not change the project configuration that works well for other people and CI. Ping me somewhere ;-)

src/rsg-components/Editor/Editor.spec.js
sapegin: Technically yes, but you shouldn’t access it anywhere except the config schema where you merge it with the new option.

package.json
SaraVieira: I'm not gonna change this , I was gonna remove it before this commit

src/rsg-components/Editor/Editor.spec.js
sapegin: You’re merging user value with defaults in the config schema, here you only have the resulting object. The Editor component doesn’t (and shouldn’t) know the difference.

scripts/consts.js
SaraVieira: should it be in here:

```
	editorConfig: {
		_defaults: {
			theme: 'base16-light',
			mode: 'jsx',
			lineWrapping: true,
			smartIndent: false,
			matchBrackets: true,
			viewportMargin: Infinity,
			lineNumbers: false,
		},
		type: 'object',
		process: (value, config) => {
```
`?

src/rsg-components/Editor/Editor.spec.js
SaraVieira: nevermind, I was thinking of another test.  Removed this one , yes , it doesn't make sense

src/rsg-components/Editor/Editor.spec.js
SaraVieira: should I discard this test then ?

scripts/consts.js
sapegin: Yup. Should work ;-£

src/rsg-components/Editor/Editor.spec.js
sapegin: It doesn’t test anything ;-)

SaraVieira: good point :p

scripts/consts.js
SaraVieira: Makes sense , but how do I access it ?

package.json
sapegin: That’s an interesting way to solve problems. But please comment that next time.

scripts/consts.js
sapegin: Actually I think we’re solving a wrong issue here. You don’t really need the whole thing in tests, I’d check one field in tests ;-)

package.json
SaraVieira: I shall , sorry

SaraVieira:
Left the defaults as a variable inside the process and just checked with https://facebook.github.io/jest/docs/en/expect.html#tohavepropertykeypath-value

scripts/__tests__/config.spec.js
```diff
@@ -197,3 +197,20 @@ it('should allow no webpack config', () => {
 	const fn = () => getConfig();
 	expect(fn).not.toThrow();
 });
+
+it('should return editorConfig if nothing passed', () => {
```
sapegin: “editorConfig option should have default values” would be more clear I think.

scripts/__tests__/config.spec.js
```diff
@@ -197,3 +197,20 @@ it('should allow no webpack config', () => {
...
+	const result = getConfig();
+	expect(result.editorConfig).toHaveProperty('mode', 'jsx');
+	expect(result.editorConfig).toHaveProperty('theme', 'base16-light');
+});
+
+it('should return apply user preferences', () => {
```
sapegin: I don’t understand what that means. “should merge default editorConfig with options provided by the user” — long but clear.

scripts/__tests__/config.spec.js
```diff
@@ -197,3 +197,20 @@ it('should allow no webpack config', () => {
...
+	const result = getConfig({
+		editorConfig: {
+			mode: 'js',
+		},
+	});
+
+	expect(result.editorConfig).toHaveProperty('mode', 'js');
+	expect(result.editorConfig).toHaveProperty('lineWrapping', true);
```
sapegin: I’d check for mode and theme in both tests to make the difference more clear.

scripts/schemas/config.js
```diff
@@ -93,6 +93,27 @@ module.exports = {
 	highlightTheme: {
 		type: 'string',
 		default: 'base16-light',
+		deprecated: 'Use the theme property in the editorConfig option instead',
+	},
+	editorConfig: {
+		type: 'object',
+		process: (value, config) => {
+			const _defaults = {
```
sapegin: No need for `_` here.

SaraVieira:
@sapegin All changes made and the lock is reverted :)

sapegin:
It still shows package-lock in the diff ;-|

SaraVieira:
Reverted now , don't know why it wasn't working :/

sapegin:
Better to review all changes you’re going to commit and uncheck files / lines you don’t need.

Let’s wait for CI 🙏

SaraVieira:
The CI can't find jest :(

sapegin:
We need to find what’s the issue because it works in master.

SaraVieira:
I know , the weirdest is that I didn't change anything on any files that have anything to do with config :(

SaraVieira:
@sapegin did you do anything ? The tests are now passing

sapegin:
I’ve cleared Travis cache and restarted them.

sapegin:
The very last thing: docs! 📖

docs/Configuration.md
```diff
@@ -227,11 +227,22 @@ module.exports = {
 }
 ```
 
-#### `highlightTheme`
```
SaraVieira: Doc changes here :)

docs/Configuration.md
```diff
@@ -227,11 +227,22 @@ module.exports = {
...
+#### `editorConfig`
 
-Type: `String`, default: `base16-light`
+Type: `Object`, default:
+```
```
sapegin: I’d put a link to the source instead of an object.

docs/Configuration.md
```diff
@@ -17,7 +17,7 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 * [`getComponentPathLine`](#getcomponentpathline)
 * [`getExampleFilename`](#getexamplefilename)
 * [`handlers`](#handlers)
-* [`highlightTheme`](#highlighttheme)
+* [`editorConfig`](#editorConfig)
```
sapegin: They aren’t ordered now.

docs/Configuration.md
```diff
@@ -227,11 +227,22 @@ module.exports = {
...
+  lineWrapping: true,
+  smartIndent: false,
+  matchBrackets: true,
+  viewportMargin: Infinity,
+  lineNumbers: false,
+}
+```
 
-[CodeMirror theme](http://codemirror.net/demo/theme.html) name to use for syntax highlighting in the editor.
+This will determine how your editor looks, you can check [here](https://codemirror.net/doc/manual.html#config) all the available options
```
sapegin: Shorter at least:

> Source code editor options, see [CodeMirror docs](https://codemirror.net/doc/manual.html#config) for all available options.

docs/Configuration.md
```diff
@@ -168,6 +168,12 @@ module.exports = {
 }
 ```
 
+#### `editorConfig`
+
+Type: `Object`, default: [scripts/schemas/config.js](https://github.com/styleguidist/react-styleguidist/blob/87efdde2d52209e9c4072368402d11030943d69f/scripts/schemas/config.js#L101)
```
sapegin: Better to point to master in case we change the options.

SaraVieira:
Docs updated! Sorry , didn't notice it was in alphabetic order 🤦‍♀️

SaraVieira:
That was dumb of me 🤦‍♀️

Done !! 🙂

sapegin:
Awesome! Praying for CI again 🙏

SaraVieira:
Weird, some links are failing in the snapshots only on node 4

sapegin:
Yup, that’s very weird. We have that in master too. Restart often helps.

https://github.com/styleguidist/react-styleguidist/pull/725#issuecomment-352754008

sapegin:
Merged! :shipit:

SaraVieira:
YES 🌈

Issue #576, opened 2017-08-14
Incorrect margins on error message of noeditor example


sapegin:
![image 2017-08-14 at 9 51 26 pm](https://user-images.githubusercontent.com/70067/29288887-d8275998-813a-11e7-8416-45c2610032ec.png)

Right now the negative margin is [in the PlaygroundErrorRenderer](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/PlaygroundError/PlaygroundErrorRenderer.js#L7) which is not good but used to work before we’ve introduced `noeditor` modifier. Would be cool to find a way to avoid negative margins entirely but I can’t think of any now.

SaraVieira:
@sapegin Can you give me an exaample code to make this happen , I can't reproduce

stepancar:
@SaraVieira create example with noeditor option
```md
```jsx noeditor
<Button dfsdfsd </Button>

stepancar:
![image](https://user-images.githubusercontent.com/2854910/29689163-dedab4fe-892a-11e7-89b9-90bbece77d30.png)

SaraVieira:
@stepancar Nevermind , I was making the errors in the js :(

whazap:
i tried to tackle this issue by creating a new helper component (PreviewRenderer).
(was looking for an issue to start looking under the hood of react-styleguidist, hope i didn't mess up too much ;))

PR #667, opened 2017-11-03
Refactor: Add PreviewRenderer

whazap:
* Moves preview styles from PlaygroundRenderer to PreviewRenderer
* Changes padding of PreviewRenderer if there is an error

Resolves: #576


sapegin:
Adding more complexity isn’t the right way to solve the issue, let’s make it simpler — #673.

Issue #464, opened 2017-06-04
Automated Visual Testing


kof:
We spoke about it once, I though it might be good to at least kick off  the discussion here.

So the idea is that we can trigger visual snapshot testing which will essentially create graphics for each demo so that next time you run it it can compare and report mismatches. 

I think react-styleguidist is a good place to do this because it already has lots of configured components if one uses it havily, so it would not cost much for the user to use this.

List of screenshot tools:

- https://github.com/garris/BackstopJS
- https://github.com/segmentio/niffy

okonet:
I was thinking about this lately and would be interested in this. @kof let's pair on the implementation?

kof:
Lets do the research, which tools are the best, who is maintaining them, maybe they can help integrating.

sapegin:
Looks like you need an API to return all examples for this?

n1313:
You're looking for something like backstopjs and you want to address examples by using existing data-attributes. They are there precisely for the purposes of integration RSG with external tools like this.

sapegin:
https://screener.io/ — visual testing for Storybook.

kof:
Looks nice, but a service …

sapegin:
But Storybook I’d say :-)

kof:
More questions:

1. Where would you want to store the images? Repository? 
1. How big can this grow? 
1. How can we snapshot in different browsers? Karma?

n1313:
I'm not sure if I understand the topic correctly. Are we talking about implementing a new feature into RSG, or about making RSG work well with existing third-party tools?

kof:
Depends on how we want it to work. A third-party tool might be also used as  a dependency and integrated into sg as well as sg might provide whatever is required to work with third-party tools nicely.

sapegin:
@n1313 I think both would work ;-) Do you know any third-party tools that we can use? Most likely it will be a plugin anyway ;-)

n1313:
RSG already has everything it needs to integrate with external visual regression tools: #238 #334. I don't know what a plugin would be needed for. Integrating an external tool into RSG sounds like a massive feature creep to me.

sapegin:
> RSG already has everything it needs to integrate with external visual regression tools: #238 #334.

Has everything *is not* integration. I have no idea what to do for example ;-)

> Integrating an external tool into RSG sounds like a massive feature creep to me.

Nobody is saying that and it won’t happen. That’s why the plugin API is so important.

[...]

PR #670, opened 2017-11-06
Add Yammer to showcase list

unindented:
As suggested here by @sapegin: https://github.com/styleguidist/react-styleguidist/issues/464#issuecomment-342114531


sapegin:
Thanks!

PR #671, opened 2017-11-06
Make styleguidist server publicly accessible

leclercb:
Add doc: "How to make styleguidist server publicly accessible?"

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -257,6 +257,8 @@ module.exports = {
 
 ## How to make styleguidist server publicly accessible?
 
+Warning: this setup is not recommended for security reasons, it should only be enabled on dev environments.
+
 In order to make styleguidist server publicly accessible, add the following to the styleguide.config.js file:
 
 ```javascript
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -14,6 +14,7 @@
 - [How to change style guide dev server logs output?](#how-to-change-style-guide-dev-server-logs-output)
 - [How to debug my components and examples?](#how-to-debug-my-components-and-examples)
 - [How to debug the exceptions thrown from my components?](#how-to-debug-the-exceptions-thrown-from-my-components)
+- [How to make styleguidist server publicly accessible?](#how-to-make-styleguidist-server-publicly-accessible)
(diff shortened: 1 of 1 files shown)


docs/Cookbook.md
```diff
@@ -254,6 +255,21 @@ module.exports = {
 
 ![](https://d3vv6lp55qjaqc.cloudfront.net/items/3i3E3j2h3t1315141k0o/debugging.png)
 
+## How to make styleguidist server publicly accessible?
+
+In order to make styleguidist server publicly accessible, add the following to the styleguide.config.js file:
```
sapegin: * styleguidist → Styleguidist

docs/Cookbook.md
```diff
@@ -257,6 +257,8 @@ module.exports = {
 
 ## How to make styleguidist server publicly accessible?
 
+Warning: this setup is not recommended for security reasons, it should only be enabled on dev environments.
```
sapegin: I don’t think the second part is necessary, I think it’s dangerous in any case ;-)

Also: `**Warning:**`

docs/Cookbook.md
```diff
@@ -257,6 +257,8 @@ module.exports = {
 
 ## How to make styleguidist server publicly accessible?
```
sapegin: styleguidist → Styleguidist

n1313:
It would be a good addition to add a link to the original issue https://github.com/webpack/webpack-dev-server/issues/887 to illustrate what exactly the security risk here is about.

docs/Cookbook.md
```diff
@@ -254,6 +255,23 @@ module.exports = {
...
+
+Warning: this setup is not recommended for security reasons, it should only be enabled on dev environments.
+
+In order to make styleguidist server publicly accessible, add the following to the styleguide.config.js file:
+
+```javascript
+module.exports = {
+  webpackConfig: {
+    ...
+    devServer: {
```
roblevintennis: Fwiw we just had to do this. We do a check `!production` in our whole dev-server config so it's "safe" I suppose. Yeah, I think this is helpful as I spent a good hour digging in to figure this out myself :) 👍

roblevintennis: Hrm, actually, we're electing NOT to do this in favor of something more like:
```
const sg = process.env.STYLEGUIDE === 'true';
...
} else if (sg) {
  config.output.publicPath = '//path.local/';

in our main webpack.config.js

Issue #665, opened 2017-11-03
Also include sections for unrecognized components


jfschwarz:
Due to the limitations of 'react-docgen` there are components which will not be recognized, e.g., when I import a component, apply a HOC, and export the result. For these components Styleguidist will generate warnings like:

> Warning: src/components/EnhancedComponent.js matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.

Why doesn't Styleguidist simply include sections for these unrecognized components in the generated style guide, too?

I get that we won't be able to get the props documentation, but at least that component would be listed as part of style guide and we could add docs and examples for it. That seems better than just omitting it altogether.

sapegin:
Sounds reasonable ;-)

Would be nice if you can comment out this code and check if it would actually work.

PR #672, opened 2017-11-07
Include unrecognized components

jfschwarz:
Implements #665: Components that react-docgen does not recognize as react-components will still be included in the style guide. The `displayName` for such components is guessed based on the file name.

I also added an example of an unrecognizable component to the basic example.

### DIFF
--- a/examples/basic/src/components/WrappedCounterButton/Readme.md
+++ b/examples/basic/src/components/WrappedCounterButton/Readme.md
@@ -1,6 +1,0 @@
-react-docgen is not able to parse prop types of enhanced components that are imported from other files.
-We'll add style guide sections for such components nevertheless.
-
-```jsx
-<WrappedCounterButton>I'm a wrapped button</WrappedCounterButton>
-```
--- a/examples/basic/src/components/WrappedCounterButton/Readme.md
+++ b/examples/basic/src/components/WrappedCounterButton/Readme.md
@@ -1,0 +1,6 @@
+react-docgen is not able to parse prop types of enhanced components that are imported from other files.
+We'll add style guide sections for such components nevertheless.
+
+```jsx
+<WrappedCounterButton>I'm a wrapped button</WrappedCounterButton>
+```
--- a/loaders/utils/getNameFromFilePath.js
+++ b/loaders/utils/getNameFromFilePath.js
@@ -1,0 +1,16 @@
+'use strict';
+const path = require('path');
+
+module.exports = function getNameFromFilePath(filePath) {
+	let displayName = path.basename(filePath, path.extname(filePath));
+	if (displayName === 'index') {
+		const parts = path.dirname(filePath).split(path.sep);
+		displayName = parts[parts.length - 1];
+	}
+
+	return displayName
+		.charAt(0)
+		.toUpperCase()
+		.concat(displayName.slice(1))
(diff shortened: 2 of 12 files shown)


loaders/utils/getProps.js
```diff
@@ -102,5 +102,11 @@ module.exports = function getProps(doc, filepath) {
 		});
 	}
 
+	if (!doc.displayName && filepath) {
+		// The file name is our best guess for the exported component's display name
```
sapegin: I think we should use the folder name if the file name is `index.*`.

examples/basic/src/components/WrappedCounterButton/Readme.md
```diff
@@ -0,0 +1,6 @@
+react-docgen is not able to parse prop types of enhanced components that are imported from other files.
```
sapegin: I’d move this component to the `sections` example, it’s too “low-level” for the basic example.

examples/basic/src/components/WrappedCounterButton/Readme.md
```diff
@@ -0,0 +1,6 @@
...
+We'll add style guide sections for such components nevertheless.
```
sapegin: “Styleguidist will add them to your style guide anyway” or something like that. They are not *sections*.

loaders/utils/getProps.js
jfschwarz: good idea

jfschwarz:
It will still show the warning.

loaders/utils/getProps.js
sapegin: Or even better:

https://github.com/nerdlabs/react-docgen-displayname-handler/blob/8cd141064910eb429f18e03e631a25cec371330a/source/index.js#L51-L66

sapegin:
> It will still show the warning.

That’s good I think. Probably we should mention the `ignore` option there too. But it may be another PR ;-)

jfschwarz:
@sapegin I've implemented the changes you requested.

loaders/utils/getNameFromFilePath.js
```diff
@@ -0,0 +1,16 @@
...
+		const parts = path.dirname(filePath).split(path.sep);
+		displayName = parts[parts.length - 1];
+	}
+
+	return displayName
+		.charAt(0)
+		.toUpperCase()
+		.concat(displayName.slice(1))
+		.replace(/-([a-z])/, (_, match) => match.toUpperCase());
+};
```
sapegin: `_.upperFirst(_.camelCase(displayName))` would do the same but more readable.

loaders/utils/getNameFromFilePath.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+const path = require('path');
+
+module.exports = function getNameFromFilePath(filePath) {
+	let displayName = path.basename(filePath, path.extname(filePath));
+	if (displayName === 'index') {
+		const parts = path.dirname(filePath).split(path.sep);
+		displayName = parts[parts.length - 1];
```
sapegin: Isn’t it the same as `path.basename(path.dirname(filePath))`?

jfschwarz:
@sapegin Sure, no problem.

sapegin:
Looks good now, merged! 🍕

Issue #467, opened 2017-06-06
Back / forward browser buttons don’t work properly


sapegin:
Was found by @n1313:

1. Click on "open isolated" button for the first example of the Button component.
2. Click "back" in the browser — **see that the "Lick me" example is no longer displayed on the page, and that the "open isolated" button for the first example has the wrong icon**.
3. Click "forward" in the browser — **the page shifts to the left (??) and nothing happens**.

There is also an error in the console:

> main.bundle.js:26222 Uncaught TypeError: Cannot read property 'type' of undefined

We may need to introduce some proper routing to fix this kind of issues. I don’t mean React Rooter, but something very simple but isolated, not hardcoded in the `index.js`.

sapegin:
Routing is a big issue described in #494 but we can probably fix an error here.

PR #676, opened 2017-11-07
Fix: Navigation between list and isolated mode issues

### LINKED ISSUE #663: Navigation is broken
Reproducible on the public example styleguide: https://react-styleguidist.js.org/examples/basic/
~~Not reproducible on `dialog`'s styleguide: https://dialogs.github.io/dialog-web-components/~~ Can also repro there

Issue is that navigation is super broken. See the GIF:

out

- Navigating to a preview and then back results in other previews from the same section disappearing
- Trying to navigate from a preview back to the main page by clicking the little icon doesn't work
- Navigating directly to a preview by changing the URL results in a nasty error
- Back button on the browser seems to do nothing

Why does this not reproduce on `dialog`'s styleguide? How is it different from your basic example?

sapegin:
1. Navigate to isolated example and back: you’ll see only this one
example, not all of them.
2. Navigate to isolated example of a component with only one example:
you’ll see isolated mode button below this example instead of a show
all components button.

Browser back / forward buttons also seems to work.

Closes #663, #467


unindented:
Thank you!

src/rsg-components/Section/Section.spec.js
```diff
@@ -6,6 +6,12 @@ import Sections from '../Sections';
 import Section from './Section';
 import { SectionRenderer } from './SectionRenderer';
 
+const options = {
+	context: {
+		displayMode: 'all',
```
okonet: I'd create a enum (`Object.freeze`) with all available modes and import it any time you reference it.

src/rsg-components/Section/Section.js
```diff
@@ -31,5 +31,5 @@ Section.propTypes = {
 };
 
 Section.contextTypes = {
-	isolatedSection: PropTypes.bool,
+	displayMode: PropTypes.string,
```
okonet: Instead of saying `string` it's better to use `oneOf` propType with exhausting set of values.

sapegin:
Yeah, good points. I decided to work on that more and was able isolate, fix and test mutation instead of using a hack with `cloneDeep`.

By the way how would you ensure that you’re not mutating source objects in tests?

okonet:
Not sure about your questions. My points are more about the maintainability. What do you mean by mutating in test? Example?

sapegin:
I mean tests to prevent mutation in code. That was exactly the issue here.

okonet:
Ah no, I don't know how to tests side-effects.

src/rsg-components/Section/Section.spec.js
sapegin: Do you have an example somewhere?

sapegin:
I’ve added deepfreeze to tests, so it should catch mutations. And I’ve removed cloneDeep hack. 

I still need to replace strings with an enum.

sapegin:
Merged manually.

Issue #677, opened 2017-11-09
Pass process.env.NODE_ENV to webpack.DefinePlugin


sapegin:
1. Pass `process.env.NODE_ENV` to `webpack.DefinePlugin` instead of Styleguidist environment (server → development, build → production):

https://github.com/styleguidist/react-styleguidist/blob/043ba230d5238e3e859a4c1521a42d42f3d06fe6/scripts/make-webpack-config.js#L44-L48

2. Add `STYLEGUIDIST_ENV` `webpack.DefinePlugin` that equals Styleguidist environment.

3. Use `STYLEGUIDIST_ENV` to check if we’re in server or build mode:

https://github.com/styleguidist/react-styleguidist/blob/043ba230d5238e3e859a4c1521a42d42f3d06fe6/src/rsg-components/ReactComponent/ReactComponent.js#L12-L15

PR #679, opened 2017-11-11
Fix node env

WaldoJeffers:
This fixes https://github.com/styleguidist/react-styleguidist/issues/677

- Add a new environment variable called `STYLEGUIDIST_ENV` to check if we're in dev server or static build mode
- Allow the user to pass `NODE_ENV` to Webpack's config, to determine if React will use its production or development build

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -262,7 +262,7 @@ module.exports = {
 3. Press the ![Continue](https://d3vv6lp55qjaqc.cloudfront.net/items/3b3c1P3g3O1h3q111I2l/continue.png) button and the debugger will stop execution at the next exception.
 
 ## How to use React's production or development build?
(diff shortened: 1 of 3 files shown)


docs/Cookbook.md
```diff
@@ -260,6 +261,32 @@ module.exports = {
...
+};
+```
+If you use code similar to this, you might encounter errors such as `Cannot read property 'isRequired' of undefined`.
+
+To avoid this, you need to tell React Styleguidist to use React's development build. To do this, simply set the `NODE_ENV` variable to `development` in your npm script.
+
+```json
+{
+  "scripts": {
+    "build": "NODE_ENV=development react-styleguidist build"
```
sapegin: Would be nice to use cross-env here.

WaldoJeffers: Good point

WaldoJeffers:
@sapegin Wait, please **do not merge**. When doing a more thorough test, I am now unavailable to replicate the results I was obtaining using `dangerouslyUpdateWebpackConfig` and definePlugin manual override. Will link into that.

WaldoJeffers:
@sapegin actually nevermind. I was seeing weird results because of `npm link`. Has anyone else encountered issues with that? Had to publish my branch on a scoped npm package, to make sure everything was running smoothly.

sapegin:
Merged, thanks!

Issue #678, opened 2017-11-11
Styleguide previews have extra unstyled wrappers


gscottolson:
Thanks for working so hard on React Styleguidist. I am loving what I am seeing.

I want to know if there is a quick fix to this issue. I am happy to open a PR with solution, but maybe I am missing something obvious.

I want to style a preview to be a specific height. My embedded React component uses`height: 100%`, which means it relies on the parent element’s height. I can style the `example wrapper` as noted in the docs:

<img width="447" alt="2017-11-10-screenshot-1" src="https://user-images.githubusercontent.com/220979/32684207-5c1ec49e-c636-11e7-9c5a-1f70a017f13c.png">

My custom styles look like:

` ```jsx { "props": { "style": { "height": "80vh" } } } ...`

This works as noted, but there are two nested `<div>`s inside of the `<div class="rsg--preview">` container. Since neither of these `<div>`s provide height styles (e.g. `height: inherit;`), the preview’s height is different than the components. Here is a screenshot from Chrome Dev Tools:

<img width="571" alt="2017-11-10-screenshot" src="https://user-images.githubusercontent.com/220979/32684145-bbb3a8f8-c635-11e7-8f5f-067ac9787568.png">

It seems like a tough problem to solve, but maybe there is a quick win here. Like passing the styles to the preview wrapper to each of the nested child elements. Or maybe setting the child elements to `height: inherit` if `height` is set on the preview.

Thoughts?

sapegin:
Yeah, it won’t be easy. These two divs are rendered here:

https://github.com/styleguidist/react-styleguidist/blob/29b8f9039419c43b368eaeea88fce40950700a72/src/rsg-components/Preview/Preview.js#L153-L154

This is also connected to a new plugin API — I may have made it worth with the new Container component — #505.

For your use case I’d suggest to try to replace the Wrapper component — it’s rendered inside these two empty divs.

Any ideas how to solve it in a generic way would be super cool! I’m not sure I know any good solutions now.

gscottolson:
Documenting my fix for others…

I was able to style these wrapper `<div>`s using JSS nested syntax:

```
module.exports = {
  ...
  styles: {
    Playground: {
      preview: {
        '& > div, & > div > div': {
          height: '100%'
        }
      }
    }
  }
  ...
}
```

sapegin:
I think we can close this.

PR #680, opened 2017-11-11
Apply styles to Preview `<div>`s when defined

gscottolson:
Fixes https://github.com/styleguidist/react-styleguidist/issues/678

This is a bit of a hack, but it achieves the desired result. Since `styleguide.config.js` contains a `styles` configuration, I am taking any defined Preview styles there and applying them to the two unclassed `<div>`s in `Preview.js` via the `context` object.

I see that this project leverages JSS, so I am open to making changes that more closely align with those patterns. However, without providing `className` hooks or turning each `<div>` into a first-class Styleguidist component, I think that solution will also lead to a bit of a mess. 😄

I am hoping that this quick solution creates a decent jumping off point for the discussion.

### DIFF
--- a/src/rsg-components/Preview/Preview.js
+++ b/src/rsg-components/Preview/Preview.js
@@ -149,9 +149,12 @@ export default class Preview extends Component {
 
 	render() {
 		const { error } = this.state;
+		const { styles } = this.context.config;
+		const previewStyle = styles && styles.Preview ? { style: styles.Preview } : null;
+
 		return (
-			<div>
-				<div ref={ref => (this.mountNode = ref)} />
+			<div {...previewStyle}>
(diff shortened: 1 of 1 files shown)


gscottolson:
I just noticed that you are using the `jss-nested` package, so this seems possible using only configuration:

**styleguide.config.js:**
```js
module.exports = {
  ...
  styles: {
    Playground: {
      preview: {
        '& > div': {
          height: '100%',
          '& > div': {
            height: '100%',
          }
        }
      }
    }
  }
}
```

sapegin:
I think it’s too much of a hack and will lead to unpredictable behavior because you’re repeating styles that supposed to be applied to a single node. So I’d recommend to use the snippet from your last comment.

gscottolson:
Agreed. Thanks!

PR #684, opened 2017-11-13
Add @return for PropTypes.func documentation 

ankri:
Adds missing functionality mentioned [here](https://github.com/styleguidist/react-styleguidist/issues/393#issuecomment-343899828)

@sapegin  Please note that I also modified the `ArgumentsRenderer` so that the dash will only be displayed when type **and** description are present. It will be omitted if one of them are left out.


einarlove:
Sweet! thank for quick response and PR! 🥇

einarlove:
Sweet! thank for quick response and PR! 🥇

einarlove:
Should it support the other synonyms like `@returns`? 
http://usejsdoc.org/tags-returns.html

src/rsg-components/Props/PropsRenderer.js
```diff
@@ -123,12 +124,15 @@ function renderDescription(prop) {
...
 	const args = [...(tags.arg || []), ...(tags.argument || []), ...(tags.param || [])];
+	const returnDocumentation = tags.return && tags.return[0];
+
 	return (
 		<div>
 			{description && <Markdown text={description} />}
 			{extra && <Para>{extra}</Para>}
 			<JsDoc {...tags} />
 			{args.length > 0 && <Arguments args={args} heading />}
+			{returnDocumentation && <Returns returnDocumentation={returnDocumentation} heading />}
```
sapegin: `Argument` component already supports returns, you don’t need to create a new component for that.

ankri: Tiny problem: `name` is required by `Argument` but `@return` does not have a name. 

The `key` also uses `name` but I could use the `title` as the `key` and make `name` optional.

ankri:
@einarlove Thanks for the reminder I will add support for `@returns`, too.

src/rsg-components/Props/PropsRenderer.js
sapegin: Name is optional and it doesn’t use `key` at all:

https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Argument/ArgumentRenderer.js

ankri: Sorry. it is require in [ArgumentsRenderer](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Arguments/ArgumentsRenderer.js#L38) and it is used as [key](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Arguments/ArgumentsRenderer.js#L29).

But if it is not required by `Argument` I could remove it from `ArgumentsRenderer` and use 
`key={props.name || props.title}`

sapegin: I still don’t understand why do you need `Arguments` here. `return` isn’t a part of arguments.

ankri: Aw man I feel so dumb, you are totally right sir. My bad. Time for some coffee

src/rsg-components/Props/PropsRenderer.js
```diff
@@ -124,15 +125,16 @@ function renderDescription(prop) {
...
 
 	return (
 		<div>
 			{description && <Markdown text={description} />}
 			{extra && <Para>{extra}</Para>}
 			<JsDoc {...tags} />
 			{args.length > 0 && <Arguments args={args} heading />}
-			{returnDocumentation && <Returns returnDocumentation={returnDocumentation} heading />}
+			{returnDocumentation && <Heading level={5}>Returns</Heading>}
+			{returnDocumentation && <Argument {...returnDocumentation} />}
```
sapegin: Could you try `<Argument returns…` instead of another heading for a more compact output?

sapegin:
Cool, thanks!

Issue #682, opened 2017-11-13
Changes in `assetDir` do not cause a reload in dev server


esturcke:
The dev server does not reload the page when files in `assetDir` change. I'm not sure if this is intentional, but I would find it helpful if the page would reload.  I think [`watchContentBase`](https://webpack.js.org/configuration/dev-server/#devserver-watchcontentbase) in `create-server.js` would get the dev server to reload.

This can be reproduced with [this repo](https://github.com/esturcke/styleguidist-no-reload-on-asset-change-bug):

1. `git clone git@github.com:esturcke/styleguidist-no-reload-on-asset-change-bug.git`
2. `cd git@github.com:esturcke/styleguidist-no-reload-on-asset-change-bug.git`
3. `yarn`
4. `yarn start`
5. Navigate to [http://localhost:6060/](http://localhost:6060/).
6. `cp public/{update,image}.png`

The page does not reflect the new image until the page is manually reloaded.

sapegin:
Could you try `watchContentBase` and check if it helps here? You can set it via webpack config.

esturcke:
Yup, this works:

```diff
diff --git a/public/image.png b/public/image.png
index f71184c..a7a4c53 100644
Binary files a/public/image.png and b/public/image.png differ
diff --git a/styleguide.config.js b/styleguide.config.js
index f08946b..9daf5bc 100644
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -4,6 +4,9 @@ module.exports = {
   components: './lib/components/**/[A-Z]*.js',
   assetsDir: 'public',
        webpackConfig: {
+               devServer: {
+                       watchContentBase: true,
+               },
                module: {
                        loaders: [
                                {
```

esturcke:
To keep using CRA's webpack config, it looks like I can do something like: 

```diff
diff --git a/styleguide.config.js b/styleguide.config.js
index f08946b..675270e 100644
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -18,4 +18,8 @@ module.exports = {
                        ]
                }
        },
+       dangerouslyUpdateWebpackConfig: config => ({
+               ...config,
+               devServer : { watchContentBase: true }
+       }),
 };
```

sapegin:
Cool! Do you want to send a PR that enables watchContentBase?

esturcke:
[...]

PR #685, opened 2017-11-13
Reload dev server when assets change

esturcke:
Address bug #682 by setting the webpack [`watchContentBase`](https://webpack.js.org/configuration/dev-server/#devserver-watchcontentbase) flag to `true` when an `assetDir` is provided.

### DIFF
--- a/scripts/create-server.js
+++ b/scripts/create-server.js
@@ -17,6 +17,7 @@ module.exports = function createServer(config, env) {
 			ignored: /node_modules/,
 		},
 		contentBase: config.assetsDir,
+		watchContentBase: config.assetsDir != null,
 		stats: webpackConfig.stats || {},
 	});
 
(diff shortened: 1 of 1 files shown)


scripts/create-server.js
```diff
@@ -17,6 +17,7 @@ module.exports = function createServer(config, env) {
 			ignored: /node_modules/,
 		},
 		contentBase: config.assetsDir,
+		watchContentBase: config.assetsDir != null,
```
sapegin: `!== undefined` would be better here. I’m wondering why ESLint isn’t complaining about that ;-)

esturcke: Sure, sounds good to be more explicit. I've been used to writing `!= null` as a short hand for `null` or `undefined` checks.

sapegin:
Cool, thanks!

PR #690, opened 2017-11-15
Allow overriding default webpack dev server options

jure:
There is currently no way to override webpack dev server options. For example, we specifically want the dev server to look at `node_modules/` too (to develop using linked dependencies) and there isn't a way to do that.


jure:
@sapegin I had a solution where I'd just play with the order of things in the merge(), but this seems more readable and explicit.

sapegin:
I’d actually add it to `merge` as the last argument — would be more readable and consistent.

jure:
Yeah, that's what I had, personally found the override a bit too implicit then. I'll do it, though, sure.

sapegin:
Thanks, merged!

PR #697, opened 2017-11-21
feat: minor content update (View Code instead of Code)

### LINKED ISSUE #693: Allowing "CODE" label to be changed via configuration?
While many places in Styleguidist are configurable, the one label that is giving trouble to our own customers is "CODE". While obvious to my teammates, it is not very clear to non-devs like Visual design or QA folks.

I've used the following configuration to reverse the uppercasing, but I was wondering if there was a way to completely change it (to something like "View code") via configuration?
Let me know, thanks!

```js
...
styles: {
  tabButton: {
    button: {
      textTransform: 'none'
    }
  }
...
```

aversini:
Simple update to change "Code" into "View Code" - this was a major pain point for non-dev trying to see the code of our examples. Fixes #693

### DIFF
--- a/src/rsg-components/slots/CodeTabButton.js
+++ b/src/rsg-components/slots/CodeTabButton.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import TabButton from 'rsg-components/TabButton';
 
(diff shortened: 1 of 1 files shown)


sapegin:
Thanks!

PR #724, opened 2017-12-04
Rolling back console.clear commit

BruceHubbard:
I'll go ahead and create this PR to save you some time.  All I did was remove the console.clear line in the Preview component.


sapegin:
We need to fix the bug if there is one. Just removing a useful feature isn’t a way to do that.

PR #725, opened 2017-12-05
Es upgrade md2jsx

probablyup:


loaders/utils/__tests__/__snapshots__/getSections.spec.js.snap
```diff
@@ -18,7 +18,11 @@ Array [
         "hasExamples": true,
         "metadata": Object {},
         "module": RequireStatement {
-          "filepath": "<rootDir>/test/components/Annotation/Annotation.js",
+          "filepath": 
+<rootDir>
```
sapegin: This is incorrect now. Can we explicitly enable jest-serializer-html where we need it? If not we’ll have to change `<rootDir>` to something else.

probablyup: Hmm, I don't think there's a way to customize it like that 🙁 

I could perhaps try and modify their package to be a bit more strict about what it considers HTML, since this clearly isn't HTML 😆

probablyup: cc @rayrutjes

sapegin: The check in jest-serializer-html is very naïve now but I guess there’s no easy way to do it. So probably we should just change `<rootDir>` for now.

sapegin:
Merged, thanks!

sapegin:
@probablyup I’ve noticed that it renders some paragraphs as `span`s without any classes:

image 2017-12-12 at 2 03 03 pm

probablyup:
It’ll do that unless the string is formatted like a paragraph (two new
lines at the end or two spaces)

sapegin:
It’s the only string ;-) Do you know how to fix that?

sapegin:
@probablyup Have you seen this issue with jest-serializer-html? Half of the time it’s failing on Node 4 because of different order of attributes:

```
    - Snapshot

    + Received

    

    @@ -11,12 +11,12 @@

          </em>

          <strong class="rsg--text-27 rsg--inheritSize-28 rsg--baseColor-32 rsg--strong-35">

            formatting

          </strong>

          and a

    -     <a href="/foo"

    -        class="rsg--link-18"

    +     <a class="rsg--link-18"

    +        href="/foo"

          >

            link

          </a>

          .

        </p>
```

probablyup:
Ah no, I haven’t. I think I test on node 8. I’ll take a look later

sapegin:
Node 6 and 8 are always find, only Node 4 has issues. Maybe `Object.keys` has different order for some reason depending on Travis machine it runs on?

probablyup:
Good lead, I’ll investigate. Btw 6.3.0 includes the ability to forceBlock
or forceInline, which should help with the coercion into a paragraph :)

aversini:
FYI @sapegin this update introduced a regression is md table rendering. See the [issue here](https://github.com/probablyup/markdown-to-jsx/issues/138)

loaders/utils/__tests__/__snapshots__/getSections.spec.js.snap
rayrutjes: Wow, sorry I missed this one.
Is this still an issue, and if so how can I help improve this?
cc @sapegin @probablyup

rayrutjes:
Please let me know if there is anything I can do to improve https://github.com/rayrutjes/jest-serializer-html. Open to feedback & suggestions!

Issue #702, opened 2017-11-24
Theme variable for tab button text transform style


sapegin:
Add a new variable to the [theme](https://github.com/styleguidist/react-styleguidist/blob/master/src/styles/theme.js) and use it instead `textTransform: 'uppercase'` in the [TabButton](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/TabButton/TabButtonRenderer.js) component.

jmurret:
@sapegin, is something like this what you had in mind : https://github.com/styleguidist/react-styleguidist/pull/728 ? 

I've tested it by changing it to capitalize and seeing the updated example 'VIEW CODE' change to 'View Code'.

Happy to modify this if you had something else in mind.

PR #728, opened 2017-12-06
Issue #702 - added tabButtonTextTransform variable to theme

jmurret:


.prettierrc
```diff
@@ -3,7 +3,7 @@
   "singleQuote": true,
   "trailingComma": "es5",
   "useTabs": true,
-  "proseWrap": "never",
+  "proseWrap": false,
```
sapegin: Looks like you forget to run `npm install` after pulling the changes. The right value is `never` now.

src/rsg-components/TabButton/TabButtonRenderer.js
```diff
@@ -3,14 +3,14 @@ import PropTypes from 'prop-types';
 import Styled from 'rsg-components/Styled';
 import cx from 'classnames';
 
-export const styles = ({ space, color, fontFamily, fontSize }) => ({
+export const styles = ({ space, color, fontFamily, fontSize, tabButtonTextTransform }) => ({
```
sapegin: I think `buttonTextTransform` would be enough.

PR #732, opened 2017-12-10
Add 'tab' key to Playground component's style

konradk:
According to https://react-styleguidist.js.org/docs/cookbook.html#how-to-change-styles-of-a-style-guide you are able to find component style names using React Developer tools. Unfortunately, last tab uses `classes.tab`, but it's not set in styles, so react renders with class `rsg-undefined`.

### DIFF
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -19,7 +19,7 @@ export const styles = ({ space, color, borderRadius }) => ({
 	toolbar: {
 		marginLeft: 'auto',
 	},
-	tab: {},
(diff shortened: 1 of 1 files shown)


src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -19,6 +19,7 @@ export const styles = ({ space, color, borderRadius }) => ({
 	toolbar: {
 		marginLeft: 'auto',
 	},
+	tab: {},
```
sapegin: Could you add a comment here? Otherwise someone will refactor it out ;-)

sapegin:
Thanks, merging!

PR #734, opened 2017-12-12
Fix: let the fontWeight property be inherited from body

kdzuin:
This fix sets ```fontWeight``` to ```inherit```, that allows custom font weights to be set in user components. Before it had affected components with ```fontWeight: 'normal'``` from Styleguidist UI.

### DIFF
--- a/src/styles/setupjss.js
+++ b/src/styles/setupjss.js
@@ -27,6 +27,7 @@ const jss = create({
 				// Allow inheritance because it may be set on body and should be available for user components
 				color: 'inherit',
 				fontFamily: 'inherit',
+				fontWeight: 'inherit',
 				lineHeight: 'inherit',
 				fontSize: 'inherit',
 			},


sapegin:
I guess we’ll have to manually set `normal` everywhere in the UI ;-|

image 2017-12-12 at 1 59 14 pm

Testing with this config:

```js
module.exports = {
	components: 'src/components/**/[A-Z]*.js',
	defaultExample: true,
	styles: {
		StyleGuide: {
			'@global body': {
				fontWeight: 'bold',
			},
		},
	},
	webpackConfig: {
		module: {
			rules: [
				{
					test: /\.jsx?$/,
					exclude: /node_modules/,
					loader: 'babel-loader',
				},
				{
					test: /\.css$/,
					loader: 'style-loader!css-loader',
				},
			],
		},
	},
};
```

Issue #737, opened 2017-12-18
Escape hatch to just add <link> tag to load a css file


arackaf:
I searched all over, but came up empty. 

Is there an escape hatch to tell react-styleguidist to "just add <link> tags for some css files I specify?"  I've already once in my life attempted to get all of webpack's loaders just right so bootstrap would load, but wound up quitting, and opting for a plain and simple <link> tag in my html file. 

Does such an option exist here?

arackaf:
For the benefit of anyone else finding this issue, the solution is to just use the template option, and provide your own html file with whatever script or link tags you want.

It seems styleguidis expects there to be a div with an id of app, so be sure to include that.  ie

```html
<html>
    <head>
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
    </head>
    <body>
        <div id="app"></div>
    </body>
</html>
```

sapegin:
Would be nice to add an example here: https://react-styleguidist.js.org/docs/cookbook.html ;-)

arackaf:
I'll see what I can do :)

stonebk:
Also looks like you can use `template.head.raw`:
https://react-styleguidist.js.org/docs/configuration.html#template
https://github.com/vxna/mini-html-webpack-template#readme

PR #746, opened 2017-12-20
template example added

arackaf:
This PR was motivated by this comment: https://github.com/styleguidist/react-styleguidist/issues/737#issuecomment-352661552

I couldn't easily find where to build and run the site to test this, but it should just be a straightforward addition to the docs.

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -373,7 +373,7 @@ devServer: {
 
 ## How to force client-side scripts or css links with Styleguidist?
 
-Use the `template` property in your styleguidist config, and point it to an html file of your own creation, into which you can add whatever `script` or `link` tags you need. The styleguide that's generated will use this html file, and any css files, or scripts you have loaded therein will be applied. Be sure to have a div in your body with an id of `app` as react-styleguidist will look for this.
+Use the `template` property in your styleguidist config, and point it to an html file of your own creation, into which you can add whatever `script` or `link` tags you need. The styleguide that's generated will use this html file, and any css files, or scripts you have loaded therein will be applied. Be sure to have a div in your body with an id of `app`, as react-styleguidist will look for this.
 
 For example
 
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -19,7 +19,7 @@
 * [Why does the style guide list one of my prop types as `unknown`?](#why-does-the-style-guide-list-one-of-my-prop-types-as-unknown)
 * [Why object references don’t work in example component state?](#why-object-references-dont-work-in-example-component-state)
 * [How to use Vagrant with Styleguidist?](#how-to-use-vagrant-with-styleguidist)
-* [How to force client-side scripts or css links with Styleguidist?](#how-to-use-client-side-scripts-css-with-styleguidist)
+* [How to add external JavaScript and CSS files?](#how-to-add-external-javascript-and-css-files)
 * [How to reuse project’s webpack config?](#how-to-reuse-projects-webpack-config)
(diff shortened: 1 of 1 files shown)


sapegin:
Thanks!

The site is here: https://github.com/styleguidist/site#running-locally though it would be tricky to test a PR. But you don’t really need it in this case ;-)

docs/Cookbook.md
```diff
@@ -19,6 +19,7 @@
 * [Why does the style guide list one of my prop types as `unknown`?](#why-does-the-style-guide-list-one-of-my-prop-types-as-unknown)
 * [Why object references don’t work in example component state?](#why-object-references-dont-work-in-example-component-state)
 * [How to use Vagrant with Styleguidist?](#how-to-use-vagrant-with-styleguidist)
+* [How to force client-side scripts or css links with Styleguidist?](#how-to-use-client-side-scripts-css-with-styleguidist)
```
sapegin: I’d call it “How to add external JavaScript and CSS files?” to make it close to “How to add custom JavaScript and CSS or polyfills?” and the difference between them more clear.

docs/Cookbook.md
```diff
@@ -370,6 +371,26 @@ devServer: {
 }
 ```
 
+## How to force client-side scripts or css links with Styleguidist?
+
+Use the `template` property in your styleguidist config, and point it to an html file of your own creation, into which you can add whatever `script` or `link` tags you need. The styleguide that's generated will use this html file, and any css files, or scripts you have loaded therein will be applied. Be sure to have a div in your body with an id of `app`, as react-styleguidist will look for this.
```
sapegin: styleguidist, react-styleguidist → Styleguidist
html → HTML
css → CSS
styleguide → style guide

docs/Cookbook.md
```diff
@@ -370,6 +371,26 @@ devServer: {
...
+
+For example
+
+```html
+<html>
+    <head>
+        <script src="assets/js/babelHelpers.min.js"></script>
```
sapegin: I’d use remote files as examples here because for local files `require` options is enough.

docs/Cookbook.md
```diff
@@ -370,6 +371,26 @@ devServer: {
 }
 ```
 
+## How to force client-side scripts or css links with Styleguidist?
+
+Use the `template` property in your styleguidist config, and point it to an html file of your own creation, into which you can add whatever `script` or `link` tags you need. The styleguide that's generated will use this html file, and any css files, or scripts you have loaded therein will be applied. Be sure to have a div in your body with an id of `app`, as react-styleguidist will look for this.
+
+For example
```
sapegin: Just put “:” at the end of the previous paragraph instead ;-)

docs/Cookbook.md
arackaf: You'd think so, but in this case it won't work.  `babelHelpers.min.js` needs to run on the client so that the global object which Babel's output looks for, exists.

That said, I'd be happy to make this a remote url - what would you like me to put?  I don't really use CDN's so I'm not sure what I realistic example would look like.

docs/Cookbook.md
arackaf: done

docs/Cookbook.md
arackaf: done

docs/Cookbook.md
arackaf: done - I think

docs/Cookbook.md
sapegin: Then id would be great to explain the difference with `require` ;-) For CDN I think Google Fonts would be a good example.

arackaf: Mkay - added a note explaining the difference, and also mentioned CDNs

sapegin:
Thanks!

PR #755, opened 2017-12-22
Feat: Introduce List components for Markdown.

ryanoglesby08:
Relates to Issue #613 and #619

### DIFF
--- a/src/rsg-components/Markdown/List/ListRenderer.js
+++ b/src/rsg-components/Markdown/List/ListRenderer.js
@@ -1,16 +1,23 @@
-import React from 'react';
+import React, { Children, cloneElement } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 
 import Styled from 'rsg-components/Styled';
-import Para from 'rsg-components/Para';
+import { styles as paraStyles } from 'rsg-components/Para';
 
-const styles = ({ space }) => ({
+const styles = ({ space, color, fontFamily }) => ({
 	list: {
+		...paraStyles({ space, color, fontFamily }).para,
 		paddingLeft: space[3],
 	},
 	ordered: {
 		listStyleType: 'decimal',
+	},
+	li: {
+		color: color.base,
+		fontFamily: fontFamily.base,
+		fontSize: 'inherit',
+		listStyleType: 'inherit',
 	},
 });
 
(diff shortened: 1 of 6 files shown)


src/rsg-components/Markdown/List/ListRenderer.js
```diff
@@ -0,0 +1,37 @@
...
+	},
+});
+
+export function ListRenderer({ classes, ordered, children }) {
+	const Tag = ordered ? 'ol' : 'ul';
+
+	const classNames = cx(classes.list, ordered && classes.ordered);
+
+	return (
+		<Para>
```
sapegin: I think it would be better to extend Para component to allow ul/ol and custom className.

src/rsg-components/Markdown/Markdown.js
```diff
@@ -122,19 +132,6 @@ const styles = ({ space, fontFamily, fontSize, color, borderRadius }) => ({
...
-	ul: {
-		composes: '$para',
-		paddingLeft: space[3],
-	},
-	ol: {
-		composes: '$para',
-		listStyleType: 'decimal',
-		paddingLeft: space[3],
-	},
-	li: {
```
sapegin: I believe li class should stay.

ryanoglesby08: At first, I mapped over the children in the `List` component to add a `className` so that the styles were explicit and contained in the `List` component. But then when using the Chrome dev tools I noticed that the `li` was already being reset with the JSS isolate plugin so that it would inherit properties from its parent, so the extra class on the li didn't seem to be necessary. Unless I am missing something?

src/rsg-components/Markdown/List/ListRenderer.js
ryanoglesby08: I could do that, though it seems like that creates an awkward interface. The `Para` component currently returns "paragraphs", or "blocks of text". Would it be odd to then extend it to also return lists? Seems like that will set the precedent for `Para` to start just being a component that renders every element.

The props interface also starts to become unintuitive if I code an example usage:

``` 
<Para semantic="ul" className={classes.list}>
  {children}
</Para>
```

Thoughts? Why do you think extending Para is the better way to go?

sapegin: I think that nesting lists into paragraphs is the wrong way and probably we should just duplicate styles: add margin-bottom to ul and text styles to li.

src/rsg-components/Markdown/Markdown.js
sapegin: I think we still need to define color and font for li to make it explicit.

ryanoglesby08: Done.

sapegin:
Sorry for the delay, merged!

PR #756, opened 2017-12-24
Added a ColorSwatch component and render it in the PropsRenderer when propName is 'color' or 'backgroundColor'

joshuaprimrose:
a790ef3e-98f6-4a07-9fed-54810182bd31

I had a little bit of free time. Let me know where you think improvements could be made and I will be happy to ammend PR. Thanks!


src/rsg-components/ColorSwatch/ColorSwatchRenderer.js
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+
+export const CodeRenderer = (props) => {
+	const styles = {
+		colorSwatch: {
+			backgroundColor: props.color,
+			border: props.border || 'none',
+			display: 'inline-block',
+			fontSize: 'inherit',
+			height: '1rem',
```
sapegin: Please use values from theme.js,

src/rsg-components/ColorSwatch/ColorSwatchRenderer.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const CodeRenderer = (props) => {
+	const styles = {
+		colorSwatch: {
+			backgroundColor: props.color,
+			border: props.border || 'none',
```
sapegin: Looks like you don’t use this prop.

src/rsg-components/ColorSwatch/ColorSwatchRenderer.js
```diff
@@ -0,0 +1,29 @@
...
+			display: 'inline-block',
+			fontSize: 'inherit',
+			height: '1rem',
+			margin: props.margin || 'none',
```
sapegin: Spacing between components should be defined outside, react-group should work here.

src/rsg-components/ColorSwatch/ColorSwatchRenderer.js
```diff
@@ -0,0 +1,29 @@
...
+	};
+
+	return (
+		<div style={styles.colorSwatch} />
+	);
+}
+
+CodeRenderer.propTypes = {
+	border: PropTypes.string,
+	color : PropTypes.string,
```
sapegin: This prop is required.

src/rsg-components/ColorSwatch/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './ColorSwatchRenderer.js';
```
sapegin: `rsg-components/ColorSwatchRenderer`

src/rsg-components/Props/PropsRenderer.js
```diff
@@ -115,8 +117,14 @@ function renderDefault(prop) {
 			}
 		}
 
-		return <Code>{showSpaces(unquote(prop.defaultValue.value))}</Code>;
+		return (
+			<Code>
+				{showSpaces(unquote(prop.defaultValue.value))}
+				{defaultColorKeys.indexOf(prop.name) > -1 ? <ColorSwatch color={unquote(prop.defaultValue.value)} margin="0 0 0 0.25rem" /> : null}
```
sapegin: I think it’d be better to test the value then the name with something like this https://github.com/olstenlarck/is-hexcolor

joshuaprimrose: That was my original thought as well, I was going to use https://www.npmjs.com/package/is-color but I wanted to avoid adding any additional dependencies.

sapegin: Avoiding *any* additional dependencies is a very wrong goal ;-) But is-color is bad because they have a list of color names. is-hexcolor is a one-line regexp to test a hex color.

Issue #494, opened 2017-06-12
Improve routing


sapegin:
Right now all routing is hardcoded into [index.js](https://github.com/styleguidist/react-styleguidist/blob/master/src/index.js) — it’s hard to maintain end extend. It also doesn’t support browser back / forward buttons (#467). We have to have several new features that would benefit from the proper router:

* Single-component mode (or single-component mode by default) — basically start style guide in isolated mode, do not have (or do not show by default) a page with all components.
* Split bundle by component or by section to make loading of large style guides faster (or make it possible at all).

I’m not sure what would be the best solution here. Should we use React Router or should we use something simpler?

Related issues: #467, #333.

tizmagik:
This is an interesting problem. I think it _should_ be doable to make dynamic routes in RRv4. I'll think through this a bit and post back/create a PR when I get a chance.

sapegin:
I’m not sure we need a RR here because we don’t really have real routes — we have a component filter. So we need a way to parse URL and call different filtering functions and react to URL changes. And allow extension by plugins.

n1313:
Would having "real" routing make it easier to potentially split the bundle into lazily loaded chunks, one component per chunk?

sapegin:
I’m not sure, probably not or it won’t be the hard part anyway.

tizmagik:
> So we need a way to parse URL and call different filtering functions and react to URL changes

Hmm but isn't that what a router such as RR essentially do? Would also take care of browser back button semantics, deep linking and isolation, no?

Happy to explore other approaches. What did you have in mind?

sapegin:
RR is mostly used to render different components depending on a route which we don’t need — we alway render the same component but with different props. I think we need something like [Navigo](https://github.com/krasimir/navigo).

tizmagik:
> RR is mostly used to render different components depending on a route which we don’t need

Right, but I think that might be part of the perf problem when you have a large number of components. This is the refactor I was alluding to in #212 

I'll take a look at Navigo though, it would probably be less of a refactor, I'm assuming.

zecaptus:
As I wanted to have a single component / page, I have done something that probably can help you.
Take a look here : https://github.com/zecaptus/styleguide

PR #768, opened 2018-01-09
feat(*): allow to show one component per page

stepancar:
@sapegin added simple implementation of "on component per page feature"
image

https://github.com/styleguidist/react-styleguidist/issues/494

### DIFF
--- a/src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
+++ b/src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
@@ -16,6 +16,22 @@ exports[`should ignore items without name 1`] = `
 </ul>
 `;
 
+exports[`should render links to isolated component page if oneComponentPerPage passed 1`] = `
+<ul>
+  <li
+    className=""
+    key="Button"
+  >
+    <Styled(Link)
+      className=""
+      href="blank#!/Button"
+    >
+      Button
+    </Styled(Link)>
+  </li>
+</ul>
+`;
+
 exports[`should render sections with nested components 1`] = `
 <ul>
   <li
--- a/src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
+++ b/src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
@@ -8,7 +8,7 @@ exports[`should ignore items without name 1`] = `
   >
     <Styled(Link)
       className=""
-      href="#button"
+      href="blank#button"
     >
       Button
     </Styled(Link)>
@@ -24,7 +24,7 @@ exports[`should render sections with nested components 1`] = `
   >
     <Styled(Link)
       className=""
-      href="#button"
+      href="blank#button"
     >
       Button
     </Styled(Link)>
@@ -35,7 +35,7 @@ exports[`should render sections with nested components 1`] = `
   >
     <Styled(Link)
       className=""
-      href="#input"
+      href="blank#input"
     >
       Input
     </Styled(Link)>
--- a/src/rsg-components/TableOfContents/TableOfContents.js
+++ b/src/rsg-components/TableOfContents/TableOfContents.js
@@ -7,7 +7,6 @@ import filterSectionsByName from '../../utils/filterSectionsByName';
 export default class TableOfContents extends Component {
 	static propTypes = {
 		sections: PropTypes.array.isRequired,
-		oneComponentPerPage: PropTypes.bool,
 	};
 	state = {
 		searchTerm: '',
@@ -26,8 +25,8 @@ export default class TableOfContents extends Component {
 
 	renderSections() {
 		const { searchTerm } = this.state;
-		const { sections, oneComponentPerPage } = this.props;
-
+		const { sections } = this.props;
+		const oneComponentPerPage = (this.context.config || {}).oneComponentPerPage;
 		// If there is only one section, we treat it as a root section
 		// In this case the name of the section won't be rendered and it won't get left padding
 		const firstLevel = sections.length === 1 ? sections[0].components : sections;
--- a/src/rsg-components/TableOfContents/TableOfContents.js
+++ b/src/rsg-components/TableOfContents/TableOfContents.js
@@ -7,12 +7,13 @@ import filterSectionsByName from '../../utils/filterSectionsByName';
 export default class TableOfContents extends Component {
 	static propTypes = {
 		sections: PropTypes.array.isRequired,
+		oneComponentPerPage: PropTypes.bool,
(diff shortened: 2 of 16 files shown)


src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
```diff
@@ -8,7 +8,7 @@ exports[`should ignore items without name 1`] = `
   >
     <Styled(Link)
       className=""
-      href="#button"
+      href="blank#button"
```
sapegin: This is weird.

src/utils/renderStyleguide.js
```diff
@@ -47,7 +47,9 @@ export default function renderStyleguide(
 			welcomeScreen={styleguide.welcomeScreen}
 			patterns={styleguide.patterns}
 			sections={sections}
+			allSections={allSections}
 			displayMode={displayMode}
+			oneComponentPerPage={styleguide.config.oneComponentPerPage}
```
sapegin: We already pass config here, we should access it via context in non-render components and pass as a prop to render components if they need it.

src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
stepancar: @sapegin have a look at master https://github.com/styleguidist/react-styleguidist/blob/09d2452bad84fc45839693b4311423e293c4e50a/src/rsg-components/slots/__snapshots__/IsolateButton.spec.js.snap#L14
it's because in test context location.pathname is blank. As i understand for this cases we shoul overwrite window.location for all our tests

sapegin: Ah, OK.

src/utils/renderStyleguide.js
stepancar: @sapegin thank you! fixed

stepancar:
@sapegin, added description in docs, added tests, fixes according your recomendations about config usage through context

okonet:
I'm wondering if we could skip the config part and make it behave like this based on the URL? In this case we could just add additional action (i.e. icon or ⎇-click action) to the sidebar navigation items to open in the isolated mode?

The problem with the configuration IMO is that it addresses one use case whereas there are several depending on the current environment—in development I might want to have one component per page but when publishing we should be able to see all. This can't be solved with the config right now.

Thoughts?

sapegin:
That's a very good point! I guess we need both. The option is good to completely disable all components on one page mode.

I'm thinking that enabling this option by default in dev mode would be very good. Or make it not a boolean but a sting: `always`, `never`, `static-only`, with the latter as a default value.

okonet:
> The option is good to completely disable all components on one page mode.

Why not URL based? My concern is that once option is here it's really hard to remove it in the future.

sapegin:
But the URL is always the same — `/` ;-)

okonet:
What do you mean? We can come up with the url scheme to differentiate between 2 modes

sapegin:
We can’t: you open a style guide always at root.

stepancar:
@okonet For dev mode you can use another config

sapegin:
Sounds like too much work ;-0

okonet:
Why do I need to change config in order to change the presentation of the component? This seems too much work for me as a user. But if @sapegin believes that config is the way, I’m okay with that. We can tackle that when a better router is in place, I guess.

sapegin:
I don’t see any other way: we need to know how many components to render when you open a style guide.

okonet:
Not sure I'm getting it:

> we need to know how many components to render when you open a style guide.

Why should the style guide behave differently in this case? We still can render all components as we do now but only display one if the URL matches? What am I missing?

Let's say we have the following URL: `localhost:6060/#!/Placeholder/edit`—the `edit` part should indicate that only one component will be rendered on this page.

sapegin:
> Why should the style guide behave differently in this case? We still can render all components as we do now but only display one if the URL matches? What am I missing?

Because the whole point of this feature is *not* to render all component at the same page.

stepancar:
@okonet @sapegin maybe we can release this feature? Do you have any comments for implementation?

sapegin:
@stepancar What do you think about that?

> I'm thinking that enabling this option by default in dev mode would be very good. Or make it not a boolean but a sting: always, never, static-only, with the latter as a default value.

I kind of like the idea but that may be breaking change to your current implementation.

bluetidepro:
@sapegin @stepancar Just chiming in, but yessss! Love this feature. +1 for adding it so it can be one component per page always.

stepancar:
@sapegin, I think, that developers can just use another config in dev mode or use NODE_ENV.

sapegin:
Of course they *can* but why would they do that? For me it makes a lot of sense to have such behavior by default. I don't know how other people use Styleguidist but for I never need to see all component in dev mode.

viljamis:
Would love to see this feature merged to master. Working on two different design systems that use StyleGuidist for docs and both would benefit from this feature. So basically here to show thumbs up 👍

iamtyce:
Would love to see individual routes per component rather than the long page showing all! Great work and good discussion here 👍

jeffkamo:
I would also really like to see this feature merged :) This is the functionality that my team's designers would like see in our project.

stepancar:
@sapegin, I'll try to enable this feature in dev mode by default today.

src/rsg-components/TableOfContents/TableOfContents.js
```diff
@@ -12,27 +12,27 @@ export default class TableOfContents extends Component {
...
 		});
-		return <ComponentsList items={items} />;
+		return <ComponentsList items={items} oneComponentPerPage={oneComponentPerPage} />;
 	}
 
 	renderSections() {
 		const { searchTerm } = this.state;
 		const { sections } = this.props;
-
+		const oneComponentPerPage = (this.context.config || {}).oneComponentPerPage;
```
monkeyjabs: It seems like `this.context` returns an empty object here. So `oneComponentPerPage` is always undefined

src/rsg-components/TableOfContents/__snapshots__/TableOfContents.spec.js.snap
```diff
@@ -108,6 +108,7 @@ exports[`should filter section names 1`] = `
                 },
               ]
             }
+            oneComponentPerPage={undefined}
```
okonet: Since `oneComponentPerPage` is boolean, can it be explicitly defined as `false` somewhere?

sapegin:
Was finished in another PR, thanks for help anyway!

Issue #324, opened 2017-02-14
Code that I have to write in examples is different from "real" React code


n1313:
Code examples are very useful as a part of documentation. They allow the reader to see how the "theory" of a component is applied in practice, and are the most useful when they can be easily put into a real application. With react-styleguidist, this works quite well in many cases: you write your code exactly as it should be written and your reader can just copy-paste it into their app and get it working.

Unfortunately, it doesn't work for more complex components that require some sort of external "scaffolding" to work, or for complicated setups, for example, with HOCs and theming. RSG has `initialState` and `setState` which work great but cannot be copy-pasted into a real React application: they have to be rewritten into some sort of constructor method and `this.setState` calls.

Don't get me wrong here though, I'm not saying that RSG is buggy or does its job poorly. It works great, I'm quite happy that I don't have to write all the `import React from 'react';` fluff around most of my components most of the time and can just plop a short JSX one-liner and see it "just work". The task of rendering arbitrary code is *hard*, and it is even harder if you want to have some sort of magical balance between writing short code and writing complicated code. But it would be great if the code in my examples was closer to what my users have to write in their projects.

I don't have a solution at hand and I don't even know if there _is_ a solution for this issue, given all the restrictions that RSG has. So let's talk about this, maybe? Does anybody here have this problem, and how do you deal with it? Is there anything that could be added to RSG to make it even better in this regard?

sapegin:
I don’t know a good way to solve this. So far I use comments and define component classes inside examples:

```jsx
/*
import TableProvider from 'kao/components/TableProvider';
import * as sort from 'kao/util/sort';
import { Header, Body } from 'reactabular-table';
import orderBy from 'lodash/orderBy';
*/
class SortTable extends React.Component {
  constructor() {
    super();
    const sortable = sort.sort({
      getSortingColumns: () => this.state.sortingColumns,
      onSort: selectedColumn => {
        this.setState({
          sortingColumns: sort.byColumn({
            sortingColumns: this.state.sortingColumns,
            selectedColumn
          })
        });
      }
    });
    this.state = {
      sortingColumns: {},
      columns: [
        {
          property: 'name',
          header: {
            label: 'Name',
            formatters: [sort.header(sortable)],
          },
        },
        {
          property: 'color',
          header: {
            label: 'Color',
            formatters: [sort.header(sortable)],
          },
          cell: {
            formatters: [fixtures.getColorName],
          },
        },
        {
          property: 'male',
          header: {
[...]

PR #769, opened 2018-01-10
Update Documenting.md

n1313:
This PR updates documentation to fix some typos and (hopefully) better demonstrate an alternative way of writing code examples (the "real" React way), as discussed in #324.

### DIFF
--- a/docs/Documenting.md
+++ b/docs/Documenting.md
@@ -205,7 +205,7 @@ class Button extends React.Component {
 
 ## Writing code examples
 
-Code examples in Markdown use the ES6+JSX syntax. They can access all the components of your style guide using global variables:
+Code examples in Markdown use ES6+JSX syntax. All components covered by the style guide can be used in all examples:
 
 ```jsx
 <Panel>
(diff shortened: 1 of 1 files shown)


docs/Documenting.md
```diff
@@ -240,33 +240,37 @@ initialState = { isOpen: false }
 </div>
 ```
 
-You _can_ create `React.Component`s in your code examples:
+Using `initialState`, `state` and `setState` helpers to write examples is easy and convenient, but you can write "real" React code as well. For example, the above example could be written like this:
```
sapegin: I think this paragraph was better. Any JSX code is real React code and the point here that you can create new components inside examples.

docs/Documenting.md
```diff
@@ -240,33 +240,37 @@ initialState = { isOpen: false }
...
+class ModalExample extends React.Component {
   constructor() {
     super()
     this.state = {
-      /* ... */
+      isOpen: false
     }
   }
+  toggleOpen() {
+    this.setState({
```
sapegin: Let's not teach people bad practices: use a functional form of setState ;-)

docs/Documenting.md
```diff
@@ -240,33 +240,37 @@ initialState = { isOpen: false }
...
-    const sortedRows = require('sortabular').sorter({
-      /* ... */
-    })(rows)
     return (
-      <TableProvider columns={columns}>
-        <Table.Header />
-        <Table.Body rows={sortedRows} rowKey="id" />
-      </TableProvider>
+      <div>
+        <button onClick={this.toggleOpen.bind(this)}>Open</button>
```
sapegin: If we bind in the constructor  (or use arrow function if Buble supports it) we could make these lines shorter and easier to read.

docs/Documenting.md
n1313: By saying "real React code" I mean "code that can be copy-pasted without modifications into any other React application". How should I rephrase it?

docs/Documenting.md
n1313: Sure, I'll move the binding to the constructor.

docs/Documenting.md
n1313: I want to keep this example as simple as possible. Functions in setState are not widely used yet, as far as I understand.

sapegin: There are or they should ;-) There are recommended by React docs: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

n1313:
Alright, I’ll change that.

docs/Documenting.md
sapegin: Maybe something like this:

`initialState`, `state` and `setState` helpers are good to show components in different states, but to allow users copy-paste code without modifications into their React app you may want to use `React.Component` instead. We can rewrite an example above like this:

n1313:
@sapegin updated the PR.

sapegin:
Thanks, merged!

Issue #772, opened 2018-01-10
Floated elements are not contained


roblevintennis:
We have some floated buttons. I noted there's no [block formatting context](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Block_formatting_context) in the styleguideist wrapper divs e.g. `display: inline-block; width: 100%;` (probably preferable to overflow hidden which would also work). Essentially, this means no floated element can be displayed and contained successfully within the main component box. Here's an example from the button on demo .. note I've added a float to the button and it's no longer contained:

![image](https://user-images.githubusercontent.com/142403/34800881-a584ddd0-f61a-11e7-8abc-bd1437fe8b4d.png)

sapegin:
Feel free to send a pull request to fix that.

roblevintennis:
@sapegin Sure thanks .. would you recommend I figure out how to just add it to the "base theme" and/or could you point me to which files best to touch? Guess I'll try that unless you have another recommendation.

sapegin:
I’d start here:

https://github.com/styleguidist/react-styleguidist/blob/ab903f751f9e337108d24dbda4410a98b80804a9/src/rsg-components/Playground/PlaygroundRenderer.js#L10-L14

roblevintennis:
Thank you very much for the 411 @sapegin .. I think my team's gonna give me some time today to do that ... I'll at mention on the PR!

sapegin:
Out in [6.2.0](https://github.com/styleguidist/react-styleguidist/releases/tag/v6.2.0)!

roblevintennis:
oh cool! thanks for the heads up .. I can change my package.json pointer!

PR #773, opened 2018-01-11
Adds a block formatting context to the preview div so that any floate…

roblevintennis:
Addresses issue: https://github.com/styleguidist/react-styleguidist/issues/772

@sapegin I looked for a place to add a spec e.g. src/rsg-components/Playground/Playground.spec.js but I didn't see anything else asserting on CSS things in there so I deemed this not a thing that needed specs. Please advise if otherwise. That said, Here are some screenshots to make this PR crystal clear…

Nothing broke:

image
---

Proof .. floated button gets contained (note I added the `float: right` manually in devtools):

image

---

Here's the preview div with block formatting context rules applied:

image

---

Lmk if there's anything else you'd like done here. Thanks

### DIFF
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -11,8 +11,9 @@ export const styles = ({ space, color, borderRadius }) => ({
 		padding: space[2],
 		border: [[1, color.border, 'solid']],
 		borderRadius,
-		width: '100%', // expands following inline-block
-		display: 'inline-block', // block formatting context required to contain floats
+		// the next 2 lines are required to contain floated components
+		width: '100%',
+		display: 'inline-block',
 	},
 	controls: {
 		display: 'flex',
--- a/src/rsg-components/Playground/PlaygroundRenderer.js
+++ b/src/rsg-components/Playground/PlaygroundRenderer.js
@@ -11,8 +11,8 @@ export const styles = ({ space, color, borderRadius }) => ({
 		padding: space[2],
 		border: [[1, color.border, 'solid']],
 		borderRadius,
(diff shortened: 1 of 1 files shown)


src/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -11,6 +11,8 @@ export const styles = ({ space, color, borderRadius }) => ({
 		padding: space[2],
 		border: [[1, color.border, 'solid']],
 		borderRadius,
+		width: '100%',
```
sapegin: I'd add a comment that it's a clearfix and why we need it here, otherwise these two lines will be eventually removed during refactoring ;-)

sapegin:
> I didn't see anything else asserting on CSS things

For that we’ll need something like Cypress which would be awesome ;-)

src/rsg-components/Playground/PlaygroundRenderer.js
roblevintennis: Ah ok, cool will do

roblevintennis:
@sapegin I added comments in a34b15b ... not sure what your line length rules are but I couldn't think of any shorter way to express .. lmk if it's making lines too long 😋

sapegin:
I think it requires too much knowledge to understand and more about how then why. I’d add more generic comment on top of these two lines, explaining what’s the problem this code is solving.

roblevintennis:
well taken ... I tried to shorten comment and speak more to the issue being solved and not the mechanics of fix per your suggestion. Hopefully what you're looking for but lmk if you have something better :)

sapegin:
Thanks, merged!

roblevintennis:
👍 thank you!

PR #781, opened 2018-01-20
Better prop-types definitions with flow types

nanot1m:
Some plans:
- [x] handle rendering of functions
- [ ] handle rendering of arrays
- [x] handle rendering of shapes
- [x] handle rendering of union-types


nanot1m:
Current result

nanot1m:
Should complex types (like shapes, or function) be rendered in the _type_ column or in the _description_ column?

sapegin:
Good start, thanks!

I think we should render them in tooltips like default values on your screenshot, they may be very large.

nanot1m:
Found a problem with react-docgen: it doesn't handle array types marked with `[]`.

For example, type `foo: string[]` would be parsed as `unknown` type

nanot1m:
Implementing rendering complex types in tooltip


Do you have any thoughts about how to improve rendering?

nanot1m:
I'm going to remove FlowButton component from basic example

sapegin:
> Do you have any thoughts about how to improve rendering?

I think it's good enough.

src/rsg-components/Props/util.js
```diff
@@ -18,6 +18,16 @@ export function getType(prop) {
...
 }
 
+/**
+ * Checks if prop has flow type definitions.
+ *
+ * @param {object} prop
+ * @returns {boolean}
+ */
+export function isFlowTyped(prop) {
+	return Boolean(prop.flowType);
```
sapegin: Can't we check `prop.flowType` directly? (You're doing it anyway ;–) This function feels unnecessary.

src/rsg-components/Props/PropsRenderer.js
```diff
@@ -34,6 +34,34 @@ function renderType(type) {
...
+			return renderComplexType(name, raw);
+		default:
+			return raw || name;
+	}
+}
+
+function renderComplexType(name, title) {
+	return (
+		<Text size="small" underlined title={title}>
+			<Type>{name}</Type>
```
sapegin: This whole thing is already wrapped in `<Type>`, is it necessary to have one more here?

src/rsg-components/Props/util.js
nanot1m: It's just abstraction, and it can be safely thrown away

src/rsg-components/Props/PropsRenderer.js
nanot1m: Text component does not inherit color from parents. Should we add a prop `color="inherit"` to Text?

nanot1m:
I've forgotten about public components methods. Have to update them too

src/rsg-components/Props/util.js
sapegin: There's no such thing as "just an abstraction" ;-) each has its own cost. I don't think this one worth its cost, and since you're not using it consistently, which makes the cost even higher.

src/rsg-components/Props/PropsRenderer.js
sapegin: OK, leave it like this for now ;-)

nanot1m:
Maybe public methods should be updated in another pull request? If so, i'm done with this task. And will continue work in another brunch

sapegin:
Good idea!

sapegin:
Could you take a look at the CI log here?

nanot1m:
```
Error: Navigation Timeout Exceeded: 30000ms exceeded
    at Promise.then (/home/travis/build/styleguidist/react-styleguidist/node_modules/puppeteer/lib/NavigatorWatcher.js:71:21)
```
Something went wrong with puppeteer.
Can you restart CI task?

nanot1m:
```
codecov/patch — 95.83% of diff hit (target 96.33%)
```
Can you explain, what does this mean? And how can i improve code coverage?

sapegin:
> codecov/patch — 95.83% of diff hit (target 96.33%)

It means that the test coverage for the lines you changed in this PR is 95.83% and recommends to add more test to match the goal of 96.33% (I'm not sure where it's coming from). You can click Details and see which lines aren't covered by tests.

nanot1m:
Thanks. I've looked at uncovered places. There are two uncovered lines, which are not referenced to my changes.

one is 
```js
export function getRowKey(row) {
  return row.name;
}
```
and another is
```js
function renderEnum(prop) {
  if (!Array.isArray(getType(prop).value)) {
    return <span>{getType(prop).value}</span>; // not hitting this line
  }
...
```

sapegin:
That would be supercool but feel free to ignore them ;-)

nanot1m:
I've added test, nothing changed (only codecov/project improved). So let's skip this warning 👍

sapegin:
Thanks, I'll merge it a bit later -- don't want to block master because we have a few unfixed bugs.

SavePointSam:
Is there an expectation that this should render as a union?

```
const THEMES: {
  PRIMARY: 'primary',
  SECONDARY: 'secondary',
} = {
  PRIMARY: 'primary',
  SECONDARY: 'secondary',
};

type Props = {
  /** display theme */
  theme?: $Values<typeof THEMES>,
};
```

Flow just added support for this type of syntax with [0.68.0](https://github.com/facebook/flow/commit/ab0789ade95090c2a07ce1dff0e6511226ed73fd#diff-ad5fd07f79149cda5d68d3b9c9e166e5R48). I'm currently getting the following render in my styleguide.

image

---

On another note, is there work planned/being done to make the experience of hover details for something like a tuple better? Having the alt text show more information is ok, but I was confused at first as there wasn't something like tooltip.js appearing. It would also be nice to expand sub-types from that message too. For example I have a type that is an array of types and it just outputs `[Action]` which isn't clear to an end user. Maybe thats good enough though as a consumer of the API will be able to discover what the `Action` would be?

Thanks for this update! My styleguide is sooo much better documented out of the box. 🎉

PR #784, opened 2018-01-22
Sort properties by 'required' and 'name' attributes

dotcs:
I've noticed that properties are not sorted. I think it would make sense to first show the required properties and then the optional ones. Each of both groups should be grouped by name.

This PR solves the issue by adjusting how `propsToArray` works.


dotcs:
@sapegin I'm not completely sure, but I guess it's not possible since `props` is an object all the time until it's converted to an array. IMHO this is the first (and only?) position where we can sort values.

sapegin:
I don't think there's a strict requirement to keep it as an object, but I'm afraid it may be too difficult to change now.

src/rsg-components/Props/Props.spec.js
```diff
@@ -1,18 +1,20 @@
...
+import PropsRenderer, { columns } from './PropsRenderer';
 import { unquote, getType, showSpaces } from './util';
 
 // Test renderers with clean readable snapshot diffs
 // eslint-disable-next-line react/prop-types
 export default function ColumnsRenderer({ props }) {
 	return (
 		<ul>
-			{propsToArray(props).map((row, rowIdx) => (
+			{/* eslint-disable react/prop-types */}
```
dotcs: Not sure why `react/prop-types` makes problems here. I've tried to add `propTypes` for this component but had no success. Do you have an idea?

```javascript
ColumnsRenderer.propTypes = {
  props: PropTypes.array.isRequired,
}
```

sapegin: And it was already disabled: `// eslint-disable-next-line react/prop-types` ;-/

dotcs:
@sapegin Changing how the props-loader works was a bit harder, but it's possible I guess. Please have another look. Since I don't know the code by heart I hope I have found every occurrence of props object vs. array, so please consider having another look into other components that are involved which I could have forgotten.

src/rsg-components/Props/Props.spec.js
dotcs: Oh, I didn't notice that. 👍

dotcs:
@sapegin What do you think? Are there still changes necessary or are you fine with it? Current review status says that there are still requested changes.

loaders/props-loader.js
```diff
@@ -12,6 +12,18 @@ const consts = require('../scripts/consts');
 
 const ERROR_MISSING_DEFINITION = 'No suitable component definition found.';
 
+function sortProps(props) {
```
sapegin: This function should be in a separate file with JSDocs comment, like other functions. Will be easier to test separately too.

loaders/props-loader.js
```diff
@@ -52,6 +64,10 @@ module.exports = function(source) {
 	}
 
 	props = getProps(props, file);
+	if (props.props) {
+		const sortedProps = sortProps(props.props);
+		props.props = sortedProps;
```
sapegin: I think `props.props = sortProps(props.props);` would be enough here.

src/rsg-components/Props/Props.spec.js
sapegin: Let's disable it for the whole file, it's a test anyway ;-)

loaders/props-loader.js
dotcs: You're right. I've separated the function into an own file, added JSDocs and tests.

loaders/props-loader.js
dotcs: That's true. Done. 👍

dotcs:
@sapegin Alright, I've fixed all findings 👍

n1313:
Can this be made optional? Props have different significance, and I personally sort/group them manually in the order of significance. For example, I would dislike having something like `onTransactionSigned` being placed above `transactions` because you wouldn't understand what `onTransactionSigned` does without reading about `transactions` first.

dotcs:
@n1313 I guess it could be made optional. In the props-loader we have access to the `config` object, which could be used to add additional config to it.
Here we could enable or disable the sorting, e.g. by having a property called `sortComponentProperties: boolean`. The default value could be `false` to keep the default as it is right now. 
@sapegin What do you think?

sapegin:
I'd suggest to make a more generic option to transform props with the default implementation that will sort them like in this PR. To disable sorting you could pass a function that does nothing. I don't really like the idea or an option that control sorting of props, that feels too specific.

dotcs:
I've added my idea of how a generic function could look like. What are your opinions on that approach?

n1313:
The option of sorting props is valuable and useful, but I would personally prefer the default prop display behaviour to not change (especially if it is coming in with only a minor version bump).

sapegin:
I'd like to collect more opinions, so far we're two against one ;-)

@okonet @kof @SaraVieira @stepancar What do you think?

dotcs:
@n1313 I understand your point, but I do not agree. The current implementation does not make any assumptions about the order of the properties. Eventually it uses the sorting of keys in an object, which is not standardised (see [here](https://stackoverflow.com/a/5525820)). So this should change anyways to make it reliable - even if this had not been an issue so far.

kof:
this sort by default makes sense, evtl optionally could be disabled, but I don't care right now about this, I hate manually decide what is more important and most of the time just put  require stuff first. Btw this should be an eslint feature which also works with --fix

n1313:
Well, truth be told, I'm still sitting on v5.3.2 (because I couldn't get over the previous UI change that came in with a minor version bump and undoing that change is too much effort), so this change won't affect me at all. But if I was on v6, I would be pissed.

kof:
One suggestion to validate users reaction could be to release a beta first and see if any one else has  hard feelings about this change. In general prettier showed that good automated defaults are well accepted by the community.

n1313:
Hey, nice, I dislike prettier too! :)

kof:
Btw. having this in eslint would completely solve the problem as you would shift this decision to the user.

n1313:
https://eslint.org/docs/rules/sort-keys exists, but it doesn't work with --fix :|

dotcs:
eslint does not solve this problem, because eslint is a tool for structuring code. But it does not change the way how the underlying JavaScript engine interprets objects and how it sorts the keys. `react-docgen` (or `react-docgen-typescript`) returns the property as an object in which the property names are the keys. This is what we get and what we have to work with! 
As mentioned sorting of keys in an dictionary is [not standardised](https://stackoverflow.com/a/5525820). This means that how the properties are sorted right now is not guaranteed in any way. eslint does not change anything - regardless of how the user sort the properties and how they get parsed by `react-docgen`.

The only way to solve this problem is to use a data-structure with a guaranteed order, such as arrays. This PR tries to archive an order of the properties by transforming the object to an array of objects where a strict order is defined and uses this array for property rendering.

So please keep this discussion focused. eslint does something completely different. The `sort-keys` rule is meant for readability not for data integrity. It does not help in this discussion in any way.

sapegin:
> Well, truth be told, I'm still sitting on v5.3.2 (because I couldn't get over the previous UI change that came in with a minor version bump and undoing that change is too much effort), so this change won't affect me at all. But if I was on v6, I would be pissed.

The let’s just do this and see if anyone else will complain. I believe this would be more useful for majority of the users and we have a way to disable this. We can’t make all users happy anyway.

docs/Configuration.md
```diff
@@ -426,6 +427,12 @@ Type: `String`, default: `<app name from package.json> Style Guide`
 
 Style guide title.
 
+#### `transformProps`
+
+Type: `Function`, optional
+
+Function that transforms component properties.
```
sapegin: An example how to disable sorting and note that the default implementation sorts props would be nice.

dotcs:
> [...] we have a way to disable this

That's true. We could disable it by using `propsTransform: props => props`. This passes the array without any transformation (also no sorting) which would replicate the way it has been done before.

To completely solve the problem and not rely on the pseudo-sorting of keys in an object we would have to change the way how `react-docgen` and `react-docgen-typescript` save properties. It should be an array already at this point such that the order is guaranteed, but this is not the scope of this PR.

docs/Configuration.md
dotcs: Good idea. I've improved the documentation. Please have another look.

n1313:
> let’s just do this and see if anyone else will complain. 

What's the plan if someone else does complain? Roll back the release?

> We can’t make all users happy anyway.

In my opinion it is important to note that current situation with props being unsorted is *not* a known source of unhappiness for any noticeable fraction of users of RSG (at the very least I am not aware of any issues or complaints about this in the repo). Many users are relying on current behaviour of RSG even though it is non-standard, and changing it will no doubt be a surprise to them. 

Please consider keeping current behaviour as default by shipping with `propsTransform: props => props` and instead adding the sorting function to documentation as a useful example of using `propsTransform`.

sapegin:
> What's the plan if someone else does complain? Roll back the release?

We can always change the default value.

> Please consider keeping current behaviour as default by shipping with propsTransform: props => props and instead adding the sorting function to documentation as a useful example of using propsTransform.

I *did* consider that and have made a decision, thanks for comments.

docs/Configuration.md
```diff
@@ -426,6 +427,22 @@ Type: `String`, default: `<app name from package.json> Style Guide`
...
+
+Type: `Function`, optional
+
+Function that transforms component properties. By default properties are sorted such that required properties come first, optional prameters come second. Properties in both groups are sorted by their property names.
+
+To disable sorting the identity function can be used:
+
+```javascript
+module.exports = {
+  transformProps: function(props) {
```
sapegin: `props => props` ;-)

okonet:
I think it’s a great thing to put required props on top and option below. I can imagine some push back from very pedantic users who carefully organize props in the source code so an opt-out should be provided but I believe most of users will appreciate this.

Issue #774, opened 2018-01-12
Providing an array of component locations in config


selbekk:
Hi there! Great project! Thanks for putting in the time to make this awesome product, and then just give it away for free.

I have a question about the styleguide.config.js configuration, before I potentially provide a PR:

When you specify components for a given section, and you want several locations, you have to provide a function that returns an array instead of a simple array. Why can't you simply specify an array? 

Here's an example config to show what I mean:

```js
module.exports = {
    title: 'My style guide',
    sections: [
        {
            name: 'Buttons',
            components: () => [ // this! why not just a simple array?
              'packages/button/src/Button.js', 
              'packages/action-button/src/ActionButton.js',
          ],
        },
    },
};
```

I'd love to provide a PR to add this functionality if there isn't any particular reason why the API is this way.

sapegin:
Simple reason: functions are more powerful and arrays are more like a shortcut. So that means I'll be happy to merge a PR that adds arrays too ;-)

PR #794, opened 2018-01-26
feat: allow use of array of component paths in the config file

glebez:
This PR addresses the following issue: #774 

Now along with a string or function it is possible to pass an array of file paths to `components` property of the config.

### DIFF
--- a/loaders/utils/getComponentFiles.js
+++ b/loaders/utils/getComponentFiles.js
@@ -4,6 +4,7 @@ const path = require('path');
 const glob = require('glob');
 const isFunction = require('lodash/isFunction');
 const isString = require('lodash/isString');
+const isArray = require('lodash/isArray');
 
 /**
  * Return absolute paths of components that should be rendered in the style guide.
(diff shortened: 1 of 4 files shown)


loaders/utils/getComponentFiles.js
```diff
@@ -21,11 +22,13 @@ module.exports = function getComponentFiles(components, rootDir, ignore) {
 	let componentFiles;
 	if (isFunction(components)) {
 		componentFiles = components();
+	} else if (isArray(components)) {
```
sapegin: You can use `Array.isArray` here, probably less consistent in this particular case, but I’d prefer a native method if it’s available and works well ;-)

glebez: Yep, was thinking about it, but than noticed that lodash is used in both other checks and decided to go with consistency.
Anyways, will update in a moment.

glebez: Aaaand it's there.

Issue #776, opened 2018-01-13
End-to-end tests with Cypress


sapegin:
Right now we only have unit test and [very simple test](https://github.com/styleguidist/react-styleguidist/blob/master/test/browser.js) that check bundles for JS errors in headless Chrome. But we don’t know that Styleguidist is actually usable.

[Cypress](https://www.cypress.io/) seems like a good choice.

- [ ] Setup Cypress
- [ ] Write first test
- [ ] Write more tests

glebez:
It feels that we can check off first two boxes. And think on breaking the third one into more actionable, i.e. 'write cypress tests for sidebar', since some core test are already in place.

sapegin:
I’ve added a few — feel free to suggest more 🦄

haquezameer:
@sapegin @glebez I would like to take this up if it is still open.

sapegin:
@haquezameer Awesome! I'm sure most of the todo items weren't done yet.

haquezameer:
ok @sapegin. I am taking this up.

dhuang612:
is this available to work on?

Thanks

sapegin:
@dhuang612 sure, there's never too many tests.

milafreckled:
Hi, I was inspired by this issue, and already made a pull request with some tests, you can check and correct me if necessary. I think it must cover all unchecked boxes above.

gulammoyuddin:
is it still available to work on?

SivalakshmiViswanathan:
I would like to contribute. Is it still available?

Snig162000:
Hi, is this issue still open? I would like to contribute.

PR #795, opened 2018-01-29
feat: setup cypress

glebez:
Hi, 

this PR is an attempt of first step for issue #776 

It includes cypress setup, new npm scripts and one simple test. I've left most of the cypress scaffolded files, since it can be a good reference for other contributors.

Currently, cypress tests can not be run fully automatically, so first you need to start a test server (`test:cypress:runTestServer`) and than either run the cypress GUI (`test:cypress:open`) or run the tests from command line (`test:cypress:run`).

I have reused the server from `run.server.js`, but changed the port to 8082 so the tests can be run in parallel with default port styleguidist instance.

Would love to get some feedback and maybe a little better understanding where do you see cypress tests fit for this project. Do you want them to run together with other tests on commit or run them in CI?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,14 +3,12 @@
   "extends": "tamia/react",
   "env": {
     "browser": true,
-    "node": true,
-    "cypress/globals": true
+    "node": true
   },
   "plugins": [
     "compat",
     "es5",
-    "import",
-    "cypress"
+    "import"
   ],
   "settings": {
     "import/resolver": {
@@ -44,6 +42,7 @@
     "classes": false,
     "shallow": false,
     "render": false,
-    "mount": false
+    "mount": false,
+    "cy": false
   }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,12 +3,14 @@
   "extends": "tamia/react",
   "env": {
     "browser": true,
-    "node": true
+    "node": true,
+    "cypress/globals": true
   },
   "plugins": [
     "compat",
     "es5",
-    "import"
+    "import",
+    "cypress"
   ],
   "settings": {
     "import/resolver": {
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
     "cheerio": "^1.0.0-rc.2",
+    "cypress": "^1.4.1",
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -102,6 +102,7 @@
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
     "cheerio": "^1.0.0-rc.2",
+    "cypress": "^1.4.1",
```
sapegin: Is it big / slow to install like Puppeteer?

.eslintrc
```diff
@@ -3,12 +3,14 @@
   "extends": "tamia/react",
   "env": {
     "browser": true,
-    "node": true
+    "node": true,
+    "cypress/globals": true
```
sapegin: I'm not sure it's really useful, I'd just define `cy` in globals.

glebez: I've just seen the recommendation to add this in the cypress docs, so don't have a strong opinion here. It also adds browser and mocha globals, but those might be in your setup already.

package.json
glebez: Hm, not sure, will check. You want to move it the installation to pre-test hook, like you're doing with puppeteer?

.eslintrc
sapegin: I don't have a strong opinion also but it feels unnecessary to have an extra dependency to basically define one or tow global variables in ESLint config.

package.json
sapegin: Only if they do something crazy: Prettier downloads Chrome, which is ~70 MB, on install.

glebez:
So, the requested changes are there and travis runs cypress successfully 🎉 
And yes, I removed cypress from dependencies, since it was installing 89 other packages and for me it took about a minute on overall installation. So now it is installed through a script.

I'm wondering if the cypress tests should run only for node 8, like current browser test, or it's ok that they run for all versions, like now?

sapegin:
Awesome!

> I'm wondering if the cypress tests should run only for node 8, like current browser test, or it's ok that they run for all versions, like now?

I think we should only run it once per CI run, I don't see how it could be useful to run it in multiple Node versions -- unit tests should be enough to test Node versions compatibility.

glebez:
All right, now it runs only once per CI run.

sapegin:
Awesome, merged!

PR #803, opened 2018-02-02
Include filepath in props to Section/ReactComponentRenderer

### LINKED ISSUE #800: Include path in section props
I would like to add "Edit this doc" links to each section. It's easy enough to do so with a custom `ReactComponentRenderer` (eg, getting the path via `pathLine`) but less so with markdown sections (via `SectionRenderer`).

Any objections to including filepath as a prop [in processSection](https://github.com/styleguidist/react-styleguidist/blob/master/loaders/utils/getSections.js#L58)?

wachunga:
Include filepath to applicable renderers. This was already the case with `ReactComponentRenderer`, but only via pathLine which may have been mutated by config. This allows renderers to do things like include an "Edit this document" link.

For the record, I don't think we should be overwriting `section.config` which starts as a path and then becomes the actual section content, but changing this would be a breaking change. So I stuck with a new property (`filepath`).

Closes #800


loaders/utils/getSections.js
```diff
@@ -23,40 +23,41 @@ function getSections(sections, config) {
 	return sections.map(section => processSection(section, config));
 }
 
+function getComponentsPerConfig(section, config) {
```
sapegin: This function should be in a separate file, with JSDocs comment, and it should have tests. Looks at other files in `utils` folder.

Also the name is confusing: something like `getSectionComponents` would be a bit better.

loaders/utils/getSections.js
```diff
@@ -23,40 +23,41 @@ function getSections(sections, config) {
...
 			throw new Error(`Styleguidist: Section content file not found: ${filepath}`);
 		}
-		content = requireIt(`!!${examplesLoader}!${filepath}`);
-	}
-
-	let ignore = config.ignore ? _.castArray(config.ignore) : [];
-
-	if (section.ignore) {
-		ignore = ignore.concat(_.castArray(section.ignore));
+		processed.content = requireIt(`!!${examplesLoader}!${filepath}`);
```
sapegin: I think it was easier to follow without mutation. Also we can make it clearer by assigning `section.content` to a variable like `contentRelativePath`, because we  have two `content`s with different meaning.

loaders/utils/getSections.js
wachunga: This helper function doesn't seem like a useful thing to expose, similar to those in [getProps](https://github.com/styleguidist/react-styleguidist/blob/master/loaders/utils/getProps.js). Its logic is still covered by existing `processSection` tests...

loaders/utils/getSections.js
wachunga: I prefer mutation (addition, really) in this case because it's clearer that `content` is not assigned a value unless `section.content` is present. Previously, you had to check the rest of the function. But I don't feel strongly about it.

I don't like the multiple meanings of `content` either. I'm a little unclear what you're proposing. Do you just mean a local reassignment like so?
```js
const contentRelativePath = section.content;
if (contentRelativePath) {
...
}
```

loaders/utils/getSections.js
sapegin: Then let’s make it an arrow function as in getProps ;-)

loaders/utils/getSections.js
sapegin: Yup!

wachunga:
Thanks for your feedback. Updated my commit.

sapegin:
Merged, thanks!

PR #805, opened 2018-02-04
Introduce markdown Blockquote component. #613

ryanoglesby08:
So that it is easier for consumers to customize their styleguide by replacing
components.

### DIFF
--- a/src/rsg-components/Markdown/Blockquote/BlockquoteRenderer.js
+++ b/src/rsg-components/Markdown/Blockquote/BlockquoteRenderer.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Styled from 'rsg-components/Styled';
+import { styles as paraStyles } from 'rsg-components/Para';
+
+const styles = ({ space, color, fontSize, fontFamily }) => ({
+	blockquote: {
+		...paraStyles({ space, color, fontFamily }).para,
+		fontSize: fontSize.base,
+		margin: [[space[2], space[4]]],
+		padding: 0,
+	},
+});
+
+export function BlockquoteRenderer({ classes, children }) {
+	return <blockquote className={classes.blockquote}>{children}</blockquote>;
+}
+BlockquoteRenderer.propTypes = {
+	classes: PropTypes.object.isRequired,
+	children: PropTypes.node.isRequired,
+};
+
+export default Styled(styles)(BlockquoteRenderer);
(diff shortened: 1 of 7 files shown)


src/rsg-components/Markdown/Blockquote/BlockquoteRenderer.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Styled from 'rsg-components/Styled';
+import { styles as paraStyles } from 'rsg-components/Para';
+
+const styles = ({ space, color, fontSize, fontFamily }) => ({
+	blockquote: {
+		...paraStyles({ space, color, fontFamily }).para,
```
sapegin: I think this is confusing: you're overwriting half of the styles, I’d just copy the relevant styles here.

src/rsg-components/Markdown/Blockquote/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './BlockquoteRenderer';
```
sapegin: Should be `rsg-components/Markdown/Blockquote`.

src/rsg-components/Markdown/Blockquote/BlockquoteRenderer.js
ryanoglesby08: Sure, can do. I was just copying the existing styles from the old style-based way in Markdown.js. Can switch to not using Para at all here.

src/rsg-components/Markdown/Blockquote/index.js
sapegin: If you see this pattern somewhere in Styleguidist, then it's a bug ;-) That prevents component overriding to work properly, so we should always use the full form.

src/rsg-components/Markdown/Blockquote/BlockquoteRenderer.js
sapegin: It was a bit better when it was defined in the same file ;-)

src/rsg-components/Markdown/Blockquote/index.js
ryanoglesby08: Ahhhhhh I see now. I knew I wasn't crazy! :) I missed this commit that changed a lot of the exports from a relative path to the rsg-components version: 

https://github.com/styleguidist/react-styleguidist/commit/46156bb932121df5a176852d0fcdfd43cad83ef3

Following suit now

sapegin:
Thanks, merged!

PR #809, opened 2018-02-08
Add example using require.context (fixes #808)

### LINKED ISSUE #808: Dynamically requiring components / require.context
Is there a way to dynamically require a component in an example?

My use case is building a gallery of all of our icons, which are all SVG files transformed into React components at load time using `svg-loader`.   We have a lot of icons, so I'd like to avoid having to manually require them for the gallery - it's also an extra build step that's easily forgotten when adding icons to our styleguide.

This works in examples, but requires manual work for each new icon:

```jsx
const icons = {
    iconA: require('./icons/IconA.svg'),
    iconB: require('./icons/IconB.svg'),
    //...,
    iconZ: require('./icons/IconZ.svg')
}

<IconGallery icons={icons} />
```

Downside is manually adding the icon + require to each example.  The following would be a small improvement, but doesn't seem possible unless I'm doing it wrong - I get an error message that `require with expressions is not supported`:

```jsx
const iconNames = ['iconA', 'iconB', /* ... */ 'iconZ'];
const icons = iconNames.reduce((icons, name) => {
    icons[name] = require('./icons/' + name + '.svg');
    return icons;
}, {});

<IconGallery icons={icons} />
```

If I were building this in straight webpack, I would do something like this -- and this would be my preference since icons can just be added by dropping them in the right folder without editing the styleguide.  But I get the error that `require.context is not a function`:

```jsx
const iconsContext = require.context('./icons/', true, /svg$/);
const icons = iconsContext.keys().reduce( (icons, file) => {
	const SVG = iconContext(file).default;
	const label = file.slice(2, -4); // strip './' and '.svg'
	icons[label] = SVG;
        return icons;
}, {});

<IconGallery icons={icons} />
```

Given the error messages I'm getting, it seems this is not possible... but I thought I'd ask in case there's a feature or configuration I'm missing.  Alternatively, any pointers for where I could look if I wanted to try to try my hand at adding this?

boxfoot:
Added a cookbook example to demonstrate dynamically loading components using require.context.

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -7,6 +7,7 @@
 * [How to use `ref`s in examples?](#how-to-use-refs-in-examples)
 * [How to exclude some components from style guide?](#how-to-exclude-some-components-from-style-guide)
 * [How to hide some components in style guide but make them available in examples?](#how-to-hide-some-components-in-style-guide-but-make-them-available-in-examples)
+* [How to dynamically load other components in an example?](#how-to-dynamically-load-other-components-in-an-example)
 * [How to set global styles for user components?](#how-to-set-global-styles-for-user-components)
 * [How to add custom JavaScript and CSS or polyfills?](#how-to-add-custom-javascript-and-css-or-polyfills)
 * [How to use React Styleguidist with Preact?](#how-to-use-react-styleguidist-with-preact)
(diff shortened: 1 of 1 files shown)


docs/Cookbook.md
```diff
@@ -86,6 +87,35 @@ global.Button = Button
 
 The `Button` component will be available in every example without a need to `require` it.
 
+## How to dynamically load other components in an example?
+
+Although examples do not direct access to use webpack's `require.context` feature, you *can* use it in a separate
```
sapegin: Did you mean “examples don’t have direct access to webpack’s…”?

docs/Cookbook.md
```diff
@@ -86,6 +87,35 @@ global.Button = Button
...
+helper file which you require in your example code.  As an example, if you wanted to create an example to load and 
+show all of your icon components, you could do this:
```
sapegin: One paragraph — one line. One space at the end of a sentence.

boxfoot:
@sapegin Thanks for the feedback.  Made corrections!

sapegin:
Thanks, merged!

PR #813, opened 2018-02-12
Fix broken fragment link to editorConfig settings

davidjb:
The current link is in camelCase which means that the link to `editorConfig` on this page https://react-styleguidist.js.org/docs/configuration.html doesn't work.  This PR normalises the case to `editorconfig`, allowing the [link](https://react-styleguidist.js.org/docs/configuration.html#editorconfig) to work.


sapegin:
Thanks!

PR #814, opened 2018-02-12
Use editorConfig.theme for CodeMirror theme

### LINKED ISSUE #788: New editorConfig option doesn't seem to work or maybe missing themes?
In the new minor update release (`v6.2.0`) notes it mentions moving from `highlightTheme` to use the `editorConfig` option, but it doesn't seem to work. Or maybe themes were removed?

I used to have `highlightTheme: 'material',`, and then I switched it to: 

```
  editorConfig: {
    theme: 'material',
  },
```

It would look like http://bluetide.pro/FKiiDg when I used the `highlightTheme` option, but now when I use the `editorConfig` it looks like http://bluetide.pro/SBR33b

Have themes been removed apart of this update?

davidjb:
This fixes #788 by firstly testing for `editorConfig.theme` in the config and using that if present.  This falls back to using `highlightTheme` (deprecated) so backwards compatibility is assured.

### DIFF
--- a/scripts/make-webpack-config.js
+++ b/scripts/make-webpack-config.js
@@ -31,7 +31,7 @@ module.exports = function(config, env) {
 		resolve: {
 			extensions: ['.js', '.jsx', '.json'],
 			alias: {
-				'rsg-codemirror-theme.css': `codemirror/theme/${config.highlightTheme}.css`,
+				'rsg-codemirror-theme.css': `codemirror/theme/${config.editorConfig.theme}.css`,
 			},
 		},
 		plugins: [
--- a/loaders/styleguide-loader.js
+++ b/loaders/styleguide-loader.js
@@ -16,7 +16,6 @@ const slugger = require('./utils/slugger');
 // Config options that should be passed to the client
 const CLIENT_CONFIG_OPTIONS = [
 	'title',
-	'highlightTheme',
 	'showCode',
 	'showUsage',
 	'showSidebar',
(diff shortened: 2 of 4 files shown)


scripts/make-webpack-config.js
```diff
@@ -31,7 +31,9 @@ module.exports = function(config, env) {
 		resolve: {
 			extensions: ['.js', '.jsx', '.json'],
 			alias: {
-				'rsg-codemirror-theme.css': `codemirror/theme/${config.highlightTheme}.css`,
+				'rsg-codemirror-theme.css': `codemirror/theme/${(config.editorConfig &&
```
sapegin: There's already a fallback in the config scheme, `highlightTheme` should only be used there.

davidjb:
@sapegin Thanks for pointing that out; I've removed it from  `styleguide-loader.js` and also changed the webpack alias to just use `editorConfig.theme` as per your suggestion.

scripts/make-webpack-config.js
davidjb: Thanks for pointing that out, I've updated the PR accordingly to just use `editorConfig.theme` in that case.

scripts/make-webpack-config.js
```diff
@@ -31,7 +31,8 @@ module.exports = function(config, env) {
 		resolve: {
 			extensions: ['.js', '.jsx', '.json'],
 			alias: {
-				'rsg-codemirror-theme.css': `codemirror/theme/${config.highlightTheme}.css`,
+				'rsg-codemirror-theme.css': `codemirror/theme/${config.editorConfig &&
```
sapegin: This condition doesn’t make sense:

1. `config.editorConfig` is always defined (check the schema)
2. If it wasn’t defined, you’d have a broken path.

davidjb: Updated :+1:.

sapegin:
Thanks, merged!

Issue #807, opened 2018-02-06
[v6.2.2] TypeError: Cannot read property 'line' of undefined


diegohaz:
Using object rest properties on js blocks causes that error:

```js
const { foo, ...rest } = { foo: 'bar', baz: 'qux' };
```

```js
const Example = ({ ...rest }) => <div {...rest} />;
```

It was introduced in v6.2.2. It works on v6.2.1. Probably related to #801.

cc/ @lgordey

lgordey:
May be it's cause of `buble` downgrade but now I can't up it https://github.com/Rich-Harris/buble/issues/107

lgordey:
But yea, support for rest properties was added in `v0.17.0` here https://github.com/Rich-Harris/buble/pull/30

lgordey:
I suppose this will help https://github.com/styleguidist/react-styleguidist/pull/815

Issue #798, opened 2018-01-30
HTML entities not decoded?


FezVrasta:
I can't understand why the HTML entities are not properly rendered:

![image](https://user-images.githubusercontent.com/5382443/35593938-645062ba-05c6-11e8-8b64-ac2531898a3d.png)

Repro:
https://github.com/FezVrasta/example

Specific file: https://github.com/FezVrasta/example/blob/master/lib/components/Button/Readme.md

sapegin:
No idea ;-/ I guess it's a bug somewhere in Styleguidist, but it needs some debugging.

Feel free to debug and / or fix if you have time.

nitinkh:
I think this should work fine:

```
<button className="button" style={styles} onClick={onClick} disabled={disabled} dangerouslySetInnerHTML={{__html: children}}>
</button>
```

nitinkh:
If it works let me know will send a pull request

lgordey:
@sapegin I guess I found failing part - https://github.com/styleguidist/react-styleguidist/pull/801

vysinsky:
Hello, it seems to be broken again in 7.0.7 (also in 7.0.5). I have this list directly in MD file (not in component example)

```markdown
Make sure to include polyfills for following JS features (if you need to support browsers which do not support them):

- fetch - recommended: [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch)
- promises - recommended: [core-js/es6/promise](https://github.com/zloirock/core-js#ecmascript-6-promise)
- Map - recommended: [core-js/es6/map](https://github.com/zloirock/core-js#map)
- Set - recommended: [core-js/es6/set](https://github.com/zloirock/core-js#set)
- ES6 array functions - recommended [core-js/es6/array](https://github.com/zloirock/core-js#ecmascript-6-array)
- requestAnimationFrame (for IE < 10) - recommended [raf](https://www.npmjs.com/package/raf)
```
Styleguide renders `(for IE < 10)` as `(for IE &lt; < 10)`.

vysinsky:
I've just checked and last version it is working correctly is 6.0.33.

lgordey:
This issue was about `buble` bugs related to https://github.com/Rich-Harris/buble/issues/104 and https://github.com/Rich-Harris/buble/issues/107
It should correctly work in `6.2.4` after we updated to `buble@0.19.2` here https://github.com/styleguidist/react-styleguidist/pull/815.

[...]

Issue #764, opened 2018-01-02
Update to v6.1.0 has caused IE11 to no longer work.


rluiten:
When something isn’t working, include:

* Your style guide config (`styleguide.config.js`).
```
const path = require('path');

module.exports = {
  title: 'MFL React Components',
  components: 'src/components/**/*.tsx',
  skipComponentsWithoutExample: true,
  ignore: [],
  require: [
    path.resolve(__dirname, './styleguide/setup.js'),
    path.join(__dirname, './node_modules/enhanced-theme/enhanced/mfl.scss'),
    path.join(__dirname, './src/main.scss')
  ],
  styleguideComponents: {
    Wrapper: path.join(__dirname, './styleguide/Wrapper')
  },
  assetsDir: 'Content'
};
```

* Webpack version.

```
webpack@3.10.0
```

* Operating system, browser and version if it’s a layout bug.

```
Windows 10 64 bit, IE11 only. (chrome, firefox, edge are ok).
```

The totally unmodified styleguidist example https://github.com/styleguidist/example produces same problem in IE11 locally.

There error recieved in IE11 is.

```
SCRIPT1002: Syntax Error
main.bundle.js (50619,39)
```

A screen capture of what is at the location in main.bundle.js follows.

![image](https://user-images.githubusercontent.com/670472/34472314-e1cfc4b0-efaa-11e7-8c1c-f7605c4d1c91.png)

Reverting to styleguidist 6.0.33 allows the example to work.

rluiten:
Just realised might be the fat arrow function!.

sapegin:
Yeah, most likely it’s an arrow function or a `const`. Do you know where this code is coming from?

rluiten:
Is there anything specific I can do ?
I can provide a compressed version of the bundle it is a bit over 450 KB compressed and about 2.6 MB uncompressed. I've snooped around in it to see if i can figure out what module it is it might be buble.

rluiten:
By the way the actual code i posted above is from the example app not from my application.

sapegin:
It’s from [acorn5-object-spread](https://github.com/adrianheine/acorn5-object-spread) which is a dependency of [buble](https://github.com/Rich-Harris/buble). From our side we can only file and issue there and ask what browsers they support.

Andarist:
@sapegin it's not only a dep of `buble` but a `buble` itself too, it doesnt transpile i.e. classes. I've filed an issue there to see what they target with the buble's codebase.

[...]

PR #815, opened 2018-02-12
Up buble version to 0.19.2

lgordey:
This fixes https://github.com/styleguidist/react-styleguidist/issues/807 and doesn't break https://github.com/styleguidist/react-styleguidist/issues/798 as well.
And maybe https://github.com/styleguidist/react-styleguidist/issues/764? I'm actually not sure


sapegin:
Cool, thanks!

PR #730, opened 2017-12-10
add react-native-web example

Gregor1971:
adds a project to the examples dir:
a basic working styleguide for a react-native app using react-native-web
the app was created via crna

... as promised in #675

PR #816, opened 2018-02-12
React-Native example

TerrySmithDC:
Hello!

I recently added react-styleguidest to a production react-native app. Our team really likes your project and we thought It would be nice to give back by providing a slimmed down example for future developers. (Credit to an earlier closed PR as my jumping off point https://github.com/styleguidist/react-styleguidist/pull/730 )

I also added extra explanation in the readme for developers less familiar with Webpack and how to add it to an existing react-native app. (As some react-native developers come from a mobile background and are not so familiar with bundling for the web.)

Any feedback is welcome.


examples/react-native/App.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
sapegin: Can we move it to `src` like in other examples?

examples/react-native/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+(Optional) Run the react-native app via simulator (or [Expo Client](https://expo.io/learn))
+
+```
+npm run start
+```
+
+Then open [http://localhost:6060](http://localhost:6060) in your browser.
+
+## Add styleguidest to an exsisting react-native project
```
sapegin: Styleguidist

examples/react-native/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+How to start locally:
+
+```
+git clone https://github.com/styleguidist/react-styleguidist.git
+cd react-styleguidist/examples/react-native
+npm install
+```
+
+Run styleguidest
```
sapegin: Styleguidist

examples/react-native/package.json
```diff
@@ -0,0 +1,49 @@
+{
+	"name": "react-styleguidist-example-react-native",
```
sapegin: tab → 2 spaces

examples/react-native/styleguide.config.js
```diff
@@ -0,0 +1,56 @@
...
+				},
+				{
+					test: /\.ttf$/,
+					loader: 'file-loader',
+				},
+			],
+		},
+	},
+	dangerouslyUpdateWebpackConfig(webpackConfig) {
+		const newWebpackConfig = webpackConfig;
```
sapegin: Doesn’t make sense: you’re mutating the original param anyway.

examples/react-native/styleguide.config.js
```diff
@@ -0,0 +1,56 @@
...
+
+		// Most react native projects will need some extra configuration, push any needed plugins here.
+		// Use with caution.
+		newWebpackConfig.plugins.push(
```
sapegin: This should work in `webpackConfig` too.

examples/react-native/styleguide.config.js
```diff
@@ -0,0 +1,56 @@
...
+			// Add __DEV__ flag to browser example.
+			new webpack.DefinePlugin({
+				__DEV__: process.env,
+			})
+		);
+
+		// auto resolves any react-native import as react-native-web
+		newWebpackConfig.resolve.alias['react-native'] = 'react-native-web';
```
sapegin: And this.

examples/react-native/App.js
TerrySmithDC: I can move the React Component and logic out of `App.js`, however the configuration for create-react-native-app mounts on `./App.js` so that file needs to remain there unless I eject it the whole app. (For the sake of a simple demo I think we should avoid doing that)

TerrySmithDC:
I've never used Travis before, but it looks like CI is failing because `react-native` is not availble because it's not listed as a dependency on the root package.json

Adding it just for a specific example doesn't seem ideal, what do you suggest is the best way to resolve this @sapegin ?

sapegin:
I think we can remove test cases from the example. I don't think they are useful if we're not going to run them.

sapegin:
Looks like CI is still failing ;-/

sapegin:
You need to disable `import/no-unresolved` ESLint rule for some files.

sapegin:
Thanks, merged!

dilnara:
So, do I understand right that it's react-native-web in fact, no react-native? Or it's really NATIVE app that is run on the device?

PR #818, opened 2018-02-13
Refactor: Introduce Input component to render in Markdown. #613

ryanoglesby08:
So that it easier is for consumers to customize their styleguide by replacing Markdown components.

### DIFF
--- a/src/rsg-components/Markdown/Input/Input.spec.js
+++ b/src/rsg-components/Markdown/Input/Input.spec.js
@@ -1,0 +1,11 @@
+import React from 'react';
+
+import Input from './index';
+
+describe('Markdown Input', () => {
+	it('should render an input with the specified type', () => {
+		const actual = render(<Input type="checkbox" />);
+
+		expect(actual).toMatchSnapshot();
+	});
+});
--- a/src/rsg-components/Markdown/Input/InputRenderer.js
+++ b/src/rsg-components/Markdown/Input/InputRenderer.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Styled from 'rsg-components/Styled';
+
+const styles = () => ({
+	input: {
+		isolate: false,
+		display: 'inline-block',
+		verticalAlign: 'middle',
+	},
+});
+
+export function InputRenderer({ classes, ...rest }) {
+	return <input {...rest} className={classes.input} />;
+}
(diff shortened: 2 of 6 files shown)


ryanoglesby08:
> Are there any other use cases except checkbox?

I had Checkbox at first, then renamed it to Input. I think the checkbox is the only GFM syntax, but you can also put raw html in markdown. So technically someone could do...

```md
## Some heading

<input type="text" />
```

Maybe not a likely use case, but is possible.

Renaming it Checkbox wouldn't really matter for a case like that though.

sapegin:
It won’t use your component, so that use case doesn’t matter.

ryanoglesby08:
It does! Just tested it out locally.

sapegin:
It looks like a bug to me. What would be the use case for that?

ryanoglesby08:
Yeah I don't know, maybe it's a bug with markdown-to-jsx? Seems legit to me though.

> What would be the use case for that?

haha I have no idea! But I'm sure someone would find a use case that we won't think of.

Anyway, I don't mind renaming to Checkbox as that is the intended use. Shall I just make that change then?

sapegin:
> Shall I just make that change then?

Yup!

src/rsg-components/Markdown/Checkbox/CheckboxRenderer.js
```diff
@@ -0,0 +1,21 @@
...
+const styles = () => ({
+	input: {
+		isolate: false,
+		display: 'inline-block',
+		verticalAlign: 'middle',
+	},
+});
+
+export function CheckboxRenderer({ classes, ...rest }) {
+	return <input {...rest} className={classes.input} />;
```
sapegin: Shouldn't we add `type="checkbox"` here? Just for clarity ;-)

ryanoglesby08: Could do.... I'm worried though that if someone happens to create another input type, then this custom component will override it and they will not know why. I think its better to leave it out as marktown-to-jsx already seems to be adding the type.

sapegin: Then it will not be a checkbox and these styles are only meant to be used for a checkbox ;-)

ryanoglesby08: Alright then. Yolo!

Issue #820, opened 2018-02-13
Support ES6 exports


seivan:
I see this issue was closed without any comments and also locked https://github.com/styleguidist/react-styleguidist/issues/633

Is there a reason only default exports are allowed and not named exports? 
We're in a position to rely on ES6 exports to allow tree shaking and would rather not use default exports.

I was wondering if there is any wiggle room here or if you stand strong on this?

marcdavi-es:
Yeah I thought the response to #633 was close-minded. 

A lot of developers use only named exports for a bunch of reasons. `eslint-plugin-import` is in the process of adding [a rule to prohibit the use of defaults exports](https://github.com/benmosher/eslint-plugin-import/issues/889).

Without this I need to hack styleguidist to make it usable.

seivan:
@marcdavi-es Well if you're in a hurry, you can get around it with 
```
const Button = require('./dist/Components/Button').Button;
```

But it's very error prone.

marcdavi-es:
I had a dig around to try a more robust solution. I'm a bit puzzled by what I found.

Given this set up...

````jsx
// styleguide.config.js
module.exports = {
  components: 'path/to/component.js',
};

// path/to/component.js
import React from 'react';

export const SomeComponent = () => <div>Hello, I'm SomeComponent</div>;

// path/to/Readme.md
```js
console.log(SomeComponent);
```

````

...I see in the console `{ SomeComponent: ...,  __esModule: true}`.

That suggests that the exports of `/component.js` are exposed as `SomeComponent`. That's kind of weird because `SomeComponent` is the name of one of the named exports, not the exports as a whole.

In this case, the following 'works'...

````js
// path/to/Readme.md
```js
<SomeComponent.SomeComponent />
```
````

...which seems odd. Perhaps I'm missing something obvious here.

If a contributor is looking at this, I'm super happy to work on a potential fix. I could do with some help pointing to where the modules are expose though. So far I've had a dig around `Preview`, `evalInContext` and associated files but I haven't had any joy.

seivan:
@marcdavi-es I just tested your idea, and as you said, it worked.

kopax:
any update on this?

seivan:
@kopax There's an attached PR https://github.com/styleguidist/react-styleguidist/pull/825

Issue #633, opened 2017-10-11
Only working with default export


hwaterke:
I don't like default export and always write my components with named exports.

But then the only way to make Markdown examples work is to write the following:
```js
const Title = require('./Title').Title;
<Title>Helloo</Title>
```

Any better way to handle named exports and avoid this `require` in all my examples?

sapegin:
> Any better way to handle named exports and avoid this require in all my examples?

Yes, use default exports.

PR #825, opened 2018-02-15
Add limited support for named exports. Add docs relating to this support.

marcdavi-es:
### Summary

This responds to the need expressed in #820 and #633 

It preserves the base case of Styleguidist loading `default` exports by default, while adding some limited support for exposing named exports where a module exposes no default export and either:

- just one named export; or
- a named export which matches the component name understood by Styleguidist.

I think these cases are non-controversial. In each case, it is almost guaranteed that the developer intends these named exports to be interpreted as the exported component.

I have added documentation covering this. The documentation is careful to emphasise that this is not a complete solution for named exports - Styleguidist will do its best but in complicated cases will just fall back to exposing the module as a whole, as it currently does.

Super happy to take questions and feedback on this. I love what you guys are doing with this tool.

### Question

In the existing implementation of `globalizeComponent`, there are references to `component.props.path`. I couldn't get this property to show its face with a bunch of different configurations/attempts. I suspect maybe it's a legacy thing which is still hanging around.

If I'm missing something here please let me know and I'll take account of this in the PR.

### DIFF
--- a/docs/Components.md
+++ b/docs/Components.md
@@ -48,74 +48,116 @@ Styleguidist _loads_ your components and _exposes_ them globally for your exampl
 
 ### Identifier
 
-It will try to use the `displayName` of your component as the identifier.
+It will try to use the `displayName` of your component as the identifier. If it cannot understand a `displayName` (for example if it is dynamically generated), it will fall back to something it can understand.
+
+<table>
+  <tr>
+    <th>Path</th>
+    <th>Code</th>
+    <th>displayName</th>
+    <th>Fallback</th>
+    <th>Global identifier</th>
+  </tr>
+  <tr>
+    <td>/component.js</td>
+    <td>
+      <pre style="margin:0;">export default function Component() { ... }</pre>
+    </td>
+    <td>Component</td>
+    <td>-</td>
+    <td>Component</td>
+  </tr>
+  <tr>
+    <td>/component.js</td>
+    <td>
+      <pre style="margin:0;">
+export default function Component() { ... }
+Component.displayName = 'SomeName';</pre>
+    </td>
+    <td>SomeName</td>
+    <td>-</td>
+    <td>SomeName</td>
+  </tr>
+  <tr>
+    <td>/component.js</td>
+    <td>
+      <pre style="margin:0;">
+export default function Component() { ... }
+Component.displayName = dynamicNamer();</pre>
+    </td>
+    <td>Component
+    </td>
+    <td>- </td>
+    <td>Component</td>
+  </tr>
+  <tr>
+    <td>/component.js</td>
+    <td>
+      <pre style="margin:0;">
+const name = 'Component';
+const componentMap = {
+  [name]: function() { ... }
+};
+export default componentMap[name];</pre>
+    </td>
+    <td>Cannot understand</td>
+    <td>File name</td>
+    <td>Component</td>
+  </tr>
+  <tr>
+    <td>/component/index.js</td>
+    <td>
+      <pre style="margin:0;">
+const name = 'Component';
+const componentMap = {
+  [name]: function() { ... }
+};
+export default componentMap[name];</pre>
+    </td>
+    <td>Cannot understand</td>
+    <td>Folder name</td>
+    <td>Component</td>
+  </tr>
+</table>
+
+
+### Default vs named exports
+
+Stylegudist will use an ECMAScript module’s `default` export or CommonJS `module.exports` if they are defined.
 
 ```javascript
-// /path/to/component-one.js
+// /component.js
 export default function Component() { ... }
 // will be exposed globally as Component
 
-// /path/to/component-two.js
-export default function Component() { ... }
-Component.displayName = 'ComponentTwo'
-// will be exposed globally as ComponentTwo
+// /component.js
+function Component() { ... }
+module.exports = Component;
+// will be exposed globally as Component
 ```
 
-If it cannot understand a `displayName` (for example if it is dynamically generated), it will fall back to something it can understand.
-
-```javascript
-// /path/to/component-three.js
-export default function Component() { ... }
-Component.displayName = getDisplayName() // can't understand this
-// will fall back to the name at declaration and expose globally as Component
-
-// /path/to/component-four.js
-const name = 'Component';
-const componentMap = {
-  [name]: function() { ... } // can't understand this
-}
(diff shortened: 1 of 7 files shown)


marcdavi-es:
One of the tests fails on one of the builds but passes on the others. It seems that the `afterEach` block in `globalizeComponent.spec.js` isn't doing its job in this case.

I tried refactoring the test to avoid reliance on the enumerable properties of `global` but still no joy. Would appreciate any input on this.

docs/Components.md
```diff
@@ -41,6 +42,81 @@ module.exports = {
...
+// /path/to/component-four.js
+const name = 'Component';
+const componentMap = {
+  [name]: function() { ... } // can't understand this
+}
+export default componentMap[name]; // can't understand this either
+// will fall back on the file name, convert it to PascalCase
+// and expose globally as ComponentFour
+```
+
```
sapegin: There's one more use case: folder name when the file name is `index.js`.

docs/Components.md
```diff
@@ -41,6 +42,81 @@ module.exports = {
...
+### default vs named exports
```
sapegin: Default

docs/Components.md
```diff
@@ -41,6 +42,81 @@ module.exports = {
...
+
+Styleguidist defaults to exposing the `default` export from your component file. It understands functions exported as CommonJs `module.exports` as default exports.
```
sapegin: Maybe “Stylegudist will use ECMAScript module’s `default` export or CommonJS `module.exports` if they are defined”?

docs/Components.md
```diff
@@ -41,6 +42,81 @@ module.exports = {
...
+```javascript
+// /path/to/component-eight.js
+export function someUtil() { ... }
+// will not be exposed
+
+export function ComponentEight() { ... }
+// will be exposed globally as ComponentEight
+```
+
+If you export several React components as named exports from a single module, Styleguidist is likely to behave unreliably. If it cannot understand which named export to expose, it will fall back to exposing the module as a whole.
```
sapegin: > it will fall back to exposing the module as a whole.

Not sure what exactly that mean, but most likey you'll not be able to access components, so we can say that instead ;-)

docs/Components.md
```diff
@@ -41,6 +42,81 @@ module.exports = {
 
 > **Note:** Use [getComponentPathLine](Configuration.md#getcomponentpathline) option to change a path you see below a component name.
 
+## Loading and exposing components
+
+Styleguidist _loads_ your components and _exposes_ them globally for your examples to consume.
+
+### Identifier
```
sapegin: Would be nice to make this section shorted and easier to scan. Maybe use one big table instead?

src/utils/getComponent.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Given a component module and a name,
+ * return the appropriate export.
+ *
+ * @param  {object} module
+ * @param  {string} name
+ * @return {function|object}
+ */
+export default function getComponent(module, name) {
```
sapegin: I guess your understanding of this code is very good now, could you please add comments to each variation? Preferably with code examples. This is one of the most obscure parts of the code base ;-/

src/utils/__tests__/globalizeComponent.spec.js
```diff
@@ -1,18 +1,30 @@
 import globalizeComponent from '../globalizeComponent';
 
+jest.mock('../getComponent');
```
sapegin: Can we avoid mocking here? Doesn't look necessary.

src/utils/__tests__/globalizeComponent.spec.js
```diff
@@ -1,18 +1,30 @@
...
+// eslint-disable-next-line import/first
+import getComponent, { returnValue } from '../getComponent';
```
sapegin: ESM imports are hoisted to the top of the file, so disabling ESLint here is unnecessary and misleading.

marcdavi-es:
Thanks for the feedback :)

I'm out this weekend but I'll get to this on Monday.

marcdavi-es:
@sapegin I think that's everything. Let me know if you're happy with those changes :)

docs/Components.md
sapegin: Could be simpler:

```
| foo | bar |
| --- | --- | 
| `foo`<br>`bar` | bar |
```

Also you can remove the last column and use the same component name everywhere, and merge displayName and fallback columns since they are mutually exclusive.

src/utils/getComponent.js
sapegin: Nice! Please add a blank line before each comment to make it easier to read.

docs/Components.md
marcdavi-es: Ah yeah nice. Done.

src/utils/getComponent.js
```diff
@@ -0,0 +1,52 @@
...
+ * Given a component module and a name,
+ * return the appropriate export.
+ * See /docs/Components.md
+ *
+ * @param  {object} module
+ * @param  {string} name
+ * @return {function|object}
+ */
+export default function getComponent(module, name) {
+	//
```
marcdavi-es: Had to leave this little 'comment' in here to stop Prettier complaining about the blank line...

marcdavi-es:
@sapegin Thanks - yeah that table was ugly. Think it's much cleaner now.

src/utils/getComponent.js
sapegin: Sorry, I wasn't clear enough: blank lines to separate code “paragraphs”, not just empty line before each comment ;-)

docs/Components.md
sapegin: Wow, it's like 10 times smaller now ;-)

src/utils/getComponent.js
marcdavi-es: So you mean changing this

````
	// If the module defines a default export, return that
	// e.g.
	// ```
	// export default function Component() { ... }
	// ```
	if (module.default) { ...
````

to this

````
	// If the module defines a default export, return that
	// e.g.
        //
	// ```
	// export default function Component() { ... }
	// ```
        //
	if (module.default) { ...
````

Or something else? :)

sapegin: Exactly! Just to separate different sections of code.

marcdavi-es:
@sapegin Sweet - comments spaced out better now :)

PR #827, opened 2018-02-15
add cypress tests for core and single component

glebez:
Hey,

wrote up some integration tests with cypress (#776).
They feel a bit fragile, I don't like hooking up to css classes or text contents for testing, but I can't come up with better ways yet.

Had some pain with code editor, since the codemirror is using an empty hidden textarea behind the scenes to track keystrokes and update the divs with representation, so the test for this is a bit hacky.

### DIFF
--- a/cypress/integration/component_spec.js
+++ b/cypress/integration/component_spec.js
@@ -1,0 +1,100 @@
+describe('Single component', () => {
+	before(() => {
+		// Open simple button component in isolation
+		cy.visit('/#!/Button/1');
+	});
+
+	describe('props and methods section', () => {
+		beforeEach(() => {
+			cy
+				.get('button')
+				.contains('Props & methods')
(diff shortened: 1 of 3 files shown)


sapegin:
Awesome, thanks!

> They feel a bit fragile, I don't like hooking up to css classes or text contents for testing, but I can't come up with better ways yet.

I think text content (especially for things like buttons) is OK. CSS classes and tab names feel fragile, but I don't see a better way unless we add a `name` to each element we want to access in tests.

> Had some pain with code editor, since the codemirror is using an empty hidden textarea behind the scenes to track keystrokes and update the divs with representation, so the test for this is a bit hacky.

Yeah, that’s expected ;-)

I’m going to merge it as is now, but feel free to send a new PR, if you’ll find a better way to access elements.

PR #828, opened 2018-02-16
Server now returns app instance and compiler instance

cmswalker:
This exposes both the compiler and the webpack-dev-server app. The reason behind this is so that wrappers built around react-stylguidist API can stop the dev server. 

Real World use case: I'm currently using this forked implementation to automatically capture screenshots for a component. StartServer => Redirect to Component => Capture Screenshot => Close Server.  Unfortunately, without these additions, there is no other way to close the server other than from exiting the process.


cmswalker:
Absolutely!

cmswalker:
@sapegin I have updated the docs and merged master

sapegin:
Great, thanks!

PR #835, opened 2018-02-21
Page per section

bitionaire:
Added routing variant that will show sections & components in isolated mode (when `pagePerSection` is set to `true` in the config part), starting with the first section or component for the index page.

There are some minor changes and tweaks in comparison to PR #768. Instead of showing the first component per default, I set the first section or component as the default. I assume most of the larger styleguides start with some kind of _Intro_ or _Getting started_ section anyways. 

This PR should address #494. 

What do you think about it?


sapegin:
Awesome, thanks for picking it up! I haven't looked into the code yet, let's finalize the API first.

> Instead of showing the first component per default, I set the first section or component as the default. I assume most of the larger styleguides start with some kind of Intro or Getting started section anyways.

I thinks that's a great idea!

What do you think about making the option a string instead of a boolean with values `always`, `never`, `static-only`, with the latter as a default value? `static-only` will show one component in dev mode and all in the static build, I believe that would be most convenient for the majority of users.

SaraVieira:
Hi @bitionaire!

This is awesome! This is just what I needed to solve a major problem in our styleguide! Anxiously awaiting for this to get merged!

bitionaire:
> What do you think about making the option a string instead of a boolean with values `always`, `never`, `static-only`, with the latter as a default value? static-only will show one component in dev mode and all in the static build, I believe that would be most convenient for the majority of users.

Sure, maybe there are even more modes we didn't think of yet. Good point! 👍 
I also noticed that the page doesn't scroll to top after `#!/` route changes. I'll fix that, too.

bitionaire:
I just added a few commits, and `pagePerSection` may be defined as a function. `static-only` may look like `pagePerSection: () => (process.env.NODE_ENV !== 'production')`.

I added `window.scrollTo(0,0)` in `componentDidUpdate` of `StyleGuide.js`. The location of this code seems really hacky, and it does not work for all use-cases (e.g. scrolling down and navigating to the same section won't trigger the execution), but I'm not sure where to put it. I'll think about a better solution. If you have any ideas, I'm open to any suggestions :)

tizmagik:
Looking forward to this! 🙏

src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
```diff
@@ -8,7 +8,23 @@ exports[`should ignore items without name 1`] = `
...
+
+exports[`should render links to isolated component page if oneComponentPerPage passed 1`] = `
+<ul>
+  <li
+    className=""
+    key="Button"
+  >
+    <Styled(Link)
+      className=""
+      href="blank#button"
```
sapegin: Is this correct value? Looks like a non-isolated link.

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -44,6 +47,15 @@ export default class StyleGuide extends Component {
 		};
 	}
 
+	componentDidUpdate(prevProps) {
+		// scroll to top of styleguide container when sections changed
```
sapegin: Probably this will be a better place: https://github.com/styleguidist/react-styleguidist/blob/master/src/utils/renderStyleguide.js

Though you may need to cache current route.

src/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -66,8 +86,11 @@ export default class StyleGuide extends Component {
 			<StyleGuideRenderer
 				title={config.title}
 				homepageUrl={HOMEPAGE}
-				toc={<TableOfContents sections={sections} />}
-				hasSidebar={config.showSidebar && displayMode === DisplayModes.all}
+				toc={<TableOfContents sections={allSections} useIsolatedLinks={pagePerSection} />}
+				hasSidebar={
+					(pagePerSection && displayMode !== DisplayModes.example) ||
```
sapegin: I guess it's time to move this condition to a function with a good comment ;-)

src/utils/renderStyleguide.js
```diff
@@ -27,7 +28,9 @@ export default function renderStyleguide(
 	// all components accessible to all examples
 	globalizeComponents(allSections);
 
-	const { sections, displayMode } = getRouteData(allSections, loc.hash);
+	const { pagePerSection } = styleguide.config;
+	const renderPagePerSection = isFunction(pagePerSection) ? pagePerSection() : pagePerSection;
```
sapegin: We should convert function to a boolean in config schema and later use boolean values, would be simpler. 

But since we don't pass any dynamic data to this function, we probably don't need it. This would also work:

```javascript
module.exports = {
  pagePerSection: process.env.NODE_ENV !== 'production'
}
```

src/rsg-components/ComponentsList/__snapshots__/ComponentsList.spec.js.snap
bitionaire: Good catch! I missed that in my refactoring

src/rsg-components/StyleGuide/StyleGuide.js
bitionaire: I thought about that. It's way more efficient to watch on location changes, yet it would mean to replace the functional component something like an "App" component or to merge the logic with the StyleGuide.js. What do you think is the better way?

src/rsg-components/StyleGuide/StyleGuide.js
bitionaire: You're right

src/utils/renderStyleguide.js
bitionaire: Sure. Just reverted the changes. So there's no need for an enum like 'always', 'never' or 'static-only', right?

sapegin: I think I could just add this as a default value with the next major release.

src/index.js
```diff
@@ -2,16 +2,21 @@
...
 
 // Examples code revision to rerender only code examples (not the whole page) when code changes
+// eslint-disable-next-line no-unused-vars
 let codeRevision = 0;
 
 const render = () => {
 	// eslint-disable-next-line import/no-unresolved
 	const styleguide = require('!!../loaders/styleguide-loader!./index.js');
-	ReactDOM.render(renderStyleguide(styleguide, codeRevision), document.getElementById('app'));
+	ReactDOM.render(
```
sapegin: I don't think it was necessary: it should rerender everything on hash change, so you can scroll in `renderStyleguide`.

bitionaire: I didn't want to introduce any kind of global variable; either the hash of the last location or a boolean that will indicate if an event listener for `hashchange` was already registered. 
I think it'll be easier to handle the component (or event handler) lifecycle in the `<App />`. But it's your project so you decide. Should I revert the changes and add an event listener for the first time `renderStyleguide` will be called?

sapegin: Yes, please. I think it makes this part much more complex then it was. 

What kind of an event handler you want to add? There's a `hashchange` handler already that rerenders everything.

bitionaire:
Allright, please have look 👍

src/index.js
```diff
@@ -4,14 +4,19 @@ import './polyfills';
...
 // Examples code revision to rerender only code examples (not the whole page) when code changes
+// eslint-disable-next-line no-unused-vars
 let codeRevision = 0;
 
 const render = () => {
 	// eslint-disable-next-line import/no-unresolved
 	const styleguide = require('!!../loaders/styleguide-loader!./index.js');
-	ReactDOM.render(renderStyleguide(styleguide, codeRevision), document.getElementById('app'));
+	ReactDOM.render(renderStyleguide(styleguide, status), document.getElementById('app'));
+	if (get(window, 'location.hash', '').startsWith('#!/')) {
```
sapegin: 1. This needs a comment.

2. Would it be triggered on code updates too? Probably better to have a separate `hashchange` just for this.

3. I guess `status` instead of `codeRevision` wasn't intended.

src/index.js
```diff
@@ -4,14 +4,19 @@ import './polyfills';
 import './styles';
 import ReactDOM from 'react-dom';
 import renderStyleguide from './utils/renderStyleguide';
+import { get } from 'lodash';
```
sapegin: Please don't import whole Lodash on frontend. But I think you don't really need it here, are there any cases where `location.hash` won't be defined?

src/rsg-components/ComponentsList/index.js
```diff
@@ -1 +1 @@
-export { default } from 'rsg-components/ComponentsList/ComponentsListRenderer';
+export { default } from './ComponentsList.js';
```
sapegin: All component imports should be like `rsg-components/ComponentsList/ComponentsListRenderer`, see more info here: https://react-styleguidist.js.org/docs/development.html

src/utils/getRouteData.js
```diff
@@ -50,7 +51,18 @@ export default function getRouteData(sections, hash) {
...
 		}
+	} else if (pagePerSection) {
+		// If one component per page mode then show demos for first component
+		sections = [getFirstSectionOrComponent(sections)];
 	}
 
 	return { sections, displayMode };
 }
+
+function getFirstSectionOrComponent(sections) {
```
sapegin: Could you move this function before the default export and use an arrow function? That's the convention we use for such smallish functions.

bitionaire:
I think we're getting close 🍻

src/index.js
```diff
@@ -1,20 +1,28 @@
...
 import ReactDOM from 'react-dom';
 import renderStyleguide from './utils/renderStyleguide';
 
 // Examples code revision to rerender only code examples (not the whole page) when code changes
+// eslint-disable-next-line no-unused-vars
 let codeRevision = 0;
 
+/** Scrolls to origin when current window location hash points to an isolated view. */
+const scrollToOrigin = () => {
+	if (window.location.hash.startsWith('#!/')) {
```
sapegin: I'm afraid `startsWith` won't work in IE11, everything else looks fine for me.

bitionaire: You're right! I replaced it with `indexOf === 0`

sapegin:
Awesome, thanks! I'll try to make a release soon 🦄

bitionaire:
Hey @sapegin,

I've added one change, can you have a look at that? 
It's hard to define a correct behaviour for the `IsolateButton` if `pagePerSection=true`. Therefore I excluded the button from any toolbars for the `slots` object, based on the styleguide configuration.

PR #842, opened 2018-02-23
Wrap Examples in Fragment

### LINKED ISSUE #840: Wrap examples in React.Fragment
This will allow to render multiple components without wrapping them in some `<div>`:

```jsx
// Before
<div>
  <Button>push</Button>
  <Button secondary>push</Button>
</div>

// After
<Button>push</Button>
<Button secondary>push</Button>
```

We should keep React 15 compatibility and check if `React.Fragment` is available. Keep the existing behavior in React 15.

tizmagik:
Here's a quick & dirty initial implementation to close #840 

image

This wraps every example, but we could regex the error message and only wrap if we get the "adjacent JSX" error, but that seems like added complexity for not much gain.

If this implementation looks OK, I can add tests and close this out.

TODO:
 - [] Tests

### DIFF
--- a/src/rsg-components/Preview/Preview.js
+++ b/src/rsg-components/Preview/Preview.js
@@ -119,7 +119,11 @@ export default class Preview extends Component {
 
 	compileCode(code) {
 		try {
-			return compileCode(code, this.context.config.compilerConfig);
+			const wrappedCode = `
+const __F = React.Fragment || <div />;
+<__F>${code}</__F>
+	`;
+			return compileCode(wrappedCode, this.context.config.compilerConfig);
 		} catch (err) {
 			this.handleError(err);
 		}


tizmagik:
So one test is failing, which is the "pizza" (syntax error case):

https://github.com/styleguidist/react-styleguidist/blob/68a72a4076c180f30b3d4135ceb3d162746720aa/src/rsg-components/Preview/Preview.spec.js#L34

Which makes sense because since it's now wrapped it's technically no longer a syntax error. This makes me think we _should_ regex the error and only wrap if we get that Adjacent JSX error. Thoughts?

tizmagik:
Ok this only wraps on error now

sapegin:
Thanks!

> we could regex the error message and only wrap if we get the "adjacent JSX" error, but that seems like added complexity for not much gain.

I don't think it's necessary, but we should document how it works.

tizmagik:
Turns out it was necessary because of the test case I linked to above.

Where/how do you think it should be documented?

sapegin:
The test case could be changed — there are many way to reproduce the error, like `<Pizza` ;-)

Somewhere here I think: https://react-styleguidist.js.org/docs/documenting.html

tizmagik:
It's still a valid test case since we only want to trigger the wrapping _iff_ the error is an Adjacent JSX error (since wrapping is the only way to fix it, which we can now do automatically for the user).

Cool. I'll add to the docs when I get a chance.

sapegin:
I'm still not convinced that conditional wrapping is needed here ;-)

tizmagik:
Are you saying to always wrap? But then undefined reference error would get swallowed

sapegin:
Undefined reference isn't a problem at all ;-) But this is:

```js
const foo = 42;
<span>{foo}></foo>
```

So probably we should check if an example starts from `<` and wrap only in that case.

src/rsg-components/Preview/Preview.js
```diff
@@ -119,7 +119,10 @@ export default class Preview extends Component {
 
 	compileCode(code) {
 		try {
-			return compileCode(code, this.context.config.compilerConfig);
+			return compileCode(
+				code.trim().startsWith('<') ? wrapCodeInFragment(code) : code,
```
sapegin: `startsWith` won't work in IE11.

tizmagik: Classic IE

PR #843, opened 2018-02-26
Issue #774: Adding support to location array

cemarte:
Hi,

This is my first time contributing, so let me know if I've done it correctly and how can I help with other issues. I've committed some changes to ./loades/utils/getComponentFiles.js, but not sure how to test it properly so codecov decreased a little. Can I follow the same pattern in styleguideloader.spec.js to get this part covered?

Thanks!

### DIFF
--- a/loaders/__tests__/styleguide-loader.spec.js
+++ b/loaders/__tests__/styleguide-loader.spec.js
@@ -94,6 +94,34 @@ it('should return correct component paths: function returning absolute paths', (
 	expect(result).not.toMatch(`'filepath': 'components/RandomButton/RandomButton.js'`);
 });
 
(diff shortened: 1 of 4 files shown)


loaders/__tests__/styleguide-loader.spec.js
```diff
@@ -94,6 +94,34 @@ it('should return correct component paths: function returning absolute paths', (
 	expect(result).not.toMatch(`'filepath': 'components/RandomButton/RandomButton.js'`);
 });
 
+it('should return correct component paths: function returning absolute paths', () => {
```
sapegin: function returning absolute paths → array of component absolute paths.

You can also add a test for relative paths.

PR #844, opened 2018-02-26
refactor: introduce Table components for use in Markdown #613

ryanoglesby08:
This makes it easier for consumers to override individual components in their styleguide, specifically tables in this PR.

* Update markdown-to-jsx to v6.4.1 to get some table rendering bug fixes. [Release notes](https://github.com/probablyup/markdown-to-jsx/releases/tag/6.3.2)

### DIFF
--- a/src/rsg-components/Markdown/Markdown.js
+++ b/src/rsg-components/Markdown/Markdown.js
@@ -12,6 +12,7 @@ import Blockquote from 'rsg-components/Markdown/Blockquote';
 import Pre from 'rsg-components/Markdown/Pre';
 import Code from 'rsg-components/Code';
 import Checkbox from 'rsg-components/Markdown/Checkbox';
+import { Table, TableHead, TableCell } from 'rsg-components/Markdown/Table';
 
 // We’re explicitly specifying Webpack loaders here so we could skip specifying them in Webpack configuration.
 // That way we could avoid clashes between our loaders and user loaders.
@@ -106,6 +107,21 @@ const getBaseOverrides = memoize(classes => {
 		input: {
 			component: Checkbox,
 		},
+		table: {
+			component: Table,
+		},
+		thead: {
+			component: TableHead,
+		},
+		th: {
+			component: TableCell,
+			props: {
+				header: true,
+			},
+		},
+		td: {
+			component: TableCell,
+		},
 	};
 }, () => 'getBaseOverrides');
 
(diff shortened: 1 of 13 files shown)


src/rsg-components/Markdown/Markdown.js
```diff
@@ -134,24 +150,6 @@ const styles = ({ space, fontFamily, fontSize, color }) => ({
...
 		borderStyle: 'solid',
 	},
-	table: {
-		composes: '$para',
-		borderCollapse: 'collapse',
-	},
-	thead: {
-		composes: '$hr',
-	},
-	tbody: {},
```
sapegin: Should we create `TableBody` and `TableRow` component  to make styling possible?

src/rsg-components/Markdown/Table/TableCellRenderer.js
```diff
@@ -0,0 +1,34 @@
...
+		fontSize: fontSize.base,
+	},
+	th: {
+		fontWeight: 'bold',
+	},
+});
+
+export function TableCellRenderer({ classes, header, children }) {
+	if (header) {
+		return <th className={cx(classes.td, classes.th)}>{children}</th>;
```
sapegin: We can use JSS composition (like in the original version: `composes: '$td'`) and avoid classnames, I this it would be a little bit nicer ;-)

src/rsg-components/Markdown/Table/TableRenderer.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { styles as paraStyles } from 'rsg-components/Para';
+import Styled from 'rsg-components/Styled';
+
+const styles = ({ space, color, fontFamily }) => ({
+	table: {
+		...paraStyles({ space, color, fontFamily }).para,
```
sapegin: I think we should define color and line-height on td, and then we only need margin from `Para` which whould be copied here.

src/rsg-components/Markdown/Markdown.js
ryanoglesby08: Added!

src/rsg-components/Markdown/Table/TableCellRenderer.js
ryanoglesby08: Done!

src/rsg-components/Markdown/Table/TableRenderer.js
ryanoglesby08: Done!

sapegin:
Thanks, merged!

PR #849, opened 2018-02-28
fix: configure webpack 4

KnisterPeter:
To have webpack 4 configured correctly we need to set
the webpack mode.

### DIFF
--- a/scripts/make-webpack-config.js
+++ b/scripts/make-webpack-config.js
@@ -2,7 +2,6 @@
 
 const path = require('path');
 const webpack = require('webpack');
-const webpackVersion = require('webpack/package.json').version;
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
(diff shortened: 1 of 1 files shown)


scripts/make-webpack-config.js
```diff
@@ -2,6 +2,7 @@
 
 const path = require('path');
 const webpack = require('webpack');
+const webpackVersion = require('webpack/package.json').version;
```
sapegin: We have the `getWebpackVersion` function for that.

scripts/make-webpack-config.js
```diff
@@ -50,6 +51,9 @@ module.exports = function(config, env) {
 			hints: false,
 		},
 	};
+	if (parseInt(webpackVersion[0], 10) >= 4) {
```
sapegin: You need an empty line before if.

KnisterPeter:
@sapegin Thanks for your hints. Updated accordingly.

sapegin:
Thanks, merged!

PR #855, opened 2018-03-05
fix(package.json): Updating html-webpack-plugin, fixing compilation.m…

boazhoch:
fix(package.json): Updating html-webpack-plugin, fixing compilation.mainTemplate.applyPluginsWaterfall error, related to incompatible version between webpack 4 and html-webpack-plugin.

[See issue here](https://github.com/jantimon/html-webpack-plugin/issues/841)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5921,23 +5921,19 @@
       }
     },
     "html-webpack-plugin": {
-      "version": "2.30.1",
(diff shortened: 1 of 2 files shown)


boazhoch:
@sapegin Well it's a breaking change, but currently webpack 4 and react-styleguidist are incompatible, i would consider having a /next branch - anyway its for your consideration.

sapegin:
You can use this PR’s branch.

PR #863, opened 2018-03-07
Docs: Tweak issue template

longility:
In order for issue writers to accurately reproduce using examples,
add a step to make aware and remind issue writers about rsg npm package version.


sapegin:
I think we should rather remove `package-lock.json` from the example repo. But I don't know how to make sure people use the right repo (mentioned in the issue template, not the main repo).

longility:
Ah, I just realized what you meant. I've been using the examples in the main repo and not the one on issue template. Let me think about it some more. In the meantime, let me close this out.

PR #865, opened 2018-03-09
chore: move cypress folder inside test

glebez:
Hi,

this PR is another step for issue #776, moving the cypress folder from root of the project into the test folder.

There's a [know issue](https://github.com/cypress-io/cypress/issues/1372), that even with custom folder structure specified, it still scaffolds default cypress folder on the first run, so I added it to the .gitignore file.


sapegin:
Thanks!

PR #868, opened 2018-03-12
feat: add updateProps config option

ryanoglesby08:
I am working on a project using react-styleguidist where all the components are versioned npm packages. I don't like that the component version is duplicated in both the package.json and the component file, which is causing me to have to manually update it in the component file when I do a release (or forget to do that). It would be more maintainable to load the version from the package.json file so that they are always in sync.

This PR is a WIP to demonstrate this.

Would like some feedback on it and then can clean it up and build it out as a full solution.


ryanoglesby08:
And of course a more complete solution would have to check it the version is a path or if its a number, among other things :)

sapegin:
I'd accept a more generic solution: a function to process an object returned by `getProps` function. So you'll read `package.json` in your style guide config.

ryanoglesby08:
Ok, attempt 2.

I've introduced a new config option called `propsPostProcessor` of type `function`. If it exists it will get invoked after loading and initial processing of the props.

This shifts the responsibility for loading the version from a file to the user.

What do you think?

scripts/schemas/config.js
```diff
@@ -131,6 +131,9 @@ module.exports = {
 	propsParser: {
 		type: 'function',
 	},
+	propsPostProcessor: {
```
sapegin: `updateProps` would be a better name to match `updateExample`: https://react-styleguidist.js.org/docs/configuration#updateexample

examples/basic/styleguide.config.js
```diff
@@ -1,6 +1,20 @@
+const path = require('path');
+
 module.exports = {
 	components: 'src/components/**/[A-Z]*.js',
 	defaultExample: true,
+	propsPostProcessor(props, file) {
```
sapegin: Can we move this to another example styleguide? I'd like to use basic example for the core functionality.

examples/basic/styleguide.config.js
```diff
@@ -1,6 +1,20 @@
...
+		// If there is a filename for the version, then load it in.
+		if (props.doclets.version) {
```
sapegin: More realistic example would be: check if `@version` has a value: if it has, use it, otherwise read version from `version.json`.

examples/basic/styleguide.config.js
ryanoglesby08: I've removed the example and instead included documentation for `updateProps`

examples/basic/styleguide.config.js
ryanoglesby08: I can make the sample code in the docs more complex if you like.... but I feel that it won't add anything since its just toy sample code.

ryanoglesby08:
Hi @sapegin any more feedback here? I've reverted the old code here and Travis CI is passing now.

Issue #864, opened 2018-03-08
Using state in combination with Immutable JS


Baudin999:
Hi, I'm trying to create a styleguide while using Immutable JS. I find it important to keep my example code as close to the actual code as possible. I was wondering if I'm doing something wrong?

The [example](https://github.com/Baudin999/example) solution.

The actual code:

```jsx
const { Map } = require("immutable");
initialState = Map({ index: 0 });
(() => {
  //this will fail as well 
  console.log(state.get("index"));
})();

<div>
  <button
    onClick={() => {
      console.log(state);
      // this will fail
      setState(state.set("index", state.get("index") + 1));
    }}
  >
    +
  </button>
  <div>Our current index is: {state.index || 0}</div>
</div>;
```

Is there a workaround? Or should I start digging through the source code and see if I can find a solution?

sapegin:
I believe that’s because of this line:

https://github.com/styleguidist/react-styleguidist/blob/dae9d6696449f25cdc6f18620074cde785776333/src/rsg-components/Preview/Preview.js#L29

We could probably just do `this.state = initialState`. Feel free to submit a pull request ;-)

Baudin999:
Cool, thank you, will try and submit it tonight.

PR #870, opened 2018-03-14
fix: change state assigment

nicoffee:
Must solve #864.

### DIFF
--- a/src/rsg-components/Preview/Preview.js
+++ b/src/rsg-components/Preview/Preview.js
@@ -26,7 +26,7 @@ class PreviewComponent extends Component {
 	// Synchronously set initial state, so it will be ready before first render
 	// Ignore all consequent calls
 	setInitialState(initialState) {
-		Object.assign(this.state, initialState);
+		this.state = initialState;
 		this.setInitialState = noop;
 	}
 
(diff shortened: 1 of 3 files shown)


sapegin:
Thanks! Would be awesome if you can add some tests for this part of the code.

nicoffee:
Okay, will do.

nicoffee:
@sapegin, i need some help testing this case.
Should i export PreviewComponent from `Preview.js` in order to test his `setInitialState()` method directly, or should i somehow test it through `Preview` component, and if so, how exactly?
For now, i did something like this:
```
it('should synchronously set initialState before first render', () => {
	const actual = mount(<Preview code={code} evalInContext={evalInContext} />, options);
	const testState = {
		error: 'error',
		count: '42',
	};
	actual.setState(testState);
	expect(actual.state()).toEqual(testState);
});
```
But i pretty sure it's wrong, because it actually renders element before state is changing.

sapegin:
@nicoffee Sorry for the delay. Ideally we shouldn't call methods directly, but I'm not sure it's possible in that case. I'll try to have a look myself.

sapegin:
It actually doesn't work. I've fixed the implementation and have added couple of tests.

nicoffee:
Thank you!

PR #871, opened 2018-03-14
fix docs: typo in cookbook link to storybook difference

StoicLoofah:
I noticed that the link was broken on the page due to an extra "e" in both the text and link. I also added a question mark onto the end of the text to match the rest of the formatting on the page.


sapegin:
Niiice, thank you! 🦄

Issue #876, opened 2018-03-19
Need a way to customize CLI message on server started…


roblevintennis:
How would you feel about being able to hijack the `printInstructions` via styleguide config? e.g.

`printInstructions: whiteLabelFunction`

Here's the use case:
- Update your /etc/hosts so that a custom url is replacing `localhost:6060`
- even if you try to supply the `serverHost`, you'll won't be able to adjust the messaging since port is hard coded in call to:

https://github.com/styleguidist/react-styleguidist/blob/03896941d8a373a5a2e6ba1878899442ab660d40/bin/styleguidist.js#L182

e.g.
![image](https://user-images.githubusercontent.com/142403/37617192-951a7bfa-2b6f-11e8-8986-79269b08bc0c.png)

ignore the `:0` because I was trying to hack away the port unsuccessfully.

So, we'd need to either:
- allow for a no-op port like -0 arg to `serverPort` which can magically have the port removed from cli message, or
- let us just customize the `printInstructions` method via config as suggested
- maybe there's a `serverUrl` that somehow takes priority over the host / port configs for this sort of special case

Any other thoughts on how this could be achieved? I'm happy to code a PR if I'm clear on if/how this would be accepted. Thanks!

sapegin:
I think the second (`printInstructions`) would be right way to do that. But I'm not sure how flexible should it be. In future we may need more useful information here. And there’s also similar thing for the production build.

Would like to hear more opinions and use cases.

roblevintennis:
Well, thinking iteratively (or is it agile??), my thought was to just take the same method signature as is there now, and anytime that changes, the callback config option would want to reflect that I suppose. In the end, the library code would simply check if there was a config.method to call, and if so use same arguments. If not, business is usual and the internal `printInstructions` is used.

In terms of opinions and use cases, I just have the one at present. Now that I know the printInstructions method via config is plausible, I might just go ahead and code up a PR so we can have a diff to visually consider. Thanks for getting back @sapegin !

sapegin:
I'd change arguments to an object — would make adding new features easier.

And I think we need a similar function for build instructions, so `printServerInstructions` and `printBuildInstructions`.

roblevintennis:
[...]

PR #878, opened 2018-03-19
Spike on rearranging the build and server methods.

roblevintennis:
WIP for rearranging the print build and server methods with hopes to expose as overridable `styleguide.config` methods. this is for https://github.com/styleguidist/react-styleguidist/issues/876

I wasn't sure how this should be verified so I essentially copied examples/basic to examples/custom-message with the idea I'd get it to print the custom console message.


roblevintennis:
@sapegin I'm a bit stuck on what to do for configuring the schema and/or getting the `sanitizeConfig` to work:

image

That's when I try to run my new custom message example. I tried to log or throw error in `scripts/utils/sanitizeConfig.js` but it's not showing up in my console so I'm obviously totally misunderstanding something there, so any guidance there would be appreciated

examples/custom-message/.babelrc
```diff
@@ -0,0 +1,9 @@
+{
```
sapegin: Please use one of the existing examples (like customised), creating a new one would be an overkill ;-)

That will probably fix the issue with unknown config option too.

roblevintennis: haha ah ok, gotcha, I'm headed to work now...will give it a whirl thanks for help @sapegin :)

roblevintennis:
So the issue I was having was that I intuitively ran the examples from examples dir like: `examples/foo$ npm run styleguide` which of course meant it'd use the node modules styleguidist. But having a look at package.json we have these covered and locally pointing to correct styleguidist.js e.g.  `yarn start:customised` in my case. Fixed the issue. I'm going to see if there's a place to add this in contributing on another PR :)

bin/styleguidist.js
```diff
@@ -94,8 +94,10 @@ function commandBuild() {
 		if (err) {
 			console.error(err);
 			process.exit(1);
+		} else if (typeof config.printBuildInstructions === 'function') {
```
sapegin: We can be sure that if a config option has a value, then it's a function (it'd fail before otherwise), so `if (config.printBuildInstructions)` would be enough.

bin/styleguidist.js
```diff
@@ -120,7 +122,12 @@ function commandServer() {
 			console.error(err);
 		} else {
 			const isHttps = compiler.options.devServer && compiler.options.devServer.https;
-			printInstructions(isHttps, config.serverHost, config.serverPort);
+			const configuration = Object.assign({ isHttps }, config);
+			if (typeof config.printServerInstructions === 'function') {
+				config.printServerInstructions(configuration);
```
sapegin: I'd do `config.printServerInstructions(config, { isHttps });`, so `config` would be always a styleguide config, and extra options would be separated — less cognitive load ;-)

docs/Configuration.md
```diff
@@ -259,6 +261,18 @@ module.exports = {
 }
 ```
 
+#### `printBuildInstructions`
+
+Type: `Function`, optional
+
+Function that allows you to override the printing of build messages to console.log. Takes the `config` object with `isHttps` merged in.
```
sapegin: I’d show an example instead of `Takes the `config` object with `isHttps` merged in.`.

loaders/styleguide-loader.js
```diff
@@ -19,6 +19,8 @@ const CLIENT_CONFIG_OPTIONS = [
 	'showCode',
 	'showUsage',
 	'showSidebar',
+	'printBuildInstructions',
```
sapegin: I don't think you need this, we use this array to pass config options to React.

bin/styleguidist.js
roblevintennis: Addressed in 0eda93

bin/styleguidist.js
roblevintennis: Gotcha. Fixed in 0eda93bd337d7166065781f9d59b7de6285df848

docs/Configuration.md
roblevintennis: Great point. I removed and added example for both methods in 0eda93bd337d7166065781f9d59b7de6285df848

loaders/styleguide-loader.js
roblevintennis: Removed and ran locally and nothing bad happened. Thanks for heads up

roblevintennis:
@sapegin thanks for your guidance. I think this is ready for another gander

sapegin:
Thanks! Will include in the next release ;-)

roblevintennis:
awesome thanks!

PR #880, opened 2018-03-20
Add documentation on how to dev on examples other then basic (which i…

roblevintennis:
Discovered when working on https://github.com/styleguidist/react-styleguidist/pull/878 PR .. while it's discernible from from examining the package.json scripts, it's pretty intuitive to want to cd in to examples/* directories and initialize those apps there (which results in pointing to THAT particular example's node_modules version of styleguidist, so changes to top level dev styleguidist won't have any effect). These directions might have made things easier for me. Adds the following to Other notes
section:

image

### DIFF
--- a/.github/Contributing.md
+++ b/.github/Contributing.md
@@ -52,6 +52,18 @@ npm run test:jest -- -u
 * We’re using [Prettier](https://github.com/prettier/prettier) to format JavaScript, so don’t worry much about code formatting.
 * Don’t commit generated files, like minified JavaScript.
 * Don’t change version number and change log.
+* If you're updating examples other then `examples/basic`, you'll need to modify your watch and start commands:
+
+```bash
+npm run compile:watch & npm start:customised # if making changes to examples/customized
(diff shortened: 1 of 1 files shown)


.github/Contributing.md
```diff
@@ -52,6 +52,18 @@ npm run test:jest -- -u
...
 * Don’t commit generated files, like minified JavaScript.
 * Don’t change version number and change log.
+* If you're updating examples other then `examples/basic`, you'll need to modify your watch and start commands:
+
+```bash
+npm run compile:watch & npm start:customised # if making changes to examples/customized
+npm run compile:watch & npm start:sections # if making changes to examples/sections
+```
+
+See the `scripts` section of the top level `package.json`. If an example doesn't have a script just point to it's config:
```
sapegin: A link to a file would be useful here.

just → “” ;-)
it's → its

roblevintennis: Argh, grammarze 😋 .. fixed

sapegin:
Thanks, merged!

roblevintennis:
sweet! thanks

PR #887, opened 2018-03-23
Fix the bug when section name is not english

ZxBing0066:
Fix the bug when section name is not english such as chinese, then switch router will be failed.

image

repeat step:
1. set option `pagePerSection: true`
2. change some section name to chinese like `测试`
3. click the link to the section
4. click other section link
5. click the chinese section again

### DIFF
--- a/src/utils/__tests__/getInfoFromHash.spec.js
+++ b/src/utils/__tests__/getInfoFromHash.spec.js
@@ -11,7 +11,7 @@ describe('getInfoFromHash', () => {
 		expect(result).toEqual({});
 	});
 
-	it('should return the right targetName when router name is not english such as chinese', () => {
+	it('should return the decoded targetName when router name is not english such as chinese', () => {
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/getInfoFromHash.spec.js
```diff
@@ -10,4 +10,9 @@ describe('getInfoFromHash', () => {
 		const result = getInfoFromHash('Button');
 		expect(result).toEqual({});
 	});
+
+	it('should return the right targetName when router name is not english such as chinese', () => {
```
sapegin: right → decoded?

English, Chinese

ZxBing0066: OK，I have changed `right` to `decoded`.

sapegin: Please also capitalize language names.

ZxBing0066: 😂 Sorry for missed about this.

PR #890, opened 2018-03-25
Add a note about working with react-css-themr

n1313:
This PR adds a tiny note to documentation about making RSG work with react-css-themr.


sapegin:
Thanks! 🦄

PR #891, opened 2018-03-25
Update Readme.md

n1313:
This PR moves the "advanced configuration" block next to the "usage" block so that two lists of links to documentation are closer together and are not separated by several paragraphs of non-documentation.


sapegin:
Thanks!

n1313:
My pleasure.

PR #907, opened 2018-03-29
WIP: html-webpack-plugin → mini-html-webpack-plugin

### LINKED ISSUE #896: Replace html-webpack-plugin with mini-html-webpack-plugin
## Pros

* Much simpler but enough for our use case.
* Users could use any version of html-webpack-plugin and it won't break Styleguidist.
* Use [mini-html-webpack-template](https://github.com/vxna/mini-html-webpack-template) to customize template without replacing the whole HTML file (that should be enough for most use cases for custom templates). 

## Cons

* Breaking change because existing custom templates won’t work (actually template files won’t be supported at all but a template function + mini-html-webpack-template options).

## API proposal

`template` options won’t accept a string but an object that will be passed to the `context` options of mini-html-webpack-template. We’ll need to overwrite some properties in this object like `title` or `container`.

```
module.exports = {
  template: {
    head: {
      links: [
        {
          rel: 'shortcut icon',
          href: 'https://assets-cdn.github.com/favicon.ico',
          type: 'image/x-icon'
        }
      ]
    }
  }
}
```

## More details

* [mini-html-webpack-plugin](https://github.com/bebraw/mini-html-webpack-plugin)
* [mini-html-webpack-template](https://github.com/vxna/mini-html-webpack-template)

sapegin:
* [x] Implementation
* [x] Tests
* [x] Docs

Closes #896

### DIFF
--- a/scripts/__tests__/config.spec.js
+++ b/scripts/__tests__/config.spec.js
@@ -214,3 +214,11 @@ it('should merge default editorConfig with options provided by the user', () => 
 	expect(result.editorConfig).toHaveProperty('mode', 'js');
 	expect(result.editorConfig).toHaveProperty('theme', 'base16-light');
 });
+
+it('should throw when old template as a string option passed', () => {
+	const fn = () =>
+		getConfig({
+			template: 'pizza',
+		});
+	expect(fn).toThrow('format has been changed');
+});


scripts/schemas/config.js
```diff
@@ -230,9 +238,18 @@ module.exports = {
...
-		default: path.resolve(__dirname, '../templates/index.html'),
-		example: 'templates/styleguide.html',
+		type: ['object', 'function', 'existing file path'],
+		default: {},
+		process: val => {
+			if (typeof val === 'string') {
+				throw new StyleguidistError(
+					`${chalk.bold(
+						'template'
+					)} config option format was changed, you need to update your config.`
```
okonet: `was` => `has been`? And probably a link to docs?

Issue #918, opened 2018-04-02
Option to open a browser


sapegin:
Similar to Create React App. There's an unfinished PR: #748 that could be a good starting point.

PR #920, opened 2018-04-02
Added new cli option, auto-open browser #918

rafaesc:


bin/styleguidist.js
```diff
@@ -122,11 +123,17 @@ function commandServer() {
...
 			const isHttps = compiler.options.devServer && compiler.options.devServer.https;
+			const host = config.serverHost;
+			const port = config.serverPort;
+			const urls = webpackDevServerUtils.prepareUrls(isHttps ? 'https' : 'http', host, port);
 			if (config.printServerInstructions) {
 				config.printServerInstructions(config, { isHttps });
 			} else {
 				printServerInstructions(config, { isHttps });
 			}
+			if (argv.open) {
```
sapegin: This would be much easier to read:

```js
// (A new line after the previous code paragraph
if (argv.open) {
  const protocol = isHttps ? 'https' : 'http';
  const { localUrlForBrowser } = webpackDevServerUtils.prepareUrls(protocol, config.serverHost, config.serverPort);
  opn(localUrlForBrowser);
}
```

Right now you define variables at the top of the block and use them much later.

rafaesc:
Ok I already fix it  :-)

sapegin:
Please merge master and bring back destructuring, we don't need this rule anymore — it was there for Node 4 ;-)

Also, could you please update help function in the styleguidist.js?

bin/styleguidist.js
```diff
@@ -182,6 +182,7 @@ function commandHelp() {
 			'',
 			'    ' + chalk.yellow('--config') + '        Config file path',
 			'    ' + chalk.yellow('--verbose') + '       Print debug information',
+			'    ' + chalk.yellow('--open') + '          Opens the url in default browser',
```
sapegin: Opens → Open (see “print” on the previous line).
url → URL ;-)

Same for the docs.

Issue #922, opened 2018-04-04
Downgrade markdown-to-jsx


stephtr:
At the moment `markdown-to-jsx` has some severe issues rendering HTML within markdown (for example [markdown-to-jsx/#170](https://github.com/probablyup/markdown-to-jsx/issues/170)). Since only some of them will get fixed, one long-term option maybe would be to switch to another parser like markdown-it and inject the components into the generated HTML code.

In the near term it maybe would be useful to keep markdown-to-jsx@^6.4.1 such that one can locally fix it to 6.4.1 if needed.

sapegin:
No, we're not going back to massive Markdown parsers. 

I guess the issues with markdown-to-jsx as soon as someone submits a pull request to fix them. 

But feel free to send a pull request to downgrade markdown-to-jsx.

stephtr:
Ah ok, I didn't know that was an issue. I just had the feeling that for smaller projects like `markdown-to-jsx` it is much more complex/problematic to support the whole Markdown specification than for a specialized parser.

bjmin:
Hi, seems https://github.com/probablyup/markdown-to-jsx/issues/170 had been fixed
should we upgrade `markdown-to-jsx` version?

stephtr:
It [isn't fully fixed](https://github.com/probablyup/markdown-to-jsx/pull/175#issuecomment-387533914) at the moment and since yarn/npm by default install the most recent version, I don't see the need for updating the dependency (at least now).

bjmin:
thanks, it works as you said.

stephtr:
FYI: Since the issue has been fixed (thanks to probablyup), I have submitted a pull request for updating the dependency (#982).

PR #925, opened 2018-04-04
Downgrade `markdown-to-jsx`

stephtr:
As discussed in #922 downgrade the minimal required version of `markdown-to-jsx` such that one can use a version without specific bugs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "loader-utils": "^1.1.0",
     "lodash": "^4.17.5",
     "lowercase-keys": "^1.0.1",
-    "markdown-to-jsx": "^6.6.0",
+    "markdown-to-jsx": "^6.4.1",
     "mini-html-webpack-plugin": "^0.2.3",
     "minimist": "^1.2.0",
     "ora": "^2.0.0",


ahuth:
@stephtr , looks like you're missing changes to the package-lock.json file. If you're on the latest version of npm that should happen automatically. Can you verify which version of npm you used?

stephtr:
Are you sure that `package-lock.json` will change if I loosen the requirements in `package.json`? 6.6.1 is still fulfilling `^6.4.1`, so there would be no need for updating any package or the lock file.
By the way, I just updated my npm to 5.8.0 (the most recent one).

roblevintennis:
Thanks for doing this @stephtr but to actually make this work on the my end, I still had to force the resolution...

image

So above, you notice that the open ended `^` seems to cause the version used by styleguidist: markdown-to-jsx "^6.5.2". And my problems persisted of course. But you can use `resolutions` in the package.json to force version. See https://yarnpkg.com/lang/en/docs/selective-version-resolutions/#toc-how-to-use-it but here's a diff experimenting on the "example" repo here:

image



image


In my case, the issue here is nested divs which seems fixed after above shannanigans. 

image

stephtr:
I think you/it is using an old version of `react-styleguidist`, which gets resolved to v6.5.3, which requires `markdown-to-jsx` v6.5.2 or higher.

roblevintennis:
@stephtr 🤦‍♂️ yah, I realized that late yesterday and forgot to come back to update my comment. Sorry for that inaccuracy! I suppose I'll leave but update the comment for posterity though...some folks my wish to do the resolution thing so they don't have to include the dependency themselves. Thanks for your correction!

stephtr:
No problem! I think that your method is probably still the easiest one for pinning it to 6.4.1

Issue #897, opened 2018-03-27
Style guide for Styleguidist UI


sapegin:
We have quite a lot of UI components in Styleguidist now and sometimes it’s hard to find an existing component and how to use it for some new feature. But we have a great tool exactly for that — Styleguidist itself. Let’s make a style guide for our own components.

* [ ] Setup a style guide
* [ ] Publish to the site

leahzxxz:
I can try to help if no one is working on it?

sapegin:
@leahzxxz Sounds great!

sapegin:
@leahzxxz Nope, it's about adding a style guide for existing components in `src/rsg-components` folder. So adding `styleguide.config.js` into a root folder and pointing it to `src/rsg-components/**/[A-Z]*.js` should be enough. So your style guide config should look like this:

```js
module.exports = {
	components: 'src/rsg-components/**/[A-Z]*.js',
	defaultExample: true,
	webpackConfig: {
		module: {
			rules: [
				{
					test: /\.jsx?$/,
					exclude: /node_modules/,
					loader: 'babel-loader',
				},
				{
					test: /\.css$/,
					loader: 'style-loader!css-loader',
				},
			],
		},
	},
};
```

leahzxxz:
I see - just want to clarify: adding `styleguide.config.js` is enough? Do I need to add any pages or anything else? (sorry still a new dev)

sapegin:
Yep, similar to examples style guides we have (`examples` folder) but to show components that is used in Styleguidist itself.

You may see some components twice (`Component` + `ComponentRenderer`) because of how we structure the code but I would ignore that for now.

And also a new script here:

https://github.com/styleguidist/react-styleguidist/blob/4396dc28f097e7374def088063e2fcf211cd1520/package.json#L143-L145

(You don't need to specify a config path for that case, it will use a default one.)

leahzxxz:
Thanks for explaining - I think I understand better now (was wondering why I added the file but it was still only loading the buttons - now I know it's pointing to the `examples` folder 😅 ). Before I put up the PR, do I need to change anything in `package.json`?

sapegin:
Yup, you need to add a command to run your new style guide like this:

```
"start:ui": "node bin/styleguidist.js server", 
```

leahzxxz:
Thanks! A few follow up questions:
[...]

PR #926, opened 2018-04-05
WIP: Style guide for Styleguidist UI - issue #897

leahzxxz:
WIP for #897


sapegin:
Thanks for the pull request! I'll look into failing tests, I didn't expect that ;-£

sapegin:
You can remove `defaultExample: true` — it will fix errors you see in the browsers console.

leahzxxz:
@sapegin thanks! there was also a React error for getting the same key for an `li` element so i added `Math.random()` to the key, let me know if you prefer another way to do it :)  Thanks for all the instructions!

Regarding the `package-lock` warning, I tried removing and `npm install` again but there's nothing new to commit (I guess it's because I just added a start script and it didn't affect any versioning in `package-lock`). any idea on what to do here? thanks!

sapegin:
> @sapegin thanks! there was also a React error for getting the same key for an li element so i added Math.random() to the key, let me know if you prefer another way to do it :) 

Please don't use random keys in React, this is much worse than using indexes as keys (which can be acceptable in rare cases): you're telling React that you have a new object every render, so React has remove a previous element, create and mount a new one, instead of updating an existing one (in case of changed props) or not doing anything at all. This is never what you want ;-) Could you try `filepath` instead of `name`?

> Regarding the package-lock warning, I tried removing and npm install again but there's nothing new to commit (I guess it's because I just added a start script and it didn't affect any versioning in package-lock).

Ignore that, we need to make the warning more clear ;-) You need to do that if you change any dependencies, which isn't the case here.

leahzxxz:
@sapegin that's really helpful to know - always good to learn best practices in React. Just pushed the change, let me know if you need anything else!

leahzxxz:
@sapegin let me know if there's anything else I can do to get this merged - thanks!

sapegin:
Thank you, merged 🦄

PR #929, opened 2018-04-06
Bump markdown-to-jsx

### LINKED ISSUE #928: Nested spans and divs get a mangled structure
I've got a test case here:
https://github.com/styleguidist/react-styleguidist/pull/927

Here's screen shot of examples/basic
image

Here's in our design system where we discovered:

image

Note, we notice this is only an issue since the last release.

We're happy to try and help, but for all I know this is due to some external parsing library being used. Lmk if you have an idea and/or if this is already slated to be fixed ... thanks!

ahuth:
Resolves #928.

Updates markdown-to-jsx to [version 6.6.1](https://github.com/probablyup/markdown-to-jsx/releases/tag/6.6.1), which has [a bugfix](https://github.com/probablyup/markdown-to-jsx/issues/168) for some rendering issues involving nested divs in markdown.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9098,9 +9098,9 @@
       "integrity": "sha1-Sz3ToTPRUYuO8NvHCb8qG0gkvIw="
     },
     "markdown-to-jsx": {
-      "version": "6.6.0",
-      "resolved": "https://registry.npmjs.org/markdown-to-jsx/-/markdown-to-jsx-6.6.0.tgz",
-      "integrity": "sha1-MR9bezwYzv0GageABTFzDTpisCk=",
+      "version": "6.6.1",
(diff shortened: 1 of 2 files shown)


ahuth:
Oh, just noticed #925, which downgrades the required version of markdown-to-jsx. I believe downgrading would fix my issue as well.

@stephtr, is downgrading still better for you than upgrading to the latest patch version?

stephtr:
I think 6.6.1 mostly [doesn't fix the issue](https://github.com/probablyup/markdown-to-jsx/issues/168#issuecomment-378585486). In my opinion downgrading would have the advantage that `^6.4.1` allows the usage of 6.6.1 (I think npm even installs the most recent (same major) version by default) but also allows the user to fix it to 6.4.1.

ahuth:
Ah, I think you're right. I'm more familiar with how yarn.lock files work, but just tested this and downgrading doesn't update the package-lock.

PR #932, opened 2018-04-07
Add --open argument to open default browser on start

rubenmoya:
Fixes https://github.com/styleguidist/react-styleguidist/issues/918

I had a look at #748 which uses `opn` directly, but I decided to go another path.

I've used `react-dev-utils`  [openBrowser](https://github.com/facebook/create-react-app/blob/next/packages/react-dev-utils/openBrowser.js), this used by `react-scripts` [here](https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/scripts/start.js) to open the browser, it has some cool things like reusing an existing tab, but it also does some things based on the `BROWSER` env variable, so I have mixed feelings about using it and I'd love to hear what you think.

Since I needed the `urls` to open the browser I changed the `printServerInstructions` function to receive the `urls` as an argument, it should be fine since `config.printServerInstructions` still receives the same arguments.

### DIFF
--- a/bin/styleguidist.js
+++ b/bin/styleguidist.js
@@ -7,6 +7,7 @@ const ora = require('ora');
 const stringify = require('q-i').stringify;
 const formatWebpackMessages = require('react-dev-utils/formatWebpackMessages');
 const webpackDevServerUtils = require('react-dev-utils/WebpackDevServerUtils');
+const openBrowser = require('react-dev-utils/openBrowser');
 const logger = require('glogg')('rsg');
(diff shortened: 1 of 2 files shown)


rubenmoya:
Unless I'm missing something, it's updated [here](https://github.com/styleguidist/react-styleguidist/pull/932/files#diff-73ad1b08d17992777e6b20062240bdf0R191) right?

sapegin:
Thanks, I'll try to make a new release soon!

PR #933, opened 2018-04-10
Add custom component display names

glebez:
Hey,

here comes my attempt on allowing users to define custom component display names for SG, that was discussed in #195.

Now, the user can define a custom name via a custom @sgDisplayName tag in the doc block, i.e.:

```
/*
* An example-less button with custom display name.
* @sgDisplayName Push Button 🎉
*/
```

or by utilising newly introduced `updateDocs` config option (#868), i.e.:

```
updateDocs(docs) { 
  if (docs && docs.displayName) { 
    docs.sgDisplayName = _.lowerCase(docs.displayName); 
  } 
  return docs; 
},
``` 

Let me know, please, if this is close to what you meant it to be. 
Also, some updates to the cookbook/docs will follow after initial review.

Cheers,
G.


loaders/utils/getProps.js
```diff
@@ -106,5 +106,13 @@ module.exports = function getProps(doc, filepath) {
 		doc.displayName = getNameFromFilePath(filepath);
 	}
 
+	if (doc.doclets && doc.doclets.sgDisplayName) {
+		doc.sgDisplayName = doc.doclets.sgDisplayName;
+		delete doc.doclets.sgDisplayName;
+		if (doc.tags) {
```
sapegin: What's the difference between `doclets` and `tags`? Worth adding a comment.

glebez: Honestly, I don't know the difference :D I just noticed, that the custom tag gets added both to doclets and tags while parsing, so i decided it's a good idea to remove it from both. I'll investigate a little more on the matter.

glebez:
All right, I've changed the name, updated componenstListRenderer, so it uses visibleName as well and threw in some docs. Also, had to rebase because of conflicts with master.

docs/Cookbook.md
```diff
@@ -479,6 +480,29 @@ See in [configuring webpack](Webpack.md#reusing-your-projects-webpack-config).
 
 See [working with third-party libraries](Thirdparties.md).
 
+## How to change the names of components displayed in Styleguidist UI?
+
+Often you might want to change your components' names to be displayed differently, for example, for stylistic purposes or to give them a more descriptive names in your styleguide.
```
sapegin: Not really often, I'd remove “often” ;-)

docs/Cookbook.md
```diff
@@ -479,6 +480,29 @@ See in [configuring webpack](Webpack.md#reusing-your-projects-webpack-config).
...
+
+This can be done by adding [@visibleName](Documenting.md#-visiblename-custom-tag) tag to your component documentation.
+
+In case you want to change components' names in bulk, for example, based on their current name, you can utilize [updateDocs](Configuration.md#updatedocs) configuration option.
```
sapegin: utilize → use
configuration → config

docs/Cookbook.md
```diff
@@ -479,6 +480,29 @@ See in [configuring webpack](Webpack.md#reusing-your-projects-webpack-config).
...
+
+Here's an example of how this can be achieved:
```
sapegin: You could end the previous paragraph with a “:” and remove this one.

docs/Cookbook.md
```diff
@@ -479,6 +480,29 @@ See in [configuring webpack](Webpack.md#reusing-your-projects-webpack-config).
...
+
+```javascript
+const _ = require('lodash')
+
+module.exports = {
+  updateDocs(docs) {
+    if (docs && docs.displayName) {
+      docs.visibleName = _.lowerCase(docs.displayName)
```
sapegin: `docs.displayName.toLowerCase()`

docs/Documenting.md
```diff
@@ -203,6 +204,23 @@ class Button extends React.Component {
 }
 ```
 
+### @visibleName custom tag
```
sapegin: I think this section should be before “Using JSDoc tags”, similar to “Ignoring props”.

docs/Documenting.md
```diff
@@ -203,6 +204,23 @@ class Button extends React.Component {
...
+
+Apart from most of the jsDocs tags that can be used with RSG, @visibleName is not present in jsDoc specs. This is a custom tag, that we have added to provide users with a way to define names for components to be used in the UI.
```
sapegin: Something like this would be shorter with the same amount of useful information:

> Use `@visibleName` JSDoc tag to define component names that are used in the Styleguidist UI:

(And you could remove the next paragraph too.)

docs/Documenting.md
```diff
@@ -203,6 +204,23 @@ class Button extends React.Component {
...
+```javascript
+/**
+ * The only true button.
+ *
+ * @visibleName The Best Button Ever 🐙
+ */
+class Button extends React.Component {
+```
+
+By defining the @visibleName tag above, you let the Styleguidist know that this name should be used whenever the component is displayed in the UI as 'The Best Button Ever 🐙' instead of it's original 'Button' name.
```
sapegin: Instead of repeating what you say in the first paragraph it would be useful to say that this won't change the actual name of the component that is used in JSX.

docs/Cookbook.md
```diff
@@ -25,6 +25,7 @@
 * [How to add fonts from Google Fonts?](#how-to-add-fonts-from-google-fonts)
 * [How to reuse project’s webpack config?](#how-to-reuse-projects-webpack-config)
 * [How to use React Styleguidist with Redux, Relay or Styled Components?](#how-to-use-react-styleguidist-with-redux-relay-or-styled-components)
+* [How to change the names of components displayed in Styleguidist UI?](#how-to-change-the-names-of-components-displayed-in-styleguidist-ui)
```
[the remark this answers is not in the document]
glebez: Hey, it seems that the rest of the items is not ordered alphabetically as well. Should I order them?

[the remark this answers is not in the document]
sapegin: Would be super cool! But preferably in a separate PR.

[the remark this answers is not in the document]
glebez: 👍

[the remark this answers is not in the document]
glebez: Hey, I finally got to alphabetically ordering the cookbook sections and want to ask: is this really a good idea? In this case, the first entry in the cookbook will be 'Are there any other projects like this?' and it feels that there are more important/useful topics in the cookbook than this... It feels that there can be a benefit in structuring the cookbook, but in some other, more semantic way, like from beginner topics to more advanced stuff or from most frequent requests to more rare/specific.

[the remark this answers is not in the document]
glebez: All right! Nope, the rest of the changes is in place already, so this one is ready to go 🚢

lukashala:
Is it possible to merge this PR please? These changes would be beneficial in our project. Thx.

sapegin:
Looks like I forget to include this PR in 7.1.0, sorry 😭 — will try to make 7.2.0 soon!

PR #934, opened 2018-04-11
bin: Don't report 'On your network' on localhost

jelly:
When explicitly setting the serverHost to 127.0.0.1 or localhost,
urls.lanUrlForTerminal is undefined. It makes no sense to print out
undefined, therefore only show the 'On your network' line when it's
defined.


sapegin:
Thanks, merged!

PR #954, opened 2018-04-25
Export overrides to support easier subclassing and configuration

jspears:
If you want to have custom react components that are rendered via markdown it requires reimplementing the  Markdown class.   By exporting the overrides someone could add there own component 

Example:
```
 import {baseOverrides} from 'react-styleguidist/src/Markdown/Markdown'
 import MySpecialComponent from 'my-special-component';
 baseOverrides.MySpecial = { component:MySpecialComponent }
```

Then later in the a markdown file.
```markdown
I would really like a custom react component
<MySpecial/>
Boom! its there.

```


sapegin:
Thanks!

PR #955, opened 2018-04-25
[DO NOT MERGE] - Mrbuilder

jspears:
So... i spent an hour or two to get this building with mrbuilder -- (Just building not actually working).   It splits up the app into a bunch of packages.   This allows for better documentation, easier reuse and better release/package management.   

If your interested in this route, I'd spend some time getting actually working, but I thought I would check.


sapegin:
Could you explain how is this useful?

jspears:
Sure sorry.

There are some advantages to turning a big project into a bunch of smaller ones and keeping them in the same repository, aka monorepos.   Having separate packaging can make testing, attribution, ownership documenting and versioning a little easier, along with dependency management-- that is it helps identify which components use which libraries.   

However monorepos in JS land have traditionally have had 2 major problems.
1) Keeping versions in sync across all the projects.  For this [Lerna](https://lernajs.io/) does a remarkabley good job.   Makes managing 50 packages in repository just about as easy 1.

2) The second problem is building said packages.   Webpack by itself doesn't do it very well, and all the other things linters, demos ... have been a problem, managing configs, and all that jazz.   [Mrbuilder](https://mr-builder.github.io) (disclaimer I wrote mrbuilder) -- solves the second problem, by wrapping webpack in a declarative json5 format, and allowing for profiles that can be customized per monorepo and also allowing for pretty easy dependency setup.     This  way all the examples, libraries, demos are built in a consistent fashion.  In addition it smooths over webpack upgrades,  by providing a consistent configuration api; which unfortunately webpack does not have. MrBuilder uses react-styleguide and has a plugin for it (pretty raw one) that allows monorepos to become styleguides.  

Anyways, I've been using your project and its pretty great; but thought it might be desirable to make it a monorepo.  As I've, and I assume others, have wanted to use a few of your components, in other projects.   Its a few more hours of work to get everything back to working,  which I don't mind doing, but wanted to see how you felt about a massive PR, to break react-styleguidist up into a bunch packages; before I did that.  

No pressure, a thanks, but no thanks is fine - just thought I'd put it out there.

sapegin:
Thanks for the detailed explanation!

> Anyways, I've been using your project and its pretty great; but thought it might be desirable to make it a monorepo.

Eventually, yes. 

> As I've, and I assume others, have wanted to use a few of your components, in other projects. Its a few more hours of work to get everything back to working, which I don't mind doing, but wanted to see how you felt about a massive PR, to break react-styleguidist up into a bunch packages; before I did that.

That's a good idea in theory. But right now components have too many dependencies on how Styleguidist work. And I don't think putting each component in to a separate package is the way to solve it.

PR #956, opened 2018-04-25
Public Path Configuration

ahuth:
This PR adds a configuration option to set the `publicPath` for Webpack. We need this for serving up a static styleguidist build from a Rails server, and the JS and CSS have to be in specific directories.

Let us know what you think about this, or if there are other solutions. 🤘

### DIFF
--- a/scripts/make-webpack-config.js
+++ b/scripts/make-webpack-config.js
@@ -37,6 +37,7 @@ module.exports = function(config, env) {
 		entry: config.require.concat([path.resolve(sourceDir, 'index')]),
 		output: {
 			path: config.styleguideDir,
+			publicPath: config.publicPath,
 			filename: 'build/[name].bundle.js',
 			chunkFilename: 'build/[name].js',
 		},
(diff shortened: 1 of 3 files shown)


scripts/make-webpack-config.js
```diff
@@ -96,13 +97,13 @@ module.exports = function(config, env) {
...
-					  * to accomodate React's API.
-					  * */
-					reduce_funcs: false
+					/**
+					 * We disable 'reduce_funcs' for keeping uglify from inlining Preact's VNode.
+					 * If enabled, the 'new VNode()' is replaced with a anonymous 'function(){}',
+					 * which is problematic for 'preact-compat', since it extends the VNode prototype
+					 * to accomodate React's API.
+					 * */
+					reduce_funcs: false,
```
ahuth: These changes were from prettier running.

sapegin:
I'm not really sure what problem you're trying to solve here. How is your use case different from [this example](https://react-styleguidist.js.org/examples/basic/), which is also not in the root folder?

ahuth:
The issue is with the paths to the Javascript and CSS for static builds, and a quirk of Rails (which we're using to serve up the static build).

Basically because of how Rails serves up static assets, we need the `<link>` and `<script>` tags to have absolute paths from a specific directory.

We figured that allowing the `publicPath` to be set (and passed through to webpack) could be useful for other folks.

I'd understand, though, if this is an edge case you don't want to support. And either way, thank you for making this project that we've gotten a tremendous amount of value out of!

sapegin:
I think you should use https://react-styleguidist.js.org/docs/configuration.html#dangerouslyupdatewebpackconfig

ahuth:
Yes, you're right. Thanks! I missed that one in the docs.

PR #964, opened 2018-05-04
Update Docs: display source code of any file.

### LINKED ISSUE #952: updateExample setting to include encoding (value.substring is not a function)
While using the [documentation example](https://react-styleguidist.js.org/docs/configuration#updateexample) for the `updateExample` setting I was getting the following error on the UI: `value.substring is not a function` 

While debugging I realized that `fs.readFileSync(filepath)` was returning the Buffer instead of a string. 

I was able to fix it by simply passing an encoding setting:

`content: fs.readFileSync(filepath, 'utf8'),`

I'm wondering if this is a problem only on certain versions of node... I would be happy to make a PR to update the documentation if you think having the encoding setting is the best way to display this example. 

This feature is so valuable that it would be great to have it in the cookbook under something like "how to display the source code of a file?". Also happy to help with that if you like it.

giocodes:
Hi @sapegin, here's the update to the docs that fixes #952

Cheers!

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -8,6 +8,7 @@
 * [How to exclude some components from style guide?](#how-to-exclude-some-components-from-style-guide)
 * [How to hide some components in style guide but make them available in examples?](#how-to-hide-some-components-in-style-guide-but-make-them-available-in-examples)
 * [How to dynamically load other components in an example?](#how-to-dynamically-load-other-components-in-an-example)
+* [How to display the source code of any file?](#how-to-display-the-source-code-of-any-file)
 * [How to set global styles for user components?](#how-to-set-global-styles-for-user-components)
 * [How to add custom JavaScript and CSS or polyfills?](#how-to-add-custom-javascript-and-css-or-polyfills)
 * [How to use React Styleguidist with Preact?](#how-to-use-react-styleguidist-with-preact)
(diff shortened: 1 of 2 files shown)


docs/Cookbook.md
```diff
@@ -119,6 +120,38 @@ const iconElements = Object.keys(icons).map(iconName => {
...
+## How to display the source code of any file?
+
+First, coode examples can receive [props and settings](Documenting.md#usage-examples-and-readme-files):
+
+    ```js { "file": "../mySourceCode.js" }
+    ```
+
+The above example adds a setting called `file` with the **relative path** to the file we want to display as value.
+
+Second, use the [updateExample](Configuration#updateexample) configuration option, to detect the setting and change the content of a fenced code block:
```
sapegin: configuration → config

docs/Cookbook.md
```diff
@@ -119,6 +120,38 @@ const iconElements = Object.keys(icons).map(iconName => {
...
+
+```javascript
+module.exports = {
+  updateExample(props, exampleFilePath) {
+    // props.settings are passed by any fenced code block, in this case
+    const { settings, lang } = props
+    if (typeof settings.file === 'string') {
+      // "../mySourceCode.js"
+      const filepath = path.resolve(exampleFilePath, settings.file) // "absolute path to mySourceCode.js"
```
sapegin: Please move all comment above lines to avoid wrapping.

giocodes:
@sapegin I've applied the feedback. Cheers! 😄

Issue #965, opened 2018-05-04
TypeError: Cannot read property 'name' of undefined


mr-winter:
When defining an example block in a markdown file that's not related to any component, opening the example in isolation mode causes a TypeError. 
This is because there is no component associated retrieve the name from.

```
    main.bundle.js:62574 Uncaught TypeError: Cannot read property 'name' of undefined
            at getPageTitle (main.bundle.js:62574)
            at renderStyleguide (main.bundle.js:39116)
            at render (main.bundle.js:16066)
        getPageTitle @ main.bundle.js:62574
        renderStyleguide @ main.bundle.js:39116
        render @ main.bundle.js:16066
```

To reproduce check out 
https://github.com/mr-winter/example/tree/issue-pagetitle

PR #966, opened 2018-05-04
Fix: Use section title for isolation mode if component is not available  (#965)

mr-winter:
Fixes issue #965 returning the section title, if no components are associated to the section that gets passed into the getPateTitle method.

The issue occurs when writing documentation in markdown with examples that don't directly relate to any component, such as typographic guidelines or colour patches.

### DIFF
--- a/src/utils/getPageTitle.js
+++ b/src/utils/getPageTitle.js
@@ -13,6 +13,10 @@ import { DisplayModes } from '../consts';
  */
 export default function getPageTitle(sections, baseTitle, displayMode) {
 	if (displayMode === DisplayModes.component || displayMode === DisplayModes.example) {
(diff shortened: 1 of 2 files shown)


src/utils/getPageTitle.js
```diff
@@ -13,6 +13,10 @@ import { DisplayModes } from '../consts';
  */
 export default function getPageTitle(sections, baseTitle, displayMode) {
 	if (displayMode === DisplayModes.component || displayMode === DisplayModes.example) {
+		if (!sections[0].components || !sections[0].components.length) {
```
sapegin: Something like this would be much more readable:

```js
// import get from 'lodash/get';
const name = get(sections[0], 'components.0.name', sections[0].name);
return `${name} — ${baseTitle}`;
```

Issue #963, opened 2018-05-03
TypeError: Object.entries is not a function


mealeyst:
This is no longer an issue for me, but I just wanted to make sure that other devs don't run into the same issue that I did. When I was trying to add scripts via the template in the styleguide.config.js file, I kept receiving the following error:

`TypeError: Object.entries is not a function`

This was my configuration:

```
const path = require('path')

module.exports = {
  webpackConfig: require('./build/webpack.config.styleguide.js'),
  components: 'patternLibrary/**/*.{js,jsx,ts,tsx}',
  styleguideComponents: {
    Wrapper: path.join(__dirname, 'patternLibrary/theme')
  },
  template: {
    favicon: '//res.cloudinary.com/roa-canon/image/upload/v1508867160/favicon.ico',
    head: {
      scripts: [
        {
          src: 'https://use.typekit.net/bfw8xpa.js'
        }
      ]
    }
  },
  ignore: ['**/theme/**/*.{js,jsx,ts,tsx}']
}
```

I spent a bunch of time trying to create an example that matched the setup we had for our project's codebase. I managed to get it set up similarly yet it wasn't erring out. I finally just googled the error itself. Turns out the reason why it was running successfully on my example and not on our legacy codebase was so simple and I noticed when I did `node -v` in both codebases.

The project codebase returned the following:
`v6.6.0`
The example codebase I was running returned the following:
`v7.7.1`

I was using ndenv to set versions of node in various project, the version I was running in our project didn't have support for `Object.entries`. Be sure to use a later version of node! Hopefully this helps some other developers if they are running into something similar!

sapegin:
We don't use `Object.entries`, are you sure it's from Styleguidist?

sapegin:
This should fix it: https://github.com/vxna/mini-html-webpack-template/pull/1

PR #972, opened 2018-05-09
Fix: Slot onClick passes correct id

### LINKED ISSUE #971: Slot onClick handler calls previous slot's onClick
There's a small coding bug in `Slot.js` where `props` is reassigned in a loop, so the `onClick` handler of all slots after the first end up delegating back to the first slot's `onClick`.

It's hard to produce a demo for this, it's more future proofing for custom slots. I'll create a PR with a unit test.

kidkuro:
Fixes #971.

Updated `Slot` `onClick` unit test to verify that correct id passed for second slot, and fixed code to make this pass.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -1,10 +1,10 @@
 ---
 name: "\U0001F41B Bug report"
 about: Something isn’t working as expected
-
 ---
 
 **Current behavior**
+
 <!-- A clear and concise description of what the bug is -->
 
 **To reproduce**
@@ -25,4 +25,5 @@ Make sure you’re using the latest version of React Styleguidist.
 -->
 
 **Expected behavior**
+
 <!-- A clear and concise description of what you expected to happen -->
--- a/.github/ISSUE_TEMPLATE/Feature_request.md
+++ b/.github/ISSUE_TEMPLATE/Feature_request.md
@@ -1,19 +1,22 @@
 ---
 name: "\U0001F680 Feature request"
 about: I have a suggestion (and might want to implement myself)
-
 ---
 
 <!-- Consider opening a pull request instead: it’s a more productive way to discuss new features -->
 
 **The problem**
+
 <!-- A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] -->
 
 **Proposed solution**
+
 <!-- A clear and concise description of what you want to happen. Add any considered drawbacks. -->
 
 **Alternative solutions**
+
 <!-- A clear and concise description of any alternative solutions or features you’ve considered. -->
 
 **Additional context**
+
 <!-- Add any other context or screenshots about the feature request here. -->
--- a/.github/ISSUE_TEMPLATE/Question.md
+++ b/.github/ISSUE_TEMPLATE/Question.md
@@ -1,7 +1,6 @@
 ---
 name: "\U0001F914 Support question"
 about: I have a question or don’t know how to do something
-
 ---
 
 --------------^ Click “Preview”!
@@ -10,7 +9,7 @@ If you have a support question, ask it on [Stack Overflow](https://stackoverflow
 
 Before submitting a new question, make sure you:
 
-* Searched existing  [Stack Overflow questions](https://stackoverflow.com/questions/tagged/react-styleguidist).
+* Searched existing [Stack Overflow questions](https://stackoverflow.com/questions/tagged/react-styleguidist).
 * Searched opened and closed [GitHub issues](https://github.com/styleguidist/react-styleguidist/issues?utf8=%E2%9C%93&q=is%3Aissue).
 * Read [the documentation](https://react-styleguidist.js.org/docs/getting-started.html).
 * Googled your question.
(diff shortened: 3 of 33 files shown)


kidkuro:
The failed check doesn't appear to have anything to do with my code change. 

#974, #976 and #978 have all failed for the same reason.

It appears #963 was raised with the same problem.

Since `Object.entries` isn't in the source code, my guess is that could be that Babel is transpiling code to use it, expects a browser environment that supports it, or has been polyfilled with core-js... but is executing in a test in Node on an agent with and older version of Node that [doesn't support it](https://node.green/#ES2017-features-Object-static-methods-Object-entries).

sapegin:
Sound be fixed in master now.

kidkuro:
Awesome, thanks. I'll merge and push up to each PR again.

sapegin:
Looks like I'm missing something, how does this fix the issue? Shouldn't it behave the same way it was before renaming?

kidkuro:
The loop reassigned the `props` argument inside the loop to a new object... but also extracted `const { onClick } = props;` Which means, the second time round the loop `onClick` referred to the first slot's `onClick` not the `onClick` passed to the function. And the first slot's `onClick` is a delegate function that passes the first slot's `id` in a closure.

Try running the test without the code change, you'll see it expects to get `'two'` but receives `'one'`

By the way, I would suggest turning on the eslint rule to prevent reassigning arguments, that will prevent this sort of bug.

sapegin:
Oh, I see!

> By the way, I would suggest turning on the eslint rule to prevent reassigning arguments, that will prevent this sort of bug.

That sounds good, but may need some changes in the code ;-)

sapegin:
Could we try to make this code more explicit?

```js
	const rendered = fills.map((fill, index) => {
		// { id: 'pizza', render: ({ foo }) => <div>{foo}</div> }
		const { id, render } = fill;
		if (id && render) {
			// Render only specified fill
			if (onlyActive && id !== active) {
				return null;
			}

			const { onClick } = props;
			const fillProps = {
				...props,
				name: id,
				// Set active prop to active fill
				active: active && id === active,
				// Pass fill ID to onClick event handler
				// eslint-disable-next-line react/prop-types
				onClick: onClick && ((...attrs) => onClick(id, ...attrs)),
			};

			const Fill = render;
		        return <Fill key={index} {...fillProps} />;
		}

                // fill is a component 
        	const Fill = fill;
		return <Fill key={index} {...props} />;
	});

```

sapegin:
Or probably we should just remove the simple form and require ID for every fill. I think the simple form is used only here now:

https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/slots/index.js#L11

kidkuro:
Sure, I'll update that to have an id

sapegin:
I can increase it even more, but not sure it will help ;-/

sapegin:
It's actually different, this is the flaky one and it's passing:

> ✓ changes the render after code change (1089ms)

And the failing test case is:

> 1) Styleguidist core shows single component in isolated mode:
>     Uncaught Error: Script error.

sapegin:
Which is probably related to your latest changes, and the test looks fine for me:

https://github.com/styleguidist/react-styleguidist/blob/7f5137e2484eb861371ede7e6e8f718e97ff9d06/test/cypress/integration/core_spec.js#L12-L18

sapegin:
Where did you put that `NO_PROXY`? ;-/ We need some docs how to run and debug these tests locally, but they actually right: the isolated mode is broken now because the `name` is overwritten by the fill name. It may be not that easy to fix though…

kidkuro:
In order to access the internet on my work network you must go via a proxy, so I have an `HTTP_PROXY` environment variable set, but I never bothered also setting a `NO_PROXY` environment variable to exclude localhost... since it didn't see to cause me any issues. However, running Cypress wasn't working, so I exported `NO_PROXY=localhost` and that allowed Cypress to run.

However, all the tests passed for me. I will have another go, but if I can't sort it out, I think I should roll back the second commit. That refactoring is unrelated to the original bugfix, and it should probably be left to a separate, more considered PR.

kidkuro:
Ok, I reproduced the error... not sure what happened last time. I have no idea what I am looking at... but... I'll give it a go.

kidkuro:
Ok, so there is a fundamental difference between the tabbutton/tabpanel slots that require an id to associate the pair, which is used as the name, and the Isolate button.

I'm going to revert the second commit. I don't really understand the idea behind the slots, but I have a feeling that it might be too generic a pattern to properly implement both of those purposes. In any case, if it needs to change, it should be a separate ticket.

sapegin:
OK, let's do that, and then I'll try to cherry pick your latest commits and fix that. The code is really confusing now ;-|

kidkuro:
Ha! OK... The changes tab shows the actual change. Its only the very first commit. If you'd like I can just close this PR, create a fresh branch, and a new PR.

kidkuro:
Oh, you've already merged. Cool 👍

sapegin:
Could you open a new PR with the changes you've reverted? That could be a good start and I could help with finishing it.

P. S. You can create a new branch directly at this repo, no need to fork it ;-)

PR #974, opened 2018-05-09
Add editorConfig to Editor

### LINKED ISSUE #973: Allow Editor to be configurable
**The problem**
The `Editor` component can only be configured via context. It would be useful for plugins to also use the `Editor` with different `editorConfig`

**Proposed solution**
Add an `editorConfig` prop to `Editor` that would override settings provided by context, e.g.:

`<Editor code={html} editorConfig={{ readOnly: true, mode: 'text/html' }} />`

I will create a PR with this change in it.

kidkuro:
Fixes #973.

Added new `editorConfig` prop to `Editor` which overrides context so that `Editor` can be used by future plugins.

### DIFF
--- a/src/rsg-components/Editor/Editor.js
+++ b/src/rsg-components/Editor/Editor.js
@@ -28,8 +28,14 @@ export default class Editor extends Component {
 		this.handleChange = debounce(this.handleChange.bind(this), UPDATE_DELAY);
 	}
 
-	shouldComponentUpdate() {
-		return false;
+	shouldComponentUpdate(nextProps) {
+		return !!(this.getEditorConfig(nextProps).readOnly && nextProps.code !== this.props.code);
+	}
+
+	getEditorConfig(props) {
+		const { editorConfig } = props;
+		const contextEditorConfig = this.context.config.editorConfig;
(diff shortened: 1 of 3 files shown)


kidkuro:
This PR also fixes #979

src/rsg-components/Editor/Editor.js
```diff
@@ -27,17 +28,31 @@ export default class Editor extends Component {
...
-	shouldComponentUpdate() {
-		return false;
+	shouldComponentUpdate(nextProps) {
+		return !!(this.getEditorConfig(nextProps).readOnly && nextProps.code !== this.props.code);
+	}
+
+	getEditorConfig(props) {
+		const { editorConfig } = props;
+		const contextEditorConfig = this.context.config.editorConfig;
+		return Object.assign({}, contextEditorConfig, editorConfig);
```
sapegin: You could use object spread here:

```js
return {
   ...this.context.config.editorConfig,
   ...this.props.editorConfig
};
```

kidkuro: Ha, of course, much cleaner. I'll update it.

PR #976, opened 2018-05-09
Split new CodeEvaluator component out from Preview

### LINKED ISSUE #975: Split code evaluation component out from `Preview`
**The problem**
The current `Preview` component does 2 jobs: it compiles and evals a code string to create a component, and it displays that created component. It would be useful to plugins to use just the code evaluator component separately from the `Preview`.

**Proposed solution**
Refactor `Preview` splitting it into `Preview` and `CodeEvaluator`.

kidkuro:
Fixes #975 

Created new `CodeEvaluator` component, by extracting it from `Preview`. This separates the concerns/responsibilities and also exposes the `CodeEvaluator` for use in plugins.


kidkuro:
Hmm... I increased coverage over original code. I'm guessing your coverage standards have increased. I'll have a go of adding more tests.

src/rsg-components/Preview/Preview.js
```diff
@@ -1,34 +1,13 @@
...
-// of this component
-class PreviewComponent extends Component {
-	static propTypes = {
-		component: PropTypes.func.isRequired,
-		initialState: PropTypes.object.isRequired,
-	};
-
-	state = this.props.initialState;
-	setStateBinded = this.setState.bind(this);
+/* eslint-disable no-invalid-this */
```
sapegin: Do we still need this?

src/rsg-components/Preview/Preview.js
```diff
@@ -107,18 +63,13 @@ export default class Preview extends Component {
...
-		const exampleComponent = this.getExampleComponent(example);
 		const wrappedComponent = (
-			<Wrapper onError={this.handleError}>
-				<PreviewComponent component={exampleComponent} initialState={initialState} />
-			</Wrapper>
+			<CodeEvaluator
+				code={code}
+				evalInContext={this.props.evalInContext}
+				onError={this.handleError}
+				compilerConfig={this.context.config.compilerConfig}
```
sapegin: Shouldn't CodeEvaluator read config from context itself, similar to what you did for the Editor?

src/rsg-components/CodeEvaluator/CodeEvaluator.js
```diff
@@ -0,0 +1,94 @@
+import React, { Component } from 'react';
```
sapegin: Do you have any new code here? If not I won't read the whole file ;-)

This is React specific, so we should reflect it in the name: something like `ReactExample` (not the best too but I have no better ideas ;-)

kidkuro: Oh, right... yeah, git isn't being your friend here, this is all moved from `Preview`. 

The only things I changed were:
* I renamed `PreviewComponent` to `InitialStateComponent`
* I passed in `compilerConfig` as a prop instead of reading it off `context` as discussed above.

I'm happy to rename it... I struggled to come up with a decent name to.

src/rsg-components/ReactExample/ReactExample.js
```diff
@@ -0,0 +1,94 @@
...
+/* eslint-disable no-invalid-this, react/no-multi-comp */
+
+const FragmentTag = React.Fragment ? 'React.Fragment' : 'div';
+
+const compileCode = (code, config) => transform(code, config).code;
+const wrapCodeInFragment = code => `<${FragmentTag}>${code}</${FragmentTag}>;`;
+
+// Wrap everything in a React component to leverage the state management
+// of this component
+class InitialStateComponent extends Component {
```
sapegin: Something like `StateHolder` would be more correct: it's not just initial state here.

src/rsg-components/Preview/Preview.js
sapegin: Yeah, true.

src/rsg-components/ReactExample/ReactExample.js
kidkuro: Gotcha, I'll change it.

kidkuro:
Oops! Sorry, used to a precommit hook that runs tests. Will fix this now.

sapegin:
Thanks, merged!

Issue #895, opened 2018-03-26
Investigate flacky Cypress test


sapegin:
```
  1) Single component preview section changes the render after code change:
     CypressError: Timed out retrying: Expected to find content: 'Push Me Harder' within the element: <button.button> but never did.
      at Object.cypressErr (http://localhost:8082/__cypress/runner/cypress_runner.js:67227:11)
      at Object.throwErr (http://localhost:8082/__cypress/runner/cypress_runner.js:67192:18)
      at Object.throwErrByPath (http://localhost:8082/__cypress/runner/cypress_runner.js:67219:17)
      at retry (http://localhost:8082/__cypress/runner/cypress_runner.js:61228:16)
      at http://localhost:8082/__cypress/runner/cypress_runner.js:53731:18
      at tryCatcher (http://localhost:8082/__cypress/runner/cypress_runner.js:7091:23)
      at Promise._settlePromiseFromHandler (http://localhost:8082/__cypress/runner/cypress_runner.js:5113:31)
      at Promise._settlePromise (http://localhost:8082/__cypress/runner/cypress_runner.js:5170:18)
      at Promise._settlePromise0 (http://localhost:8082/__cypress/runner/cypress_runner.js:5215:10)
      at Promise._settlePromises (http://localhost:8082/__cypress/runner/cypress_runner.js:5290:18)
      at Async._drainQueue (http://localhost:8082/__cypress/runner/cypress_runner.js:2023:16)
      at Async._drainQueues (http://localhost:8082/__cypress/runner/cypress_runner.js:2033:10)
      at Async.drainQueues (http://localhost:8082/__cypress/runner/cypress_runner.js:1907:14)
      at <anonymous>
```

sapegin:
Sould be fixed in #1073.

PR #978, opened 2018-05-09
Playground passes evalInContext to Slot

### LINKED ISSUE #977: Pass evalInContext to Slot
**The problem**
In order to support a proposed "show HTML" slot plugin, the `Slot` component will need access to `evalInContext` which is passed to the `Playground` so that it can use the `CodeEvaluator` proposed in #976.

**Proposed solution**
Update `Playground` to pass `evalInContext` to `Slot`

kidkuro:
Fixes #977

`Playground` passes `evalInContext` to `Slot`

### DIFF
--- a/src/rsg-components/Playground/Playground.js
+++ b/src/rsg-components/Playground/Playground.js
@@ -86,7 +86,7 @@ export default class Playground extends Component {
 						name="exampleTabs"
 						active={activeTab}
 						onlyActive
-						props={{ code, onChange: this.handleChange }}
+						props={{ code, onChange: this.handleChange, evalInContext }}
 					/>
 				}
 				toolbar={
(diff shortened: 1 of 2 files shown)


kidkuro:
I am getting 503 errors attempting to run Cyprus locally to try to reproduce the problem. The error seems unlikely to be related to my change, since this simply passes an extra config option which isn't being used. Not sure how to proceed from here.

kidkuro:
The "Single component preview section changes the render after code change" is failing with a timeout. The code waits for 500ms. I just tried manually executing the scenario by typing into the code editor... and it looks like the lag between type and update could easily exceed that timeout in a real browser environment. Could it be that timeout is too short and is fragile? Again... this doesn't appear to be due to my change.

sapegin:
Yeah, we have some issues with this test: https://github.com/styleguidist/react-styleguidist/issues/895

Restarting the CI usually helps.

sapegin:
Thanks!

(I’ve also increased the timeout in that test — hope it would help, couldn't find a way to avoid the timeout unfortunately.)

PR #982, opened 2018-05-13
Update `markdown-to-jsx` dependency

stephtr:
The issue of #922 has been resolved with `markdown-to-jsx@6.6.3` (thanks probablyup), therefore it would make sense to update the dependency.


sapegin:
Thanks, merged!

PR #983, opened 2018-05-13
Fix: repair user alias rsg-components when provided

kopax:
Current code doesn't permit to alias `rsg-components`.

Only way of doing it is using `config.styleguideComponents`.

This allow user to set an alias.

### DIFF
--- a/scripts/make-webpack-config.js
+++ b/scripts/make-webpack-config.js
@@ -141,9 +141,10 @@ module.exports = function(config, env) {
 	}
 
 	// Add components folder alias at the end so users can override our components to customize the style guide
-	// (their aliases should be before this one)
-	webpackConfig.resolve.alias['rsg-components'] = path.resolve(sourceDir, 'rsg-components');
-
+	// (their aliases should be before this one
+	if (!webpackConfig.resolve.alias['rsg-components']) {
+  		webpackConfig.resolve.alias['rsg-components'] = path.resolve(sourceDir, 'rsg-components');
+	}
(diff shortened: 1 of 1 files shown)


kopax:
Can we include this in futur release?

sapegin:
What problem you're trying to solve? It's not supposed to be changed via `webpackConfig` because it may break Styleguidist. You could change it with `dangerouslyUpdateWebpackConfig`.

kopax:
Fw https://github.com/styleguidist/react-styleguidist/issues/984 @sapegin we have hit this while trying to use the alias. I wonder why we should keep such a restriction, the `:fix` could be accepted as a `:feature` if not forbidden. Otherwise, this should be clearly stated in the documentation as a restriaction. 

> It's not supposed to be changed via `webpackConfig` because it may break Styleguidist. 

Could you elaborate or give user case where it break styleguidist?

> You could change it with `dangerouslyUpdateWebpackConfig`.

Thanks, I am not reading method starting with `dangerously` but now I did read it : https://github.com/styleguidist/react-styleguidist/blob/master/docs/Configuration.md#dangerouslyupdatewebpackconfig

I'd like if possible to know when it can break the react-styleguidist.

sapegin:
I still don't see what problem you're trying to solve, that can't be solved with `styleguideComponents` or `dangerouslyUpdateWebpackConfig` optons. Any changes to webpack config required by Styleguidist itself are potentially dangerous and not allowed.

Issue #591, opened 2017-08-24
Support for font-awesome icons


sabhyankarxactly:
How do I get the font-awesome library to load in the styleguidist? I tried the method for loading external CSS (see below) but it does not load the font-awesome icons.

- Your style guide config (`styleguide.config.js`).
```js
const path = require('path');

module.exports = {
    components: 'src/components/**/index.js',
    showUsage: true,
    require: [
        path.join(__dirname, 'node_modules/font-awesome/css/font-awesome.css')
    ]
};
```
- Webpack version.
2.2.1
- Operating system, browser and version if it’s a layout bug.
Windows 10, Chrome v60

- Reproducible demo:
I can add a reproducible demo if required.

n1313:
I bet you need to require the font file itself, too.

By the way, if your components depend on font-awesome, shouldn't it be the dependency of your package and be required from your components?

sapegin:
Please use fenced blocks with language tags (for example ```` ```js````) to make code examples more readable.

dbertella:
You can use an html template and include the link in the html as I was asking in another issue (#593) or as I did previously, include the whole css, I was using the scss version and include it as you are doing into the `require`. If you are doing that, and use the scss version use this variable `$fa-font-path:        "//netdna.bootstrapcdn.com/font-awesome/4.7.0/fonts" !default; // for referencing Bootstrap CDN font files directly` because it will load the fonts from the cdn and you don't have to import them

sabhyankarxactly:
@dbertella Thanks for the pointer, I was able to use the custom template to load font-awesome icons.
@n1313 I tried loading the font-awesome CSS and font files in the stylguide.config.js require section but did not work. I do like the idea of making font-awesome a dependency of my package and load them through my components, will work on that.

PR #985, opened 2018-05-15
Update Cookbook.md to add comment about loading 3rd party CSS

bugzpodder:
Inspired by https://github.com/SpareBank1/designsystem/blob/master/styleguide.config.js#L93

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -201,6 +201,8 @@ module.exports = {
 }
 ```
 
+You may also want to use MiniCssExtractPlugin.loader in conjunction when loading 3rd-party CSS.
+
 ## How to use React Styleguidist with Preact?
 
 You need to alias `react` and `react-dom` to `preact-compat`:
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -201,7 +201,7 @@ module.exports = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/Cookbook.md
```diff
@@ -201,6 +201,8 @@ module.exports = {
 }
 ```
 
+You may also want to use MiniCssExtractPlugin.loader in conjunction when loading 3rd-party CSS.
```
sapegin: This sentence doesn't say anything without an example.

sapegin: And the main question: why would you want to use it in a style guide?

bugzpodder:
For example, if I want to load css for font-awesome from node_modules.
Putting it in require section as described does not work as it is not automatically loaded on every page.
https://github.com/styleguidist/react-styleguidist/issues/591
and the other example I referenced have their use cases for this as well, ie:
https://github.com/SpareBank1/designsystem/blob/master/styleguide.config.js#L93

bugzpodder:
ie this is basically adding <link rel="stylesheet" href="path/to/local/style.css" /> as adding the plugin will include the css as part of the build.

sapegin:
CSS extraction an CSS loading are completely independent things, you don't need to extract CSS to a file to do anything, you've described.

sapegin:
I advice you to read what MiniCssExtractPlugin actually does and how CSS imports work in webpack instead of complaining here.

PR #989, opened 2018-05-17
Feat: Added exampleModes and usageModes settings per sections

rafaesc:


loaders/utils/getSections.js
```diff
@@ -51,6 +51,14 @@ function processSection(section, config) {
 
 	return {
 		name: section.name,
+		exampleModes:
+			section.exampleModes || config.showCode === undefined
```
sapegin: I think we can use `process` method in the config schema to support deprecated values, and keep them in the schema only. Will be easier to remove later.

docs/Configuration.md
```diff
@@ -15,6 +15,7 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 * [`dangerouslyUpdateWebpackConfig`](#dangerouslyupdatewebpackconfig)
 * [`defaultExample`](#defaultexample)
 * [`editorConfig`](#editorconfig)
+* [`exampleModes`](#examplemodes)
```
sapegin: Shouldn't it be singular — `exampleMode`?

viljamis:
Nice work! @sapegin what do you think it’d require to get this feature out?

sapegin:
It looks good to me! Thanks for all the hard work! I think we can include this in 7.1.0, I’ll try to find some time to make a release.

Issue #919, opened 2018-04-02
Scope according to sections


celiolatorraca:
Hi all!

First of all, thanks for this incredible lib! It is really making my life easier since I'm managing a UI toolkit for my company.

But, here is my question... Is there a way to scope the navigation according to sections?

My problem is that I have some components that have the same name (like a `Button` block component).

All my components are composed of a `set` of some `block` components.
Something like this:
```
├── components
│   ├── playkit-programs-list
│   │   └── src
│   │       ├── blocks
│   │       │   ├── button.js
│   │       │   ├── index.js
│   │       │   ├── map-programs-data.js
│   │       │   └── with-programs-data.js
│   │       ├── index.js
│   │       ├── programs-list.css
│   │       └── programs-list.js
├── libs
│   ├── playkit-buttons
│   │   └── src
│   │       ├── button.css
│   │       ├── button.js
│   │       ├── index.js
│   │       ├── outline-button.css
│   │       └── outline-button.js

```

But, when I click on `Button` link to see its documentation, `react-styleguide` filters components with the exact same name, but not scoped by the sections we build...

Do we have a way to get around it?

PS: I had to override the `ComponentsListRenderer` because it is still using the `name` as `key`, instead of `filepath` (as we can see here https://github.com/styleguidist/react-styleguidist/issues/33, I think you just changed the `ComponentRenderer` key prop isn't it?).

sapegin:
I'm not sure there's an easy way to solve this. Right now all components are global (which isn't a good thing, a would be nice to fix that — #325).

As a workaround (though ugly) you may try to `require` the right component in the example code.

> PS: I had to override the ComponentsListRenderer because it is still using the name as key, instead of filepath (as we can see here #33, I think you just changed the ComponentRenderer key prop isn't it?).

According to blame it uses `name` for over a year, if it was ever different ;-) Feel free to send a PR to fix this.

celiolatorraca:
Thanks for you help @sapegin !!

> I'm not sure there's an easy way to solve this. Right now all components are global (which isn't a good thing, a would be nice to fix that — #325).

Indeed that is the problem!
For now, I did a workaround scopping all my `blocks` with the package name...
Something like:

```js
PlaykitMediaCover_Button
PlaykitVideoStage_Button
```

But with that ugly workaround, I have to require my own components in examples as you sad before...

> As a workaround (though ugly) you may try to require the right component in the example code.

Di you know if there is a way to use ES6 in example?

Thanks!

[...]

Issue #325, opened 2017-02-14
RSG puts components into global namespace


n1313:
All components that are loaded into the styleguide are put into the `global` namespace, so I think it is possible to accidentally overwrite properties of `window` with an unfortunately named component.

sapegin:
The good solution would be to extend code that we use for the `context` option. But it’d be tricky ;-)

n1313:
For historical purposes [here's](https://github.com/styleguidist/react-styleguidist/commit/d07e825e00720f71cf4aea348f1bb12b9dab58b2) my initial stab at this issue. It works but is a bit too dirty of a hack.

rcline:
@n1313 That really doesn't look that hacky to me.  Is there any side effects I am not seeing?

I would love to see this get added!  (Currently having to rename my `Image` component so that it doesn't conflict with `window.Image` :( )

sapegin:
@rcline Doesn’t look super hacky to me now, and I’m sure we can avoid a global variable here. Feel free to send a pull request for that ;-)

rcline:
@sapegin It looks like this code has been changed/moved.
https://github.com/styleguidist/react-styleguidist/blob/d07e825e00720f71cf4aea348f1bb12b9dab58b2/src/rsg-components/Preview/Preview.js#L54-L56

Any direction on where to look to make a similar change? (passing in namespaced components to a `with ()` context call)

sapegin:
Not sure which code has been moved where ;-) 

I think the right place to do it is here: 

https://github.com/styleguidist/react-styleguidist/blob/916e173ecfeb29d89085ed0eddfbefc9f49afb7a/loaders/examples-loader.js#L22-L23

But it’ll require some experiments, you’ll need to move props-loader call there I think:

https://github.com/styleguidist/react-styleguidist/blob/fc6a3ba0a0167de92a8c0c2a7e16d2ae7a3d190a/loaders/utils/processComponent.js#L35

kopax:
I am also having that issues, I'm not sure if I understand the fix, did anybody tried @sapegin solution?

linde12:
Yup. This makes things very difficult when your library has dependencies to third party code. E.g. in my specific case i have a dependency to a component which in turn depends on a component named `Object` resulting in React saying "Object.defineProperty is not a function"

ZxBing0066:
Is there an option to simply ignore components expose to the global? like this below add a `doNotExposeToGlobal` option to section?

```js
[...]

PR #998, opened 2018-05-30
Fix name conflicts by using slugs for routing in isolated mode

jonatansberg:
Hello @sapegin and all of you! 
We're using styleguidist as a part of a big client project, and alost immediately ran in to problems with that the components all live in a global namespace, mentioned in #325, #919 and other places. 

Requiring the specific components that you need is a bit ugly but works to solve naming conflicts in the examples. I'd like to take a stab at fixing this as well at some point, but we had a more critical problem with the routing.
 
When using styleguidist with the `pagePerSection` option set to true the routing use the section or component name instead of it's slug as a path/hash, meaning sections with conflicting names are completely unreachable.

Since the slugs we're already in place to (I assume) handle this very problem when you don't use the `pagePerSection` option using them here as well seemed like a reasonable solution. 

The slug generation could possibly be improved further by prefixing child-section slugs with the parent section, allowing for for more meaningful and stable urls (e.g. `button/button` and `toolbar/button` instead of `button` and `button-1) but this might be a breaking change for some.

### DIFF
--- a/examples/sections/src/components/Button/Readme.md
+++ b/examples/sections/src/components/Button/Readme.md
@@ -1,13 +1,15 @@
 Basic button:
 
 ```jsx
-<Button>Push Me</Button>
+const Button = require('./Button').default
+;<Button>Push Me</Button>
 ```
 
 Big pink button:
 
 ```jsx
-<Button size="large" color="deeppink">
+const Button = require('./Button').default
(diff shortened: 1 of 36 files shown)


wachunga:
It would be good to fix these conflicts. Or at least warn when they happen...

PR #999, opened 2018-05-30
Fix: serverPort can't be string

kopax:
I am sorry for that, I was about to sleep and noticed my mistake :)


sapegin:
Thanks, merged!

Issue #1003, opened 2018-06-01
Bug IE11, Object doesn't support property or method 'find'


rafaesc:
**Current behavior**

Error when you isolate a section in IE11

**To reproduce**

1. run `npm run start:sections`
2. click on an isolated section in IE

it shows on console
`Object doesn't support property or method 'find'`
of this file 
https://github.com/styleguidist/react-styleguidist/blob/master/src/utils/findSection.js#L8-L10

**Expected behavior**

It should work

aabenoja:
It's like a game of wack-a-mole, since I last fixed this in #912

rafaesc:
Interesting, thanks @aabenoja 
I have to update the PR and to use lodash

PR #1004, opened 2018-06-01
Fix: findSection for ie11 #1003

rafaesc:
Issue: https://github.com/styleguidist/react-styleguidist/issues/1003

In addition I had to change this line, because also it appeared other issue on ie11

https://github.com/styleguidist/react-styleguidist/blob/c7a68c99f3b0ee81c62dbd4fcc32b8d4cf74ba37/src/utils/findSection.js#L14-L15

When the file was compiled by babel, it added `Simbol[iterator]` that generated the other issue

image

### DIFF
--- a/src/utils/findSection.js
+++ b/src/utils/findSection.js
@@ -9,16 +9,21 @@ import find from 'lodash/find';
  */
 export default function findSection(sections, name) {
 	// We're using Lodash because IE11 doesn't support Array.find.
-	let found = find(sections, { name });
+	const found = find(sections, { name });
 	if (found) {
 		return found;
 	}
 
-	sections.forEach(section => {
-		if (section.sections && section.sections.length && !found) {
-			found = findSection(section.sections, name) || found;
+	for (let i = 0; i < sections.length; i++) {
+		const section = sections[i];
+		if (!section.sections || section.sections.length === 0) {
+			continue;
 		}
-	});
+		const found = findSection(section.sections, name);
+		if (found) {
+			return found;
+		}
+	}
 
-	return found;
+	return undefined;
 }
--- a/src/utils/findSection.js
+++ b/src/utils/findSection.js
@@ -1,5 +1,3 @@
-import find from 'lodash/find';
-
 /**
  * Recursively finds a section with a given name (exact match)
  *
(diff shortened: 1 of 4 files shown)


src/utils/findSection.js
```diff
@@ -6,12 +8,13 @@
  * @return {object}
  */
 export default function findSection(sections, name) {
-	const found = sections.find(section => section.name === name);
+	const found = find(sections, { name });
```
sapegin: Please add a comment that we're using Lodash because IE11 doesn't support Array.find.

src/utils/findSection.js
```diff
@@ -6,12 +8,13 @@
...
 	if (found) {
 		return found;
 	}
 
-	for (const section of sections) {
+	for (let i = 0; i < sections.length; i++) {
```
sapegin: Can we use Array.map here?

aabenoja: I don't think `map()` is appropriate here when we don't need to transform the collection.

sapegin: Then `_.find`, anything that's not a loop.

src/utils/findSection.js
```diff
@@ -6,20 +6,14 @@
...
-	for (const section of sections) {
-		if (!section.sections || section.sections.length === 0) {
-			continue;
-		}
-		const found = findSection(section.sections, name);
-		if (found) {
-			return found;
+	let found;
+	sections.forEach(section => {
+		if (section.name === name) {
```
aabenoja: I like this approach much better since we only iterate over the whole collection once instead of twice. If we can gracefully break out of this `forEach` once a match has been found that would be even better.

sapegin: I think it was actually better to loop twice, because the first one only loops the first level, and the second one is recursive. 

And we can use `_.find` to avoid unnecessary cycles and an ugly `found` variable ;-)

src/utils/findSection.js
```diff
@@ -6,20 +8,17 @@
...
 	}
 
-	for (const section of sections) {
-		if (!section.sections || section.sections.length === 0) {
-			continue;
-		}
-		const found = findSection(section.sections, name);
-		if (found) {
-			return found;
+	sections.forEach(section => {
```
rafaesc: I had to use `forEach`, because [find](https://lodash.com/docs/4.17.5#find), doesn't support deep level.

rafaesc: jeje sorry,  yes how you said that we can use [_.find](https://lodash.com/docs/4.17.5#find) to avoid unnecessary cycles and an ugly found variable
I understood that we don't need the `found` variable, but I am not sure that there is a way to avoid to use found variable because I would like to use `find` to search deep in the array `section.sections`, but it isn't possible, so for that reason, I used a `forEach` to make a recursivity.

sapegin: Yeah, looks like we have to use a loop here (and that's why it was a loop in the first place): iterating over the whole tree isn't a good idea ;-/

sapegin:
Thanks, merged!

PR #1005, opened 2018-06-01
Use latest ecmaVersion parsable by acorn

jquense:
I ran into an issue trying to run `async`/`await` code in an example. I'm only targeting new Chrome and FF for the styleguide so i've mostly disabled the buble transforms. The problem tho is that this file here, also tries to parse the code, which fails silently.

I think this should be a safe change, and backwards compatible

### DIFF
--- a/src/utils/splitExampleCode.js
+++ b/src/utils/splitExampleCode.js
@@ -21,7 +21,7 @@ const unsemicolon = s => s.replace(/;\s*$/, '');
 export default function splitExampleCode(code) {
 	let ast;
 	try {
-		ast = acorn.parse(code, { ecmaVersion: 10 });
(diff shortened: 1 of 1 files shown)


src/utils/splitExampleCode.js
```diff
@@ -21,7 +21,7 @@ const unsemicolon = s => s.replace(/;\s*$/, '');
 export default function splitExampleCode(code) {
 	let ast;
 	try {
-		ast = acorn.parse(code);
+		ast = acorn.parse(code, { ecmaVersion: 10 });
```
sapegin: Can we use `2019` as more common name?

jquense: Sure, I didn't notice that you could use years as an alias for the option

sapegin:
Thanks, merged!

PR #1007, opened 2018-06-03
Add Maintainter guide

sapegin:
Feel free to comment and ask more questions if anything isn't clear.

### DIFF
--- a/docs/Maintenance.md
+++ b/docs/Maintenance.md
@@ -1,0 +1,77 @@
+# Maintainer guide
+
+<!-- To update run: npx markdown-toc -i docs/Maintenance.md -->
+
+<!-- toc -->
+
+* [We need you!](#we-need-you)
+* [Commit message conventions](#commit-message-conventions)
+* [Pull requests](#pull-requests)
+* [Releases](#releases)
+  * [Patch releases](#patch-releases)
+  * [Minor and major releases](#minor-and-major-releases)
+* [Change logs](#change-logs)
+  * [What is a good change log](#what-is-a-good-change-log)
+  * [Preparing a change log](#preparing-a-change-log)
+
+<!-- tocstop -->
+
+_See also [Developer guide](Development.md)._
+
+## We need you!
+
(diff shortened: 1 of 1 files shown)


docs/Maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+
+<!-- tocstop -->
+
+_See also [Developer guide](Development.md)._
+
+## We need you!
+
+Help us develop and maintain Styleguidist:
+
+* Answer questions in [GitHub issues](https://github.com/styleguidist/react-styleguidist/issues) and [Stack Overflow](https://stackoverflow.com/questions/tagged/react-styleguidist).
```
bluetidepro: Should we also add the Gitter chat as a place to answer questions in?

docs/Maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+* Review [pull requests](https://github.com/styleguidist/react-styleguidist/pulls).
+* Fix bugs and add new features.
+* Write articles and talk about Styleguidist on conferences and meetups (we’re always happy to review your texts and slides).
+
+## Commit message conventions
+
+[Commit message conventions](https://github.com/tamiadev/semantic-release-tamia/blob/master/Convention.md) are based on Angular conventions.
```
bluetidepro: Is there any tools that allow us to like lint this or validate it on an automation level? May make it easier to enforce? Or maybe a "precommit" hook?

docs/Maintenance.md
sapegin: I’m actually thinking about removing all the Gitter links — changes to have an answer there are very small, and questions are usually of very low quality, so you they can't have an answer without asking 10 more questions ;-)

docs/Maintenance.md
sapegin: We don't care what contributors write in commit messages, or how many they add to a pull request ;-) So the only way to make it actually works is to do squash-merge and rewrite messages.

PR #1009, opened 2018-06-04
Feat: added ignoreNodeModules option

mpachin:
Right now there is no option in styleguidists config which will allow hot-reload work on specific node_modules.
Projects directory tree may look like this:
```
project
│
└─── node_modules
│   │
│   └─── ...
│   
└─── src
│   │
│   └─── node_modules
│   │   │
│   │   └─── @subproject1
│   │   │   │
│   │   │   └─── ...
│   │   │
│   │   └─── @subproject2
│   │   │   │
│   │   │   └─── ...
...
```
`@subproject`'s directories hold components which I want styleguidist to keep track of.

With `ignoreNodeModules` now we can specify which exact directory styleguidists hmr should ignore.

### DIFF
--- a/scripts/create-server.js
+++ b/scripts/create-server.js
@@ -13,7 +13,7 @@ module.exports = function createServer(config, env) {
 			hot: true,
 			quiet: true,
 			watchOptions: {
-				ignored: /node_modules/,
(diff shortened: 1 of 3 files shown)


sapegin:
You can do this with `devServer` webpack config option.

PR #1011, opened 2018-06-09
feat(cli): added --schema cli option

### LINKED ISSUE #1010: [Feature Request] Hability to extend schema configuration
**The problem**

react-stylegudiist provide a way for using aliases to update rsg-components; but when it comes to extend the features, it becomes impossible to add configuration due to the way the sanitize fonction and webpack are linked together.

**Proposed solution**

Adding a new cli arguments: https://react-styleguidist.js.org/docs/cli.html

`--schemas` : it will contain the source to use to link the schema to be passed to getConfig

kopax:
Fix #1010.

We need this and I'll appreciate if you could directly release a new react-styleguidist version when you'll accept this.

For the record, It will help us to create layouts, customize configuration per layout while using react-styleguidist schema validation system. 

Thanks in advance and best.


sapegin:
Could you show an example of how are you going to use this? It's likely not the best solution, and I'd like to avoid it.

kopax:
@sapegin, we are trying to optimize the rendering (all in one page) so we wanted to prepare new layouts (like with react-router).

We wanted to add to the option a 'layout' key to select the proper layout.

Because the layout might have a different menus and feature, they should declare their own schema extension. 

At first I wanted to import the main schema then extend it, or turn it to function so it can be extended for each layout..

Then we noticed this was all done before webpack. and it was not in `rsg-component` directory, it was hard to override.

Right now we had to move to a fork because of this.

sapegin:
I don't think it's the right solution for the problem. Plugins would be much better and more flexible, and for your use case you need a very simple implementation: real plugins list from a style guide config and make it available for Styleguidist components. There's a draft implementation #505 but it tries to do much more than that (probably that's why it was never finished), but we could start from small things.

kopax:
@sapegin I am not sure if the module is not existing yet how you would want it coming in execution in your code.

We need to have layout available asap, perhaps I can use the API and you'll give me the API of the load-schema module and I can try to implement just my use case. 

Then somebody has to take care. Elsewhise , is there a way to turn off the schema validation ? I don't see any hack that does't require to fork our own react-styleguidist, do you have one?

kopax:
@sapegin, maybe a `config` that would allow us to disable the schema validation is that something you are willing to accept?

sapegin:
This is exactly the same as your first "solution”, I can't accept it. As I've said before, I see plugins as the only solution now.

sapegin:
This is exactly the same as your first "solution”, I can't accept it. As I've said before, I see plugins as the only solution now.

kopax:
My first solution was to use a `--schema` to allow user to change the schema.

My last option was to use a `schemaCheck: false` to skip schema validation.

I take not of the urgency of the plugins and will try my best to help.

PR #1012, opened 2018-06-09
Fixed missed argument in updateDocs config method

xomyaq:
Example code from `updateDocs` option is broken because of missing `file` argument


sapegin:
Thanks!

Issue #915, opened 2018-04-01
Drop webpack 3 support


sapegin:
Looks like there's no way to support both at the same time and avoid deprecation warning on both versions. Let's wait for [webpack 4 support](https://github.com/facebook/create-react-app/pull/4077) in Create React App, and drop webpack 3 support in Styleguidist.

- [x] Remove conditions from [make-webpack-config.js](https://github.com/styleguidist/react-styleguidist/blob/master/scripts/make-webpack-config.js)
- [x] Update [min webpack version](https://github.com/styleguidist/react-styleguidist/blob/dfcb73de01d2b3ecf0864a95d358b216dcf65618/scripts/utils/ensureWebpack.js#L10)
- [x] Remove webpack 3 env from [Travis CI config](https://github.com/styleguidist/react-styleguidist/blob/dfcb73de01d2b3ecf0864a95d358b216dcf65618/.travis.yml#L8)
- [ ] Remove webpack conditions introduced in #1148

nicoffee:
Hello!
I'l be happy to do it.

jklemm:
Is this PR related to this warnings?

```
warning "react-styleguidist > react-docgen-displayname-handler@2.1.0" has incorrect peer dependency "recast@0.12.6 || 0.13.0".

warning "react-styleguidist > webpack-dev-server@2.11.2" has incorrect peer dependency "webpack@^2.2.0 || ^3.0.0".

warning "react-styleguidist > webpack-dev-server > webpack-dev-middleware@1.12.2" has incorrect peer dependency "webpack@^1.0.0 || ^2.0.0 || ^3.0.0".
```

I'm using `webpack==4.17.2`, `webpack-dev-server==3.1.9`.

SavePointSam:
Create React App 2.0 has been released with Webpack 4 support.

AndreyGladkov:
Remove webpack-dev-server warning https://github.com/styleguidist/react-styleguidist/pull/1180

PR #1023, opened 2018-06-24
chore(*): drop webpack3 support

stepancar:
https://github.com/styleguidist/react-styleguidist/issues/915


stepancar:
@sapegin we should mock copy-webpack-plugin to check parameters passed to it constructor.
I tried mock using jest 22 and faced with problems. [This example]( https://facebook.github.io/jest/docs/en/es6-class-mocks.html#the-4-ways-to-create-an-es6-class-mock) not working in jest 22. Its reason why i switched to jest 23.

sapegin:
Thanks! This is awesome! Though we have to wait for CRA2 final release.

Jest 23 is awesome too! Could you please send it as a separate PR? We could merge it right away.

stepancar:
@sapegin we need update package-lock. Now master crashes after update package-lock
Look at https://github.com/styleguidist/react-styleguidist/pull/1030

After it will be closed - I will create Jest23 migration PR

stepancar:
@sapegin , https://github.com/styleguidist/react-styleguidist/pull/1034

PR #1027, opened 2018-06-26
feat(StyleGuideRenderer): support StyleGuideRenderer composition

cdaringe:
# problem

i desire, but ATM cannot have:

1) free maintenance and updates to RSG components as they are developed, **plus**
2) the ability to tweak RSG components slightly (beyond the theme APIs)

# solution

- improve RSG component composition by proxying unhandled props to the target DOM node
    - use `transform-object-rest-spread` to do this.  woo hoo!
    - ideally it would be nice to extend this pattern to other components, but i thought a small focused change would help the discussion along

# example

i want to apply a class name and/or an id to the `<StyleGuideRenderer />` 
```jsx
import * as React from 'react'
import OrigStyleGuideRenderer from 'react-styleguidist/lib/rsg-components/StyleGuide/StyleGuideRenderer.js'
export default function StyleGuideRenderer (props) {
  return <OrigStyleGuideRenderer {...{ ...props, id: 'bananas', className: 'sans-serif' }} />
}
```

# discussion

- the `contributing.md` steps don't work for me, even on master.  `jest` craps out with:

```
react-styleguidist/test/jestsetup.js:15
    import { configure, shallow, render, mount } from 'enzyme';
           ^

    SyntaxError: Unexpected token {

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:316:17)
```

### DIFF
--- a/src/rsg-components/StyleGuide/StyleGuideRenderer.js
+++ b/src/rsg-components/StyleGuide/StyleGuideRenderer.js
@@ -55,9 +55,10 @@ const styles = ({ color, fontFamily, fontSize, sidebarWidth, mq, space, maxWidth
 	},
 });
 
-export function StyleGuideRenderer({ classes, title, homepageUrl, children, toc, hasSidebar }) {
+export function StyleGuideRenderer(props) {
+	const { className, classes, title, homepageUrl, children, toc, hasSidebar, ...rest } = props;
 	return (
-		<div className={cx(classes.root, hasSidebar && classes.hasSidebar)}>
+		<div className={cx(classes.root, hasSidebar && classes.hasSidebar, className)} {...rest}>
 			<main className={classes.content}>
 				{children}
 				<footer className={classes.footer}>
@@ -79,6 +80,7 @@ export function StyleGuideRenderer({ classes, title, homepageUrl, children, toc,
 
 StyleGuideRenderer.propTypes = {
 	classes: PropTypes.object.isRequired,
+	className: PropTypes.string,
 	title: PropTypes.string.isRequired,
 	homepageUrl: PropTypes.string.isRequired,
 	children: PropTypes.node.isRequired,


sapegin:
Can we make it more explicit?

```js
export default function StyleGuideRenderer (props) {
  return <OrigStyleGuideRenderer { ...props}  containerProps={id: 'bananas', className: 'sans-serif' }} />
}
```

cdaringe:
Sure, that's definitely an option.  I'm not in favor of it personally.  I'm not sure if there are any significant advantages to isolating those props, only to shim them right back into the root node.  In fact, it makes it harder to distinguish between handled and unhandled props.  It's also an extra API to maintain.  Passing on the unhandled props gives a great user experience, IMHO, and adds ~nil complexity.  If I do `<Thing id="best_id" />`, it feels like working with the DOM directly, and has natural imference.

sapegin:
There are quite a lot of advantages:

* You may have more than on HTML element that you want to modify.
* You won't accidentally leak undesired props to DOM (resting all props is dangerous).
* The API is much more clear.

cdaringe:
I respect these points, but disagree with them for the following reasons:

> You may have more than one HTML element that you want to modify.

If that is necessary, than that other element should also be exposed.  SUIR, for example, often hangs related elements off of a root element for this reason. in [this  example](https://react.semantic-ui.com/collections/form), you can see they have things like `<Form.Button />` & `<Form.XYZ />` to use instead of the defaults, if you want to override internal components with other props/goodies.

It's also worth noting that SUIR use a similar pattern to that which I've proposed, but do so via a babel plugin [ref1](https://github.com/Semantic-Org/Semantic-UI-React/blob/c8fabc7805ad23eeeaed9ec37c128ca4d508a3aa/package.json#L92) [ref2](https://www.npmjs.com/package/babel-plugin-transform-react-handled-props).

This plugin based approach is a more advanced solution, which is probably also worth considering.  In the PR'd case, this component is a functional stateless component, so the benefits of the plugin approach are less, but still valid--especially if we have `Component`/`PureComponent`s in RSGs component tree!

> You won't accidentally leak undesired props to DOM (resting all props is dangerous).

I respectfully don't think that this carries much weight.  If one has leaked a prop, he/she has done it upstream, and the bug still exists.

> The API is much more clear.

The API IMHO is much less clear because it's an entirely new API to find and use.  By making it a separate API, you can't just add props to your node, like you would naturally with any other HTML tag.


Anyway, thanks for having this discussion.  I'm not trying to be stubborn/ill-willed/etc, just advocating.  I'd still be happy to use `containerProps` over _nothing_.

cdaringe:
bump (to suppress stale bot)

Issue #1026, opened 2018-06-25
Deprecated Hover for isDeprecated


roblevintennis:
We noticed that when a component heading title (in SectionHeading) is marked @deprecated it gets a strikethrough, but when you hover over it, it switches to underline. I prefer it staying strike through and letting something else like the color or opacity indicate hover.

 It just feels weird to me that it's hovered then sort of looks enabled. I do the following in my local styleguidist config:

```javascript
    SectionHeading: {
      isDeprecated: {
        '&:hover': {
          textDecoration: 'line-through',
        },
      },
    },
```
and that fixes, but I wonder if you might feel the same and have it default to this from within your lib code which I could simply add to:
https://github.com/styleguidist/react-styleguidist/blob/34e8778bf352e4784caba4e9d3a95f7c4ff4cbcf/src/rsg-components/SectionHeading/SectionHeadingRenderer.js#L42

Lmk if you'd like a PR for that, and if not go ahead and just close this :)

sapegin:
It sounds reasonable, though, I'm not sure about the hover state. Maybe change the color to black and keep it strikethrough?

roblevintennis:
In our env, it looks like the heading it's under takes care of making it `#33` on hover from original `#76` so I don't think anything needs to be added to get desired affect:

![image](https://user-images.githubusercontent.com/142403/41896200-cc991ea4-78d8-11e8-99d6-44b6897362f0.png)

PR #1029, opened 2018-06-26
When hovering on and isDeprecated maintains the link-through.

roblevintennis:
For https://github.com/styleguidist/react-styleguidist/issues/1026 I didn't need to using `inherit` after all...

Unhovered:
screen shot 2018-06-26 at 10 05 20 am

Hovered:

screen shot 2018-06-26 at 10 06 11 am

### DIFF
--- a/src/rsg-components/SectionHeading/SectionHeadingRenderer.js
+++ b/src/rsg-components/SectionHeading/SectionHeadingRenderer.js
@@ -42,6 +42,9 @@ const styles = ({ color, space }) => ({
 	isDeprecated: {
 		textDecoration: 'line-through',
 		color: color.light,
+		'&:hover': {
+			textDecoration: 'line-through',
+		},
 	},
 });
 


src/rsg-components/SectionHeading/SectionHeadingRenderer.js
```diff
@@ -42,6 +42,9 @@ const styles = ({ color, space }) => ({
 	isDeprecated: {
 		textDecoration: 'line-through',
 		color: color.light,
+		'&:hover': {
```
sapegin: '&, &:hover' — this should work, so we won't repeat `textDecoration: 'line-through'`.

roblevintennis: ah yes, thanks, that's better. pushed up commit

roblevintennis:
👍

PR #1033, opened 2018-06-28
Feat: Add support for using glob patterns in arrays and functions

wkillerud:
This commit introduces support for using glob patterns in arrays
and functions provided to the `components`. This feature lets you
for instance include multiple glob patterns under the same section.
This can be useful if your repository has related components spread
across multiple folders, in which case maintaining a single glob
pattern that would include all folders could get tricky. This feature
may for instance be useful if you work with a monorepo.

I've included some very minor changes that I figured I'd include since
I was in the neighbourhood. One test case had a misleading description,
and the JSDoc lacked Array as a valid input to the loader.

I hope you'll consider this feature. It would be useful for us in 
https://github.com/SpareBank1/designsystem where we now do do a
workaround by using glob ourselves (with the error of not including ignores,
but that's a different story :) )

### DIFF
--- a/loaders/utils/getComponentFiles.js
+++ b/loaders/utils/getComponentFiles.js
@@ -3,6 +3,18 @@ const glob = require('glob');
 const isFunction = require('lodash/isFunction');
 const isString = require('lodash/isString');
 
+function getFilesMatchingGlobs(components, rootDir, ignore) {
+	return components
+		.map(listItem => {
+			if (glob.hasMagic(listItem)) {
+				return glob.sync(path.resolve(rootDir, listItem), { ignore });
+			}
+			// Wrap path in an array so reduce always gets an array of arrays
+			return [listItem];
+		})
+		.reduce((accumulator, current) => accumulator.concat(current), []);
+}
+
 /**
  * Return absolute paths of components that should be rendered in the style guide.
  *
@@ -18,9 +30,9 @@ module.exports = function getComponentFiles(components, rootDir, ignore) {
 
 	let componentFiles;
 	if (isFunction(components)) {
-		componentFiles = components();
+		componentFiles = getFilesMatchingGlobs(components(), rootDir, ignore);
 	} else if (Array.isArray(components)) {
-		componentFiles = components;
+		componentFiles = getFilesMatchingGlobs(components, rootDir, ignore);
 	} else if (isString(components)) {
 		componentFiles = glob.sync(path.resolve(rootDir, components), { ignore });
 	} else {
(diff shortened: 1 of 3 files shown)


loaders/utils/getComponentFiles.js
```diff
@@ -3,10 +3,22 @@ const glob = require('glob');
 const isFunction = require('lodash/isFunction');
 const isString = require('lodash/isString');
 
+function getFilesMatchingGlobs(components, rootDir, ignore) {
+	return components
+		.map(listItem => {
+			if (glob.hasMagic(listItem)) {
```
sapegin: What this method does? I'm not sure I'm following the logic of this function.

loaders/utils/getComponentFiles.js
```diff
@@ -3,10 +3,22 @@ const glob = require('glob');
 const isFunction = require('lodash/isFunction');
 const isString = require('lodash/isString');
 
+function getFilesMatchingGlobs(components, rootDir, ignore) {
```
sapegin: Consider either moving this function to a separate file or defining it as an arrow function — these are two conventions we usually use.

loaders/utils/getComponentFiles.js
wkillerud: It's a way to check if a string looks like a glob pattern. That way we only use glob if needed. The name isn't good, but that's what they called it :( 

https://www.npmjs.com/package/glob#globhasmagicpattern-options

loaders/utils/getComponentFiles.js
wkillerud: Sure :) I don't know if the function is useful outside this loader, so perhaps an arrow function is the way to go.  Would this be OK?

```js
const getFilesMatchingGlobs = (components, rootDir, ignore) => 
	components
		.map(listItem => {
			if (glob.hasMagic(listItem)) {
				return glob.sync(path.resolve(rootDir, listItem), { ignore });
			}
			// Wrap path in an array so reduce always gets an array of arrays
			return [listItem];
		})
		.reduce((accumulator, current) => accumulator.concat(current), []);
```

Basically this diff (and adjust indent)

```diff
+ const getFilesMatchingGlobs = (components, rootDir, ignore) =>
- function getFilesMatchingGlobs(components, rootDir, ignore) {
+   components
-	return components
		.map(listItem => {
			if (glob.hasMagic(listItem)) {
				return glob.sync(path.resolve(rootDir, listItem), { ignore });
			}
			// Wrap path in an array so reduce always gets an array of arrays
			return [listItem];
		})
		.reduce((accumulator, current) => accumulator.concat(current), []);
- }
```

loaders/utils/getComponentFiles.js
sapegin: Could you add a comment? The name is really confusing ;-/

PR #1034, opened 2018-06-29
chore(*): up to jest 23

stepancar:
We need upgrade jest to 23 for mock module in https://github.com/styleguidist/react-styleguidist/pull/1034


sapegin:
Thanks, merged! 🦄

Issue #1032, opened 2018-06-28
[Flow] Problem with union and literal props not displaying correctly


fseglard:
Hi there!

**Current behavior**
The props table does not correctly display the list of possible values when using Flow.

_When using React PropTypes:_

![proptypes](https://user-images.githubusercontent.com/13200484/42044532-8f3cffda-7af9-11e8-8c13-511ad3b05d80.png)

_When using FlowTypes:_

![flow](https://user-images.githubusercontent.com/13200484/42044554-99ecccb2-7af9-11e8-8777-5635e3f04d0a.png)

It also display the raw output of the Flow definition when hovering the `union` keyword. I don't know if this is the expected behavior, but it seems weird to me:

![test](https://user-images.githubusercontent.com/13200484/42045329-980613d4-7afb-11e8-8484-1aaf058901a0.png)

**To reproduce**

[https://github.com/fseglard/react-styleguidist-example](https://github.com/fseglard/react-styleguidist-example)

**Expected behavior**

We should have a line displaying the list of possible values like with React PropTypes.

sapegin:
Feel free to send a pull request with a fix! ;-)

stepancar:
@sapegin @fseglard flow doesn't support enums, but react-docgen [support it](https://github.com/reactjs/react-docgen#result-data-structure) and emits it in case of use
propTypes.oneOf(...). I just implemented rendering of union literals in extra of description.

stepancar:
Also we can consider Flow union of literals as enum.
```
export function getType(prop) {
	if (prop.flowType) {
		if (
			prop.flowType.name === 'union' &&
			prop.flowType.elements.every(elem => elem.name === 'literal')
		) {
			return {
				...prop.flowType,
				name: 'enum',
				value: prop.flowType.elements,
			};
		}
		return prop.flowType
	}
	return prop.type;
}
```
@sapegin what do you think about it

sapegin:
This sounds like a good idea!

stepancar:
@sapegin , ok, PR updated

sapegin:
Out in 7.1.0.

PR #1035, opened 2018-06-29
fix(*): fix flow union props rendering

stepancar:
https://github.com/styleguidist/react-styleguidist/issues/1032

### DIFF
--- a/src/rsg-components/Props/util.js
+++ b/src/rsg-components/Props/util.js
@@ -15,7 +15,20 @@ export function unquote(string) {
  * @returns {object}
  */
 export function getType(prop) {
-	return prop.flowType || prop.type;
+	if (prop.flowType) {
+		if (
+			prop.flowType.name === 'union' &&
+			prop.flowType.elements.every(elem => elem.name === 'literal')
+		) {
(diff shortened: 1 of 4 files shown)


sapegin:
Cool, thanks! Could you please have a look if we can add some tests here? 

> 83.33% of diff hit (target 97.56%)

src/rsg-components/Props/util.js
```diff
@@ -15,7 +15,20 @@ export function unquote(string) {
  * @returns {object}
  */
 export function getType(prop) {
-	return prop.flowType || prop.type;
+	if (prop.flowType) {
+		if (
```
sapegin: Could you please add a comment with an example how this looks in Flow?

stepancar: @sapegin hm, https://github.com/styleguidist/react-styleguidist/pull/1035/files#diff-6a050c2c62e825106dba14a944eda318R369

sapegin: It's hard to connect a test with a with a particular line branch of code ;-)

PR #1041, opened 2018-07-04
Fix react-docgen version to exactly 3.0.0-beta12

eemeli:
The `^` or `~` semver operators don't work as expected here, because `'beta12'` ends up being compared as a string, and `'beta9' > 'beta12'`.

This issue has also been reported upstream as reactjs/react-docgen#275.


sapegin:
Thanks!

PR #1042, opened 2018-07-05
Added accessibility semantic changes

eduardoinnorway:
To pass accessibility testing tools like  WAVE

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,6 +138,7 @@
     "pretest": "npm run lint",
     "test": "npm run test:jest",
     "posttest": "npm run format",
+    "snapshot": "jest --updateSnapshot",
     "test:jest": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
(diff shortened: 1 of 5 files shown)


eduardoinnorway:
No need to update package-lock.json, I did not remove or add any packages. I just added a new "script" in package.json to create snapshots for JEST testing.

package.json
```diff
@@ -138,6 +138,7 @@
     "pretest": "npm run lint",
     "test": "npm run test:jest",
     "posttest": "npm run format",
+    "snapshot": "jest --updateSnapshot",
```
sapegin: There's no need for the script, `npx jest -u` is enough.

src/rsg-components/Pathline/PathlineRenderer.js
```diff
@@ -27,6 +27,7 @@ export function PathlineRenderer({ classes, children }) {
 				title="Copy to clipboard"
 			>
 				<MdContentCopy />
+				<span hidden>Copy to clipboard</span>
```
sapegin: `hidden` elements are hidden from both screen and screen readers, so this code doesn't do anything. Could you show a particular error from WAVE? We likely need to duplicate `title` with `arial-label`.

PR #1043, opened 2018-07-09
Feat: Add a user-configurable optional "version" indicator to the sidebar

eemeli:
When documenting libraries (of components or otherwise), I at least find it useful to see which version of said library the documentation pertains to. This PR adds a "version" string field to the config that's passed on to the renderer and if not empty, is included in the sidebar just under the title, with some discreet styling that fits the base theme.

I've also added the version to the basic example, where it's read from the example's `package.json` file.

screen shot 2018-07-09 at 15 02 59

### DIFF
--- a/src/rsg-components/Version/VersionRenderer.js
+++ b/src/rsg-components/Version/VersionRenderer.js
@@ -4,7 +4,7 @@ import Styled from 'rsg-components/Styled';
 
 const styles = ({ color, fontFamily, fontSize }) => ({
 	version: {
-		color: color.lightest,
+		color: color.light,
 		margin: [[5, 0, 0, 0]],
 		fontFamily: fontFamily.base,
 		fontSize: fontSize.base,
@@ -13,7 +13,11 @@ const styles = ({ color, fontFamily, fontSize }) => ({
 });
 
 export function VersionRenderer({ classes, children }) {
-	return <h2 className={classes.version}>{children}</h2>;
+	return (
+		<p aria-label="version" className={classes.version}>
+			{children}
+		</p>
+	);
 }
 
 VersionRenderer.propTypes = {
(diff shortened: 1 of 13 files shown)


src/rsg-components/Version/VersionRenderer.js
```diff
@@ -0,0 +1,24 @@
...
+		color: color.lightest,
+		margin: [[5, 0, 0, 0]],
+		fontFamily: fontFamily.base,
+		fontSize: fontSize.base,
+		fontWeight: 'normal',
+	},
+});
+
+export function VersionRenderer({ classes, children }) {
+	return <h2 className={classes.version}>{children}</h2>;
```
sapegin: I don't think we should use a heading here, but probably it would make sense to add an accessible label:

```html
<p aria-label="version">{children}</p>
```

src/rsg-components/Version/VersionRenderer.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Styled from 'rsg-components/Styled';
+
+const styles = ({ color, fontFamily, fontSize }) => ({
+	version: {
+		color: color.lightest,
```
sapegin: Please use `light` for text, `lightest ` doesn't have enough contrast.

sapegin:
Thanks! Please update Jest snapshots, and I could merge it ;-)

eemeli:
Oops, sorry, had forgotten about that! Should now be fixed.

stepancar:
As for me - new option 'version' increase complexity. New propType, new logic (show if passed) etc. Maybe in the future you will need another option like 'isBeta' 'isAlpha' etc. More simple way to provide this functionality - just [unwrap title](https://github.com/styleguidist/react-styleguidist/pull/1043/files)
and pass title and version to RSGMainComponent like this
```jsx
<StyleGuideRenderer
        {...anotherProps}
 	title={[
            <h1>{config.title}</h1>,
            <h2>{config.version}</h2>
        ]}
/> 
```

sapegin:
@stepancar I agree with you in general, but I think this particular feature is a nice addition to GitHub ribbons.

stepancar:
@sapegin , ok)

PR #1046, opened 2018-07-10
fix(*): fix preview of method optional argument

stepancar:


src/rsg-components/Argument/ArgumentRenderer.js
```diff
@@ -13,6 +13,11 @@ export const styles = ({ space }) => ({
 });
 
 export function ArgumentRenderer({ classes, name, type, description, returns, block, ...props }) {
+	const isOptional = type && type.type && type.type === 'OptionalType';
```
sapegin: `type && type.type === 'OptionalType'` would be enough.

src/rsg-components/Argument/ArgumentRenderer.js
```diff
@@ -22,7 +27,11 @@ export function ArgumentRenderer({ classes, name, type, description, returns, bl
 					{type && ':'}
 				</span>
 			)}
-			{type && <Type>{type.name}</Type>}
+			{type && (
+				<Type>
+					{type.name + (isOptional ? '?' : '') + (defaultValue ? `=${defaultValue}` : '')}
```
sapegin: Maybe `{type.name}{isOptional && '?'}{!!defaultValue && `=${defaultValue}`}`?

loaders/utils/getProps.js
```diff
@@ -52,9 +52,14 @@ module.exports = function getProps(doc, filepath) {
 
 	// Parse the docblock of the remaining methods with doctrine to retrieve the JsDoc tags
 	doc.methods = doc.methods.map(method => {
-		return Object.assign(method, {
-			tags: getDoctrineTags(doctrine.parse(method.docblock)),
-		});
+		const tags = getDoctrineTags(doctrine.parse(method.docblock, { sloppy: true, unwrap: true }));
```
sapegin: Would make sense to explain what is happening here: quite a lot of tricky code.

src/rsg-components/Argument/ArgumentRenderer.js
stepancar: Fixed

src/rsg-components/Argument/ArgumentRenderer.js
stepancar: agree

loaders/utils/getProps.js
stepancar: add comment for this lines

stepancar: @sapegin , I also suggest to filter tags with title 'param' and 'returns' , because we mapped it to params and returns method info. What do you think about it?

sapegin: You mean to accept synonyms, or am I misunderstanding you?

* @param (synonyms: @arg, @argument)
* @returns (synonyms: @return)

stepancar: @sapegin , I suggeset map @param, @arg, @argument, @return, @returns info to method 'returns', 'params' and remove it from 'tags'.

sapegin: Makes sense I guess. Do you want to try it here or should I merge it as is?

stepancar: @sapegin, I want try it here.

stepancar: @sapegin, I tried - look at updated tests

stepancar:
@sapegin , as I understand, it's CI problems

sapegin:
You need this commit to fix the Cypress issue: ff4014841c85ed6181327ff02015e9d915e5cfe7

craigkovatch:
@stepancar @sapegin I'm working on supporting parameter types in react-docgen-typescript#172 and we could use some clarify from you on this PR -- it seems to us that the format for an optional parameter would be `param?: type` rather than `param: type?`. Are we missing something? Of course in TypeScript types/parameter optionality are a 1st-class concept rather than a jsdoc concept, so maybe there's some impedance mismatch there.

PR #1048, opened 2018-07-12
Feature: Custom mounting point id

manatok:
By default RSG binds to a div with an id of `rsg-root` with no way of overriding this. We needed to import multiple RSG bundle files into our work documentation site and as such needed to be able to specify the id of the element we wanted the bundles to bind to.

In your `styleguide.config.js` you can now specify an optional `mountPointId` to specify the id of the DOM element you want the app to bind to. It will default to the existing `rsg-root` if not specified.

I have tried to add some tests but might need some guidance here to make sure I am on the right track.


manatok:
Going to fix the conflicts now.

manatok:
I accidentally branched off of the release tag originally and ran into issue merging in master where it was reverting the changes done on master. Strange. Going to branch off of master now and will create a new PR soon.

PR #1050, opened 2018-07-12
Feat: Made the mounting point id configurable.

manatok:
By default RSG binds to a div with an id of `rsg-root` with no way of overriding this. We needed to import multiple RSG bundle files into our work documentation site and as such needed to be able to specify the id of the element we wanted the bundles to bind to.

In your `styleguide.config.js` you can now specify an optional `mountPointId` to specify the id of the DOM element you want the app to bind to. It will default to the existing `rsg-root` if not specified.

I have tried to add some tests but might need some guidance here to make sure I am on the right track.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -281,7 +281,7 @@ module.exports = {
 
 Type: `string`, defaults: `rsg-root`
 
-The id of the DOM element that React-Styleguidist mounts.
+The ID of the DOM element that React-Styleguidist mounts.
 
 #### `pagePerSection`
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -21,6 +21,7 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 - [`handlers`](#handlers)
 - [`ignore`](#ignore)
 - [`logger`](#logger)
+- [`mountPointId`](#mountPointId)
 - [`pagePerSection`](#pagepersection)
 - [`printBuildInstructions`](#printbuildinstructions)
 - [`printServerInstructions`](#printserverinstructions)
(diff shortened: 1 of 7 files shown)


sapegin:
Could you elaborate a bit on your use case? I'm not against adding this option, but it sounds a bit weird anyway ;-)

manatok:
We have an open source documentation tool (http://presidium.spandigital.net/) that serves .md files as static html pages. We have been working on a number of plugins to pull content into presidium  (javadocs, swagger definitions etc) and we are currently adding a plugin for react-styleguidist.

Each article in presidium lives in it's own .md file, which is then linked to the navigation and allows users to interact with it (e.g adding comments). In order to achieve this with RSG, so that it fits in nicely with presidium, our plugin runs RSG against each component separately, so that we can generate a separate bundle and .md per component. As a result, we need to embed multiple components on one page.

docs/Configuration.md
```diff
@@ -276,6 +277,12 @@ module.exports = {
 }
 ```
 
+#### `mountPointId`
+
+Type: `string`, defaults: `rsg-root`
+
+The id of the DOM element that React-Styleguidist mounts.
```
sapegin: id → ID
React-Styleguidist → Styleguidist

scripts/make-webpack-config.js
```diff
@@ -27,7 +27,7 @@ module.exports = function(config, env) {
 	const htmlPluginOptions = {
 		context: Object.assign({}, templateContext, {
 			title: config.title,
-			container: 'rsg-root',
+			container: config.mountPointId || 'rsg-root',
```
sapegin: You already have a default value from the config schema, so you don't need `||` here.

Issue #1061, opened 2018-07-23
`react-docgen-displayname-handler` has a peerDependency of `recast`


everdimension:
**Current behavior**

`react-styleguidist` depends on `react-docgen-displayname-handler`, which, in turn, has a peerDependency on `recast`. I get an npm warning:

```
npm WARN react-docgen-displayname-handler@2.1.0 requires a peer of recast@0.12.6 || 0.13.0 but none is installed. You must install peer dependencies yourself.
```

**Expected behavior**

`react-styleguidist` should list `recast` as a dependency.

sapegin:
I guess it's something new, otherwise it would break before. Would you mind sending a pull request with a fix?

everdimension:
Sure

everdimension:
@sapegin I actually think that this is a more appropriate fix: https://github.com/nerdlabs/react-docgen-displayname-handler/pull/6

But since I don't how long it will take them to accept this PR and whether they will do it at all, I created a fix here as well: https://github.com/styleguidist/react-styleguidist/pull/1062

sapegin:
Yeah, I think your other solution is better. Let's wait a bit if they merge it soon. 

Am I right that this isn't blocking issue and just a warning?

everdimension:
I still think you should explicitly list `recast` as a dependency simply because you're depending on a package that has it as a `peerDependency`.

Right now you do not have warnings locally because you have other packages that install it. To be more precise:

<pre>
react-styleguidist@0.0.0-development
├─┬ react-docgen@3.0.0-beta12
│ └── <b>recast@0.13.2</b>
└─┬ webpack-cli@2.1.5
  ├─┬ jscodeshift@0.5.1
  │ └── <b>recast@0.15.0</b>
  └─┬ webpack-addons@1.1.5
    └─┬ jscodeshift@0.4.1
      └── <b>recast@0.12.9</b>
</pre>

But you shouldn't rely on this.

cwmacdon:
Unfortunately this issue should probably be re-opened, as the peerDependency uses exact versions:

`"recast": "0.12.6 || 0.13.0"`

and the commit in this repo uses an inexact version:

`"recast": "^0.13.0"`

This will install `0.13.2` which does not meet the peerDependency requirement :(

everdimension:
[...]

PR #1062, opened 2018-07-23
Add explicit recast dependency

everdimension:
...because `react-docgen-displayname-handler` has it as a peerDependency


everdimension:
fixes https://github.com/styleguidist/react-styleguidist/issues/1061

sapegin:
Merging, thanks!

PR #1063, opened 2018-07-23
Update Readme.md

amandakoster:
Hello, I though changing 'Lick me' to 'Click me' was good.


sapegin:
Why do you think so? Did you find “lick me” offensive? I'm find to change it if there's any reason to do so ;-)

tizmagik:
I found it kind of awkward, personally, so I'm a fan of this change 😆 

But the GIF in the main readme should also be updated to match

PR #1068, opened 2018-07-30
Fix: restoring support for evalInContext for custom slots (#1067)

### LINKED ISSUE #1067: Custom slot evalInContext support regressed
#977 added support for custom slots (such as #968)  to use `evalInContext`. 

#1040 removed that support, probably just to clean up apparently unnecessary code.

I believe the plugins api in #505 will eventually be the right way to handle this, but in the meantime, I'd like to restore the support for `evalInContext` in slots.

kidkuro:
Fixes #1067


sapegin:
Thanks!

PR #1069, opened 2018-08-01
Add support for html webpack plugin

notgiorgi:
Adds ability to use `html-webpack-plugin` for more advanced features, like custom templates.

#896


sapegin:
Thanks for the PR, but we're not going back to the html plugin. Custom templates are already supported:

https://github.com/styleguidist/mini-html-webpack-plugin#custom-templates

PR #1087, opened 2018-08-13
slots components path changes

### LINKED ISSUE #1070: Overriding Slots or UsageTabButton
I have to segregate my props and methods into 2 different tabs. (Currently, they are all shown under 1 tab).

I have been trying to override UsageTabButton in particular, but have been unable to do so. 

Custom rendering and therefore, overriding doesn't seem to be working for either Slot or contents of slots.

Please let me know if there's a way.

sibamohanty995:
Fixes #1070


sapegin:
Thanks!

PR #1091, opened 2018-08-17
Fix: Appear page not found when a component is isolated

rafaesc:
If isolated is `true`, so it shoud try to search with recursivity until find the component.


sapegin:
Thanks!

PR #1116, opened 2018-08-30
Feat: Do not put component into global namespace

sapegin:
BREAKING CHANGE:

Style guide componets are no longer accessible on window. Only current component is accessible in the example context (not on window). You need to explicitly import any other component.


examples/basic/src/components/Button/Readme.md
```diff
@@ -44,6 +44,22 @@ Fenced blocks with other languages are rendered as highlighted code:
 <h1>Hello world</h1>
 ```
 
+Current component (like `Button` in this example) is always accessible by its name in the example code. If you want to use other component, you need to explicitly import them:
```
tizmagik: grammar, either:

> If you want to use **another** component, you need to explicitly import it

or

> If you want to use other component**s**, you need to explicitly import them

tizmagik:
I really like this change! Will help a lot of large styleguides. 👍

phpnode:
This change is bittersweet to me, I understand the reasoning but it does take away some convenience, especially when experimenting with examples in the code editor. To be honest it's making me reconsider using markdown at all - normal .js files obviously have tooling advantages.

Would you accept a PR which:
- Makes this an opt-in configuration option.
- Injects components into the function scope rather than polluting `window`?

I've got a project with 283 markdown files here and not enjoying the prospect of converting them by hand!

sapegin:
Keep in mind that the current component is still available, and in my experience it's the majority of the use cases. 

You can use [context](https://react-styleguidist.js.org/docs/configuration.html#context) option to list components you want to access in your examples — that should be more reliable than autoloading everything on Styleguidist side, which will have similar issues with duplicated component names, that we have now.

PR #1121, opened 2018-08-31
Remove error default value for serverPort

BTMPL:
I checked the source but I don't think this config reads in any env variables?


sapegin:
You're right, and `NODE_ENV` isn't a port anyway ;-/

PR #1124, opened 2018-08-31
Add alias for test folder

elevatebart:
To simplify the tests, I will try to replace all references to the `test` folder, usually `../../../test/`, with simply `@test/`. It should make the code easier to refactor.

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -7,7 +7,7 @@
     "jsx": "react",
     "baseUrl": "./src",
     "paths": {
-      "@test/*": ["../test/*"]
+      "_test/*": ["../test/*"]
     }
   },
   "exclude": [
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -5,7 +5,10 @@
       "esnext"
     ],
     "jsx": "react",
-    "baseUrl": "./src"
+    "baseUrl": "./src",
+    "paths": {
+      "@test": ["test"]
+    }
   },
   "exclude": [
     "node_modules"
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -7,7 +7,7 @@
     "jsx": "react",
     "baseUrl": "./src",
     "paths": {
-      "@test": ["test"]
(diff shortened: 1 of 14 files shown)


elevatebart:
As styleguidist is using eslint no-extraneous and eslint no-unresolved we cannot use aliases starting with an @ sign. I was hoping as well that require.resolve would use the aliases of jest moduleMapper. Turns out it does not care. I ended up using another solution, setting up a global variable.

sapegin:
Thanks for the effort! I don't think it solves any problem though, but introduces an abstraction that developers has to know when they read or write tests. We'll have to check for that in code reviews as well. And all that just not to write `../../..` in a few places.

elevatebart:
You are absolutely right @sagepin, i got tricked by github into making a PR to your master instead of mine. This was actually an experiment to simplify the migration to the multipackage/monorepo architecture I am looking into adding to react-styleguidist.

My final goal is to find a way to separate the common grounds of vue-styleguidist and react-styleguidist to make it a third package. This package called styleguidist-core could hold the shared code between rsg and vsg. This way all the code that you maintain could stay in your package and we could just maintain the part that concerns vuejs.

What I do not want is to make life harder for you in any way. I would like you to continue maintaining only one repo with automated semantic release.

elevatebart:
@tizmagik thanks for the tip. It was indeed my first instinct to use webpack aliases. More accurately I was tring to use the [moduleNameMapper option of jest](https://jestjs.io/docs/en/configuration.html#modulenamemapper-object-string-string) since all the references to the test folder are in jest tests. 

Unfortunately rsg is using the `path.resolve` function to obtain the path to the tested config directory. Like in [styleguideloader.spec.js](https://github.com/styleguidist/react-styleguidist/blob/master/loaders/__tests__/styleguide-loader.spec.js#L9). That does not take aliases into account. Neither does require.resolve.

I will try it another way and close this PR. Thank you for the feedback.

PR #1125, opened 2018-09-03
fix: allows pathline to wrap

derz:
- the "Pathline" breaks the responsive view if it gets too long as there will be a horizontal scrollbar/or the whole document will be scaled down (depending on the browser)


sapegin:
Thanks!

Issue #879, opened 2018-03-20
Improve accessibility


sapegin:
## To do

(Will be done in #893)

- [x] Tweak colors to comply with WCAG 2 AA contrast ratio (gray text, for example).
- [x] Add an invisible label to the search field (`aria-label`).
- [x] Add [eslint-plugin-jsx-a11y](https://github.com/evcohen/eslint-plugin-jsx-a11y)

## To think

- [ ] Missing `lang` attribute in `html` tag. Do we need a config option for that?

Am I missing anything? 

/cc @okonet

gergely-nagy:
I'd like to jump in on this! :rocket:

sapegin:
@gergely-nagy Awesome! 🦄

Jlevyd15:
@sapegin noticed a few color contrast issues with the example Styleguide docs. I've created a PR #1126 to address those. I figured those would land within the scope of this issue but if you think it needs a separate issue let me know.

PR #1126, opened 2018-09-03
fix - partial fix for #879

Jlevyd15:
### Background
- improves accessibility in the example application for various components
- to meet WCAG 2.0 color contrast requirements foreground/background must have a ratio of 4.5.1 or greater see docs [HERE](https://webaim.org/resources/contrastchecker/)
- using aXe chrome plugin for testing more info. [HERE](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US)
### Changes
- side-bar navigation links, darkened link color #1978c8 -> #1873bf
- PropType table, darkened prop ```name``` color #7f9a44 -> #667c36
- PropType table, darkened ```type``` color #b77daa -> #a15994
- VersionRenderer - changed the text color from ```color.light``` -> ```color.error```


sapegin:
Thanks! This is very good kind of PR! Could you please have a look at the `next` branch, I've changed a few colors there, and sidebar links should already WCAG 2.0 compliant.

And we probably need to add another shade of gray text — using error color is incorrect here, and it's way to bright for such thing as version number.

PR #1128, opened 2018-09-04
Fix(config): better resolving of components on windows

elevatebart:


sapegin:
> nocase Perform a case-insensitive match. Note: on case-insensitive filesystems, non-magic patterns will match by default, since stat and readdir will not raise errors.

If I understand this correctly, it's case-insensitive on Windows in any case, so your fix is likely relying on some filtering they do when this flag is enabled. Probably we should report it to glob as well.

loaders/utils/getComponentFiles.js
```diff
@@ -8,7 +8,11 @@ const getFilesMatchingGlobs = (components, rootDir, ignore) =>
 		.map(listItem => {
 			// Check if the string looks like a glob pattern by using hasMagic
 			if (glob.hasMagic(listItem)) {
-				return glob.sync(path.resolve(rootDir, listItem), { ignore });
+				return glob.sync(listItem, {
+					cwd: rootDir,
+					ignore,
+					nocase: process.platform === 'win32',
```
sapegin: Please add a comment with the explanation here ;-)

loaders/utils/getComponentFiles.js
```diff
@@ -28,21 +32,25 @@ module.exports = function getComponentFiles(components, rootDir, ignore) {
...
 
-	let componentFiles;
-	if (isFunction(components)) {
-		componentFiles = getFilesMatchingGlobs(components(), rootDir, ignore);
-	} else if (Array.isArray(components)) {
-		componentFiles = getFilesMatchingGlobs(components, rootDir, ignore);
-	} else if (isString(components)) {
-		componentFiles = glob.sync(path.resolve(rootDir, components), { ignore });
-	} else {
+	const componentGlobsArray = isFunction(components)
```
sapegin: Just `componentGlobs` will be enough.

loaders/utils/getComponentFiles.js
```diff
@@ -28,21 +32,25 @@ module.exports = function getComponentFiles(components, rootDir, ignore) {
...
+		? components()
```
sapegin: A function will be more readable than a nested ternary:

```js
const getComponentGlobs = components => {
  if (isFunction(components)) {
    return components();
  }
  // ...
}
```

loaders/utils/getComponentFiles.js
```diff
@@ -28,21 +32,25 @@ module.exports = function getComponentFiles(components, rootDir, ignore) {
...
+		: Array.isArray(components)
+			? components
+			: isString(components)
+				? [components]
+				: 0;
```
sapegin: `false` or `undefined` would be a more clear way to say that the input was incorrect.

loaders/utils/getComponentFiles.js
```diff
@@ -28,21 +35,26 @@ module.exports = function getComponentFiles(components, rootDir, ignore) {
...
 
-	let componentFiles;
-	if (isFunction(components)) {
-		componentFiles = getFilesMatchingGlobs(components(), rootDir, ignore);
-	} else if (Array.isArray(components)) {
-		componentFiles = getFilesMatchingGlobs(components, rootDir, ignore);
-	} else if (isString(components)) {
-		componentFiles = glob.sync(path.resolve(rootDir, components), { ignore });
-	} else {
+	const getComponentGlobs = components => {
```
sapegin: Could you please move this function to the top of the  file?

elevatebart:
Done ;)

PR #1131, opened 2018-09-07
adds inline-scroll on props-table 

derz:
Same as: https://github.com/styleguidist/react-styleguidist/pull/1125
Document gets a horizontal scrollbar on a small viewport because of the props-table which gets too wide.

### DIFF
--- a/src/rsg-components/ReactComponent/ReactComponentRenderer.js
+++ b/src/rsg-components/ReactComponent/ReactComponentRenderer.js
@@ -16,6 +16,11 @@ const styles = ({ color, fontSize, space }) => ({
 	tabButtons: {
 		marginBottom: space[2],
 	},
+	tabBody: {
+		overflowX: 'auto',
+		maxWidth: '100%',
+		'-webkit-overflow-scrolling': 'touch',
+	},
 	docs: {
 		color: color.base,
 		fontSize: fontSize.text,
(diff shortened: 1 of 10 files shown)


src/rsg-components/ReactComponent/ReactComponentRenderer.js
```diff
@@ -16,6 +16,11 @@ const styles = ({ color, fontSize, space }) => ({
 	tabButtons: {
 		marginBottom: space[2],
 	},
+	tabBody: {
+		overflowX: 'auto',
+		maxWidth: '100%',
+		'-webkit-overflow-scrolling': 'touch',
```
sapegin: Will `WebkitOverflowScrolling` work too?

derz: Sure, yes.  :-) However it is the same as here: https://github.com/styleguidist/react-styleguidist/blob/b66bd10665bc134135a8affbc807a01f92417b6c/src/rsg-components/StyleGuide/StyleGuideRenderer.js#L40 - but I am not super into styled-components, so I don't really know what's the intended way to do it. Btw.: Thanks for your great work on it so far!

sapegin: Both should work the same way, camelCased one is just more idiomatic, because all other attributes are camelCased.

It's JSS, not styled-components by the way ;-)

sapegin:
And please update snapshots too ;-)

sapegin:
Thanks for fixing these small details! ❤️

PR #1133, opened 2018-09-11
Fix(config): sections not rendering correctly

### LINKED ISSUE #1132: Side navigation broken in 7.3.6
**Current behavior**
Out of 6 parent sections in my sidebar only 3 are displaying, and none of their component links are showing up.

**To reproduce**

This section is taken from my styleguide.config.js, which has been working through 7.3.5:

```
  sections: [
    {
      name: 'Themes',
      content: path.join(baseComponentDir, 'theme/Themes.md')
    },
    {
      name: 'Base',
      components: path.join(baseComponentDir, 'base/**/*.js')
    },
    {
      name: 'Containers',
      content: path.join(baseComponentDir, 'containers/Containers.md'),
      components: path.join(baseComponentDir, 'containers/**/*.js')
    },
    {
      name: 'Controls',
      components: path.join(baseComponentDir, 'controls/**/*.js')
    },
    {
      name: 'Navigation',
      components: path.join(baseComponentDir, 'navigation/**/*.js')
    },
    {
      name: 'Patterns',
      content: path.join(baseComponentDir, 'patterns/Patterns.md'),
      components: path.join(baseComponentDir, 'patterns/**/*.js')
    }
  ],
```

**Current behavior 7.3.6**

**Expected behavior (7.3.5)**

I should note that the components themselves are also not rendering in the main content area. Just the headings.

elevatebart:
Closes #1132

Fixes and simplifies configuration on windows


loaders/utils/getComponentFiles.js
```diff
@@ -18,21 +18,15 @@ const getComponentGlobs = components => {
 
 const getFilesMatchingGlobs = (components, rootDir, ignore) =>
 	components
-		.map(listItem => {
-			// Check if the string looks like a glob pattern by using hasMagic
-			if (glob.hasMagic(listItem)) {
```
elevatebart: @sapegin, why did you add this optimization? Was glob too slow with some configs?
I believe removing it will make @darrken config work again.

sapegin: It wasn't me ;-) I guess it won't be a problem to treat all strings as glob patterns.

elevatebart: on another note, at line 58 here, you absolutize components paths again.
```js
// Make paths absolute
const componentFilesAbsolute = componentFiles.map(file => path.resolve(rootDir, file));
```

Why is that? If you already do
```js
glob.sync(path.resolve(path.resolve(rootDir, listItem))
```
it seems unnecessary since the glob is gonig to return absolute paths anyway.

elevatebart: @sapegin looking at the git history, it seems like it was necessary for all non-glob values. Since we are considering everything as a glob now, I will remove that line.

sapegin: There's no “me”, it's an open source project ;-)

Maybe it is unnecessary now, maybe it used to be necessary, I can't say now.

loaders/utils/getComponentFiles.js
```diff
@@ -1,5 +1,5 @@
 const path = require('path');
-const glob = require('glob');
+const glob = require('fast-glob');
```
elevatebart: Since the issue seemed to come from glob anyway, I dedided to see if [fast-glob](https://github.com/mrmlnc/fast-glob) was working better. It is faster, more accurate and works on windows.

@sapegin I hope you are ok with it. This simplifies greatly my fix.

sapegin: I'm fine with that! My only question is how compatible it is with `glob` and will it require a major release, which would be fine too — I'm finishing 8.0.0 anyway.

loaders/utils/getComponentFiles.js
```diff
@@ -16,24 +16,20 @@ const getComponentGlobs = components => {
...
-					ignore,
-					// in order to avoid detecting each component twice on windows
-					// when matching 2 cases of the same word, like {Src,src}
-					// we remove case-sensitivity on windows
-					nocase: process.platform === 'win32',
-				});
-			}
-			// Wrap path in an array so reduce always gets an array of arrays
-			return [listItem];
+			const compFiles = glob.sync(listItem, {
```
sapegin: Now you could return `glob.sync` results directly without an extra variable and brackets.

loaders/utils/getComponentFiles.js
```diff
@@ -16,24 +16,20 @@ const getComponentGlobs = components => {
...
+				cwd: rootDir,
+				ignore,
+			});
+			return compFiles;
 		})
-		.reduce((accumulator, current) => accumulator.concat(current), []);
+		.reduce((accumulator, current) => {
```
sapegin: Same here: no need for explicit return.

loaders/utils/getComponentFiles.js
elevatebart: I believe it does not need a major. I was under the impression that both implementations were almost absolutely identical. The difference, according to the author was in the use of promises and streams internals.

elevatebart: [Food For Thoughts](https://github.com/mrmlnc/fast-glob#compatible-with-node-glob)
Long story short it is supposed to behave like node-glob for the options we use.

One issue though, it seems to be some problems with *ignore*. 
mrmlnc/fast-glob#86 mrmlnc/fast-glob#112

It might break some styleguides. It is probably better not to switch too quickly.

I think I just have to remove the faulty `nocase`. Do you think it is ok to match on `components` lower case 'c' and forget about `Components` for now?

Or should we just report a bug to node-glob and hope for the phoenix to be reborn...

isaacs/node-glob#345 & isaacs/node-glob#346 are necessary for correct case management on windows.

I will post an issue.

In the meantime do you mind if I change the default glob on windows to 
```js
const DEFAULT_COMPONENTS_PATTERN =
	// on windows, the case insensitivity makes each component appear twice
	// to avoid this issue, the case management is removed on win32
	// it virtually changes nothing
	process.platform === 'win32'
		? `src/components/**/*.{${EXTENSIONS}}`
		: `src/@(components|Components)/**/*.{${EXTENSIONS}}`;
```

sapegin:
Looks like you need to revert test case changes, otherwise looks good! I'd also add a `HACK:` in front of the comment ;-)

elevatebart:
Hello @sapegin,
 I changed the test cases because I have no choice. The removal of `glob.hasMagic` has changed the behaviour of `getComponentFiles`. It now always goies through the glob. This means that the array of non existent files is not returned anymore if we pass it to the function. I changed the paths to point to real test files. And the tests are passing again. 

Codecov was actually complaining about another file: config.js. I added a new branch, the windows specific glob without testing it. Since I do not think we plan on making tests pass on windows, I simply ignored the coverage for his line.

Bart

sapegin:
Ah, I see! That's OK ;-)

ThorHelms:
Just FYI: It seems that this breaks Styleguidist on Windows when using uppercase 'Components' folder, as the resources will be requested with lowercase 'components' which doesn't exist.
For now I've been lazy and simply reverted to 7.3.7.

elevatebart:
Hello @ThorHelms, what can you share your version of windows, the command prompt you are using and your version of nodejs? I would like to investigate this issue.

elevatebart:
@ThorHelms another question, did you use the `Components` folder before 7.3.7 ?

ThorHelms:
I am using:
Windows 10 Pro, version 1803
PowerShell 5.1
NodeJS 8.11.2

I used the `Components` folder with Styleguidist 7.2.2, and after an update to 7.3.8 it no longer worked.

PR #1140, opened 2018-09-17
Allow chunkify to accept other extensions than js, jsx and html

elevatebart:
I am in the process of trying to simplify vue-styleguidist and I need to extract every react specific detail of styleguidist.
In this spirit, we want to be able to use example-loader.
Bad luck, we allow the use of "vue" as a language for our examples. 

This brought https://github.com/vue-styleguidist/vue-styleguidist#200

I thought the best way to clean up vsg's code while making rsg a little more powerful was to add an option to the example-loader customLangs.

Wether it's additional langs or replace entirely can be changed.

Thank you for your remarks.
Bart

### DIFF
--- a/loaders/utils/__tests__/chunkify.spec.js
+++ b/loaders/utils/__tests__/chunkify.spec.js
@@ -128,14 +128,30 @@ it('should call updateExample function for example', () => {
 	expect(actual).toEqual(expected);
 });
 
-it('should even parse examples with customized extensions', () => {
+it('should even parse examples with custom extensions', () => {
 	const markdown = `
-Pass props to CodeRenderer
+Custom extensions
 
 \`\`\`vue
-<AppButton>Example in frame and Without editor</AppButton>
+<AppButton>Example in vue</AppButton>
 \`\`\`
 `;
 	const actual = chunkify(markdown, undefined, ['vue']);
+	expect(actual).toMatchSnapshot();
+});
+
+it('should parse undefined custom extensions without throwing', () => {
+	const markdown = `
+Undefined extensions (default)
+
+\`\`\`jsx
+<AppButton>Example in jsx with undefined extensions</AppButton>
+\`\`\`
+
+\`\`\`pizza
+<AppButton>Example in pizza with undefined extensions</AppButton>
+\`\`\`
+`;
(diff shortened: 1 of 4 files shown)


loaders/utils/chunkify.js
```diff
@@ -11,9 +11,11 @@ const CODE_PLACEHOLDER = '<%{#code#}%>';
  *
  * @param {string} markdown
  * @param {Function} updateExample
+ * @param {Array<string>} playgroundLangs
  * @returns {Array}
  */
-module.exports = function chunkify(markdown, updateExample) {
+module.exports = function chunkify(markdown, updateExample, playgroundLangs) {
+	playgroundLangs = playgroundLangs || PLAYGROUND_LANGS;
```
sapegin: You could use a default value here:

```js
function chunkify(markdown, updateExample, playgroundLangs = PLAYGROUND_LANGS)
```

loaders/utils/__tests__/chunkify.spec.js
```diff
@@ -127,3 +127,15 @@ it('should call updateExample function for example', () => {
 	const actual = chunkify(markdown, updateExample);
 	expect(actual).toEqual(expected);
 });
+
+it('should even parse examples with customized extensions', () => {
```
sapegin: “should parse examples with custom extensions”?

loaders/utils/chunkify.js
elevatebart: @sapegin unfortunately the default value here will not work.
The use of chunkify in the `examples-loader` is 
```js
// Load examples
const examples = chunkify(source, updateExample, query.customLangs);
```
where `query.customLangs` is going to be `undefined`, not ommitted.
If I use a default value here, it breaks all the existing uses of the loader.

sapegin: That's exactly what Babel will generate: comparison with `undefined`, as well as Node.js:

```
❯ node
> function x(y=3) { return y }
undefined
> x()
3
> x(undefined)
3
```

So the problem must be something else.

loaders/utils/__tests__/chunkify.spec.js
elevatebart: Your English is better than mine. Shame on me.

PR #1142, opened 2018-09-19
Import aliases

### LINKED ISSUE #1076: Import rewrites
The idea is to make example’s code more real-life by allowing you to `import` components like the user would import them in their app, including all dependencies like React. For example:

```js
import React from 'react'
import Button from 'my-awesome-library/Button'
```

Technically it's not needed because both React and the current component are available in the example context by default, but it will make example easier to use, because users could copy code directly to their apps.

To make this possible we'll need to add a new config option like:

```js
rewriteRequire(module) {
  return module.replace('my-awesome-library', '../')
}
```

The only think I'm not sure about is what to do with relative path.

sapegin:
Closes #1076

Add new config option `moduleAliases` to define aliases for modules, that you can import in your examples, to make example code more realistic and copypastable:

```javascript
const path = require('path')
module.exports = {
  moduleAliases: {
    'rsg-example': path.resolve(__dirname, 'src')
  }
}
```

````jsx static
// ```jsx inside Markdown
import React from 'react'
import Button from 'rsg-example/components/Button'
import Placeholder from 'rsg-example/components/Placeholder'
````

image 2018-09-19 at 1 29 23 pm

Check out the [webpack resolve.alias documentation](https://webpack.js.org/configuration/resolve/#resolve-alias) for available syntax.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -271,9 +271,33 @@ module.exports = {
 }
 ```
 
+#### `moduleAliases`
+
+Type: `object`, optional
+
+Define aliases for modules, that you can import in your examples, to make example code more realistic and copypastable:
+
+```javascript
+const path = require('path')
+module.exports = {
+  moduleAliases: {
+    'rsg-example': path.resolve(__dirname, 'src')
+  }
+}
+```
+
+````jsx static
+// ```jsx inside Markdown
+import React from 'react'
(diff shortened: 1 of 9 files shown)


kopax:
This is broken in v9.0.3 when I use it with webpack 4.29.6.

PR #1148, opened 2018-09-23
Fix: remove webpack 4 deprecation warning

elevatebart:
After seeing the following warning a great enough number of times, I decided to tackle the issue.
```bash
(node:13435) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
```

Three object were still using the webpack 3 way:
 - `StyleguidistOptionsPlugin`, but I guess you already worked on that
 - `bin/styleguidist.js` that was hooking directly onto webpack
 - `webapck-dev-server@2.X.X` that is incompatible with webpack 4. 

I therefore upgraded the latest and fixed both the other files. The warning is gone.

Enjoy

### DIFF
--- a/bin/styleguidist.js
+++ b/bin/styleguidist.js
@@ -107,13 +107,21 @@ function commandBuild() {
 	verbose('Webpack config:', compiler.options);
 
 	// Custom error reporting
-	compiler.plugin('done', function(stats) {
+	function customErrorBuild(stats) {
 		const messages = formatWebpackMessages(stats.toJson({}, true));
 		const hasErrors = printAllErrorsAndWarnings(messages, stats.compilation);
 		if (hasErrors) {
 			process.exit(1);
 		}
-	});
+	}
+
+	if (compiler.hooks) {
+		// webpack 4
+		compiler.hooks.done.tap('rsgCustomErrorServer', customErrorBuild);
+	} else {
+		// webpack 3
+		compiler.plugin('done', customErrorBuild);
+	}
 }
 
 function commandServer() {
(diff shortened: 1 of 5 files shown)


bin/styleguidist.js
```diff
@@ -107,13 +107,21 @@ function commandBuild() {
...
 		const hasErrors = printAllErrorsAndWarnings(messages, stats.compilation);
 		if (hasErrors) {
 			process.exit(1);
 		}
-	});
+	}
+
+	if (compiler.hooks) {
+		// webpack 4
+		compiler.hooks.done.tap('rsgCustomErrorServer', customErrorBuild);
```
sapegin: `rsgCustomErrorBuild`?

package.json
```diff
@@ -87,7 +87,7 @@
     "type-detect": "^4.0.8",
     "uglifyjs-webpack-plugin": "1.2.7",
     "unist-util-visit": "^1.3.1",
-    "webpack-dev-server": "^2.11.2",
+    "webpack-dev-server": "^3.1.8",
```
sapegin: Have you tried it with webpack 3?

elevatebart: I tested with webpack 3.10.0 using npm link and it was working fine except for a peer dependency warning at install.

bin/styleguidist.js
elevatebart: good catch ! ;-)

package.json
elevatebart: even if according to [this release note](https://github.com/webpack/webpack-dev-server/releases/tag/v3.0.0) it should not. I will check their changelog.

sapegin: I remember I had troubles finding a version that would work in webpack 3 and 4.

You can do `npm i --no-save webpack@3` and then `npm start`.

elevatebart:
Hello @sapegin I finally found the time to run those webpack 3 tests. Even if the tests seem to pass, when testing manually I have a big old error in webpack-dev-server...

```
Cannot destructure property `compile` of 'undefined' or 'null'.
```

elevatebart:
I removed the update of webpack-dev-server and left the other fixes. 
This is a first step towards compatibility, since we should not drop webpack 3 in the middle of a minor version.

Tested with webpack 3 & webpack 4.
Webpack3, purrs like a kitten.
Webpack4, still has the warning but works.

sapegin:
I believe this was either merged manually of fixed by another pull request. Thanks for the contribution!

PR #1149, opened 2018-09-23
Fix doc issues related to updateExample method

wwwebman:
1.  Wrong path bug:
styleguide.config.js
 ```
const filepath = path.resolve(exampleFilePath, settings.file) 
``` 
Readme.md
``` 
{ "file": "./Basic.jsx" }
```
So,  `absolute/path/to/file/Readme.md/Example.jsx` will be resolved, which is wrong.
To fix it as developer I have to change ./Basic.jsx to ../Basic.jsx, which is wrong :)

Proposition is in doc.

2. ```import``` is not supported in *.md. So, when we wants to add to *.md any  js|jsx example that  use ```import``` `CompileError: import is not supported` will be thrown. 
 
So, for now ```updateExample()``` method should always with be used with ```settings.static = true``` in case of all js|jsx with ```import```. For us as developers it has more sense, otherwise it push us to use 'require' in our components.

I have removed second example, because keep it separately haven't no any sense in my opinion.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -647,8 +647,12 @@ module.exports = {
   updateExample(props, exampleFilePath) {
     const { settings, lang } = props
     if (typeof settings.file === 'string') {
-      const filepath = path.resolve(exampleFilePath, settings.file)
+      const exampleFileDir = path.dirname(exampleFilePath);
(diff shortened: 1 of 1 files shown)


wwwebman:
Ok, thanks.

Issue #1155, opened 2018-10-02
Request: Hide "Props & Methods" button and table for components with no props


teddybradford:
**The problem**

For components that have no props, I don't think there's any reason to show a blank props table nor the "Props & Methods" button.

e.g.:
<img width="946" alt="image" src="https://user-images.githubusercontent.com/3684553/46330126-550b8f80-c5df-11e8-9ca7-41c72c495e82.png">

**Proposed solution**

Hide the "Props & Methods" button and table if the component has no props defined.

**Alternative solution**

Create a config option to automatically hide the "Props & Methods" button and table if the component has no props defined.

**Additional context**

My components are built with TypeScript and I have Styleguidist configured to use the `react-docgen-typescript` propsParser.

Here's an example component I created that has no props, but still causes a "Props & Methods" button and table to be rendered in Styleguidist:

```tsx
import React from "react";

export class Paragraph extends React.Component {
  public render() {
    return <p>Hello, world!</p>;
  }
}
```

sapegin:
Feel free to send a pull request to fix that ;-)

teddybradford:
@sapegin If I can find some time I'll take a stab at it. Do you know if this bug occurs in JS environments as well as TypeScript ones?

sapegin:
I believe it's TypeScript only — likely react-docgen returns different data.

sapegin:
I guess it's just not smart enough, and is triggered by using the `props` name. Maybe we can use `_.isEmpty` here instead of `Array.isArray` and the length check.

teddybradford:
Ah, okay. I'll give that a try.

teddybradford:
It worked! Thanks for suggesting to use `_.isEmpty`. I've made a PR with the changes.

PR #1158, opened 2018-10-03
Improve check for props and methods rendering

teddybradford:
In support of #1155, this PR improves the check for whether props and methods exist in Usage and UsageTabButton components.


sapegin:
Awesome, thanks!

Issue #833, opened 2018-02-20
Add anchor ids to md doc headings


bluetidepro:
*I'll probably look at building it out myself, but wanted to make a record of the idea just in case someone beats me to it.*

Basically, the idea is that anytime in your component docs that you use a header (ex: `## Blah`) it would auto add anchor id's to it so you could ref it in the url. This is a similar concept to what GitHub auto does for `.md` files. For example, I can link to https://github.com/styleguidist/react-styleguidist#usage and it will auto scroll to the `Usage` header in that `.md` file.

To expand, if you had a header in the Button component docs, you could link to it to like https://react-styleguidist.js.org/examples/basic/#!/Button#example4 or something similar **without** going into isolation mode, but still linking to that part of the page in the doc.

We could also look at doing some fancy styling like GitHub has that shows an icon (on hover) of headers to get that url:

![screen shot 2018-02-20 at 12 21 00 pm](https://user-images.githubusercontent.com/410701/36441547-96403db8-1638-11e8-9483-88e640d29eba.png)

What is everyone's thoughts on this? Do you all think it would be worth having this feature? Would it conflict with the current routing at all, by any chance?

seivan:
@bluetidepro You mean, keep the sidebar around when linking to  e.g `http://localhost:6060/#!/Button/1` ?

bluetidepro:
@seivan No, I mean like if you were to go to `https://react-styleguidist.js.org/examples/basic/#!/Button` first, then in the examples you had various headers in the markdown (making the page extremely long possibly), you would be able to add an anchor link to your url (`https://react-styleguidist.js.org/examples/basic/#!/Button#Headername` for example) to auto scroll to that section on the long component docs.

Again, the best example I have is something like this: https://github.com/styleguidist/react-styleguidist#third-party-tools - Where when you click that link it auto scrolls all the way down to the "third-party tools" section because the header has the anchor id on it.

Really, think should be fairly simple to do. Just need to add id's to the component that generates the headers from the markdown and then make sure it works with the routing in place.

seivan:
Oh ok, you want your markdown tags have anchor links generated for them? 
Because there are links right now for each example of a component. 
It just hides the side bar. Not sure if you knew or not.

bluetidepro:
[...]

PR #1163, opened 2018-10-05
Feat: Add generated id to markdown headings (#833)

wkillerud:
This PR introduces preliminary support for linking to headings
within an example text using markdown headings.

Since linking in this way requires using the # character it means you
can't link to a section in isolated mode (for instance to
http://localhost:6060/#!/WrappedButton#example-section) on account of
the #! being reserved for routing.

It may still be useful to the reporter of #833 for linking to a section in the list
view, and may act as a base for further improvements 🙂

### DIFF
--- a/src/client/rsg-components/Markdown/Markdown.spec.js
+++ b/src/client/rsg-components/Markdown/Markdown.spec.js
@@ -23,7 +23,7 @@ describe('Markdown', () => {
 		expectSnapshotToMatch('a [link](http://test.com)');
 	});
 
-	it('should render headings', () => {
+	it('should render headings with generated ids', () => {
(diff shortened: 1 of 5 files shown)


wkillerud:
I don't know the inner workings of the RSG router, but if it's possible to rewrite (long term goal for sure) to free the `#` character using the HTML5 History API it would make this feature much more useful.

Still, even with its limitations this version may add some value :)

wkillerud:
> markdown-to-jsx already passes an ID, so we can use it instead of generating our own

Allright, didn't know that - nice 👍 Made the requested changes and removed the stuff that became redundant.

sapegin:
Thanks, merged!

PR #1165, opened 2018-10-07
Remove webpack 4 warning

elevatebart:
This allows react-styleguidist to run more smoothly on webpack 4 but drops support for webpack 3.
This compatibility drop is necessary if we want to update webpack-dev-server.


elevatebart:
done another way ;)

PR #1171, opened 2018-10-11
Attempt to support inline examples

jgillick:


NigelGreenway:
@jgillick this looks like a great idea, I just wondered if there are any updates on this and when it might merged? 😃

Issue #1169, opened 2018-10-09
Examples not found


DavidHenri008:
**Current behavior**
In some conditions, the examples are not found by Styleguidist.

The examples are found when:
- the component filename is ComponentName.js and the example filename is ComponentName.md
- the component filename is index.js and the example filename is Readme.md

The examples are **not** found when:
- the component filename is index.js and the example filename is ComponentName.md

**To reproduce**
1. Create new project using create-react-app.
2. Add Styleguidist package.
3. Add new basic React component following the following folder structure:
    src -> components -> ComponentName
4. Make sure the component filename is index.js.
5. Add a ComponentName.md example file in the ComponentName folder.
6. Run Styleguidist.

See example [HERE](https://github.com/DavidHenri008/Styleguidist-example).

**Expected behavior**
I am expecting, as stated by the documentation, that the examples to be found if the file is inside the component folder and it is named Readme.md or ComponentName.md.

sapegin:
> When something isn’t working for you, include a reproducible demo:
>  1. Fork the example project: https://github.com/styleguidist/example
>  2. Modify it to reproduce the issue.
>  3. Push to GitHub and paste the link here.
>
> Most likely you’ll find an issue yourself, while creating a demo.
>
> **Issues without a demo will be closed!**

DavidHenri008:
I will add a demo today.
Thanks.

DavidHenri008:
@sapegin I added a link to a fork in the **To reproduce** section of my initial entry.

PR #1185, opened 2018-10-24
Add foldername.md pattern to getExampleFilename when component is ind…

DavidHenri008:
Fix issue [#1169](https://github.com/styleguidist/react-styleguidist/issues/1169)
Update to getExampleFilename in order to detect example file when the component definition file is `index.js` and the examples are in FolderName.md.
Test has been added.
Unfortunately all other tests do not pass but it looks like it was like this when I forked.

### DIFF
--- a/scripts/schemas/config.js
+++ b/scripts/schemas/config.js
@@ -82,16 +82,17 @@ module.exports = {
 		default: componentPath => {
 			const files = [
 				path.join(path.dirname(componentPath), 'Readme.md'),
+				// ComponentName.md
 				componentPath.replace(path.extname(componentPath), '.md'),
+				// FolderName.md when component definition file is index.js
+				path.join(path.dirname(componentPath), path.basename(path.dirname(componentPath)) + '.md'),
 			];
-
 			for (const file of files) {
 				const existingFile = fileExistsCaseInsensitive(file);
 				if (existingFile) {
 					return existingFile;
 				}
 			}
-
 			return false;
 		},
 		example: componentPath => componentPath.replace(/\.jsx?$/, '.examples.md'),
--- a/examples/basic/src/components/Label/Label.md
+++ b/examples/basic/src/components/Label/Label.md
@@ -1,0 +1,7 @@
+Basic Label:
+
+    <Label>Hi there !!!</Label>
+
+Pink background label:
+
+    <Label background="deeppink" color="white">Click Me</Label>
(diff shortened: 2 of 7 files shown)


scripts/schemas/config.js
```diff
@@ -82,16 +82,17 @@ module.exports = {
 		default: componentPath => {
 			const files = [
 				path.join(path.dirname(componentPath), 'Readme.md'),
-				componentPath.replace(path.extname(componentPath), '.md'),
+				// ComponentName.md
+				path.join(componentPath.replace(path.extname(componentPath), '.md')),
```
sapegin: Looks like you don't need `path.join` here.

DavidHenri008: I will remove it.

DavidHenri008:
@sapegin This is my very first PR and I am working on Windows. I have a hard time to complete the PR for two reasons:
1. When I run `npm run test` I am getting many errors that has nothing to do with my modifications. I cloned the master branch in a different repository and I still get the same errors. Do you know if tests (`npm run test`) are successful?
2. Each time I run `npm run test` or `npm start` I am getting 400+ changes in git. But there is no apparent differences. I suspect a crlf/lf issue. But I do not know how to solve this. I am using VSCode as IDE. I think this is the cause of the final differences (since my last commit) that you see in my package-lock.json.

Thanks.

DavidHenri008:
As a note I found out that many tests based on snapshots returns error because of path delimiter, ex:
-       "filepath": "components/Price/Price.js",
+       "filepath": "components\\Price\\Price.js", 
Again I just do not know how to fix this on my VSCode Windows setup.

sapegin:
1. Tests are passing in master — you can see it in the Readme — [Build Status](https://travis-ci.org/styleguidist/react-styleguidist)

    But I'm sure we have some issues on Windows ;-( If you're feeling adventurous and want to help ;-) it would be super cool.

2. It could be because your Git is converting line endings on checkout — https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/

sapegin:
> Again I just do not know how to fix this on my VSCode Windows setup.

The issue with tests is in the repo, not in your setup. But I don't know what could be the right solution here, so any help is appreciated.

sapegin:
I'll update the snapshot, thanks for the work! ;-)

I think removing paths from snapshots is the right way. We can use something like [slash](https://github.com/sindresorhus/slash) to compare paths outside snapshots.

PR #1190, opened 2018-11-02
Use rsg-components import path in Preview to allow for custom ReactExample

emortlock:
Preview uses a relative path to grab `ReactExample` which is causing issues when trying to use an aliased version of it.


sapegin:
Cool, thanks!

PR #1192, opened 2018-11-07
Update Webpack.md to include the react babel preset

MariusMeiners:
If you dont have a .babelrc file and the react preset, webpack is not going to pick up on jsx...

### DIFF
--- a/docs/Webpack.md
+++ b/docs/Webpack.md
@@ -110,6 +110,18 @@ If your project doesn’t use webpack you still need loaders for your files. You
 npm install --save-dev webpack-blocks
 ```
 
+Make sure you add the babel react preset by running
+```bash
+npm install --save-dev @babel/preset-react
+```
+
+Add a .babelrc file in the root of your project and add the preset to it
+```js
+{
+  "presets": ["@babel/react"]
+}
+```
+
 Then add a `webpackConfig` section to your `styleguide.config.js`:
 
 ```javascript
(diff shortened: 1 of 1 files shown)


MariusMeiners:
you can recreate this behaviour by forking my [template](https://github.com/MariusMeiners/componentLibraryTemplate) and removing the .babelrc file

MariusMeiners:
@sapegin why does it then not work when deleting the .babelrc file ? Did you try and reproduce the problem ?

sapegin:
Because your code requires it, not Styleguidist or React.

MariusMeiners:
@sapegin i would appreciate it if you could explain that in a little more detail. I felt like i followed the instructions on the website 100% and it still didnt work. What did i do differently and where did i require the babel preset in my code ?

MariusMeiners:
do you mean when i use any es6 i have to pump it through babel or what do you mean exactly ? i thought your webpack config does that automatically by getting babel from webpack-blocks

sapegin:
Yup, if you use JSX or any other features not supported by browsers in _your_ code, you need to setup Babel, TypeScript or whatever you like in _your_ webpack config. But none of this is _required_ to use Styleguidist or React. Styleguidist doesn't do any changes to your code. If you use webpack-blocks, it's, again, in your webpack config.

I agree that this isn't super clear and I agree that Babel is common for React projects. So I'll be happy if you make the docs better.

MariusMeiners:
Thanks for clarifying that ! I updated the readme a little but im not sure the wording is optimal. Maybe you can take a look at it and change it to indicate that in order to use jsx and other features configuring the toolchain is mandatory. You can close the issue if you want to :)

sapegin:
Please have a look at #1193 — is this more clear now?

PR #1193, opened 2018-11-08
Docs: Make webpack config docs more clear

sapegin:
Closes #1192.

### DIFF
--- a/docs/Webpack.md
+++ b/docs/Webpack.md
@@ -60,7 +60,7 @@ module.exports = {
   webpackConfig: {
     module: {
       rules: [
-        // Babel loader, will use your project’s .babelrc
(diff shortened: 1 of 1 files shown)


sapegin:
None of this is _mandatory_ ;-) but env + react presets is a very common setup.

The problem with webpack-blocks is that the stable release doesn't work with Babel 7.

MariusMeiners:
@sapegin i think in the new webpack version the syntax is:
```js
  module: {
    rules: [
      { test: /\.jsx?$/, exclude: /node_modules/, loader: 'babel-loader' },
    ]
  }
```
so `rules` instead of `loaders` ?

sapegin:
Yeah, good point!

Issue #1186, opened 2018-10-25
Highlight currently selected page on sidebar


jkarttunen:
**The problem**

On styleguides with a lot of content, it easy to get lost on where you were on the sidebar.

**Proposed solution**

A minimum useful solution would be highlighting (e.g. using background color) the currently selected page.

**Alternative solutions**

Related solutions would be 
- Supporting collapsible sections to enable better browsing of the main sections only.
- Scrolling the sidebar to right position when user arrives to the page using direct link.

sapegin:
Feel free to send a pull request for that! ;-)

cj:
@jkarttunen @sapegin according to this pull request https://github.com/styleguidist/react-styleguidist/pull/1195 it should be highlighting the sidebar.  Did this merged pull request get removed at some point?

jure:
There's something funky going on on line https://github.com/styleguidist/react-styleguidist/blob/master/src/client/rsg-components/ComponentsList/ComponentsListRenderer.js#L51. Probably the link/href format has changed since then? You can temporarily fix this bit by overwriting the ComponentsListRenderer and changing this bit of the functionality.

nathanmarks:
are you guys using `pagePerSection`? it doesn't seem to work with that option properly

KesselRun-xyz:
Related: https://github.com/styleguidist/react-styleguidist/issues/1375

jkarttunen:
@nathanmarks Yep, that was it.

jkarttunen:
Closing as a duplicate of #1375

PR #1195, opened 2018-11-10
Highlight currently selected page on sidebar

rfbotto:
https://github.com/styleguidist/react-styleguidist/issues/1186

Implemented check on `ComponentsListRenderer` for the current `window.location.hash` in order to verify which one of the items is currently selected. Added a  `background-color` change to the selected item, and also added some padding to the item to make the highlight selection better due to the previous lack of padding.

kapture 2018-11-10

### DIFF
--- a/src/rsg-components/ComponentsList/ComponentsListRenderer.js
+++ b/src/rsg-components/ComponentsList/ComponentsListRenderer.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 import Link from 'rsg-components/Link';
 import Styled from 'rsg-components/Styled';
+import { hasInHash, getHash } from '../../utils/handleHash';
 
 const styles = ({ color, fontFamily, fontSize, space, mq }) => ({
 	list: {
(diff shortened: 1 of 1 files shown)


src/rsg-components/ComponentsList/ComponentsListRenderer.js
```diff
@@ -42,21 +47,30 @@ export function ComponentsListRenderer({ classes, items }) {
...
-				<li
-					className={cx(classes.item, (!content || !content.props.items.length) && classes.isChild)}
-					key={href}
-				>
-					<Link
-						className={cx(heading && classes.heading)}
-						href={href}
-						target={external ? '_blank' : undefined}
+			{items.map(({ heading, visibleName, href, content, external }) => {
+				const windowHash = window.location.pathname + getHash(window.location.hash);
```
sapegin: I don't think you need to do this in each loop iteration.

rfbotto:
Yes, you are right, completely overlooked the unnecessary calculation inside of the loop. Also, updated the code to have the selected item be bold instead of changing the background color.  Think it might be better like this too.

kapture 2018-11-12

sapegin:
Thanks!

PR #1196, opened 2018-11-11
Fix: [Docs] config for CRA with typescript includes webpack version that is incompatible

### LINKED ISSUE #1134: [Docs] config for CRA with typescript (react-scripts-ts) includes webpack version that is incompatible
Started a new create-react-app project with react-scripts-ts and noticed that the [installation guide](https://react-styleguidist.js.org/docs/webpack.html#create-react-app-with-typescript) says to use the following config:

```javascript
module.exports = {
  propsParser: require('react-docgen-typescript').parse,
  webpackConfig: require('react-scripts-ts/config/webpack.config.dev')
}
```

The problem with the above config is that react-scripts-ts@next still uses webpack 3.8.1 while styleguidist currently uses webpack 4.15.1. This probably should be removed from the docs or point to a relevant webpack config package for typescript.

rasheedhamdawi:


rasheedhamdawi:
Done, thanks

sapegin:
Do we still need this line by the way? Styleguidist should look for CRA webpack config by default.

PR #1211, opened 2018-11-21
Removed <Code> styles from Name and Type renderers

### LINKED ISSUE #1205: Styling text within back-ticks also affects the component props' styles
They should have different styles, so that the props will not show a border around them.

screen shot 2018-11-19 at 16 20 42

screen shot 2018-11-19 at 16 21 02

In `styleguide.config.js`:

```
styles: {
    Code: {
      code: {
        fontSize: '14px',
        borderRadius: '5px',
        border: '1px solid #BCBEC0',
        background: '#F1F3F5',
        padding: '2px'
      }
    }
  },
```

gazpachu:
This should fix #1205

### DIFF
--- a/src/rsg-components/Name/NameRenderer.js
+++ b/src/rsg-components/Name/NameRenderer.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Code from 'rsg-components/Code';
 import Styled from 'rsg-components/Styled';
 import cx from 'classnames';
 
-export const styles = ({ fontSize, color }) => ({
+export const styles = ({ fontFamily, fontSize, color }) => ({
 	name: {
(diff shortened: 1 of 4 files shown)


src/rsg-components/Name/NameRenderer.js
```diff
@@ -19,11 +19,7 @@ export function NameRenderer({ classes, children, deprecated }) {
 	const classNames = cx(classes.name, {
 		[classes.isDeprecated]: deprecated,
 	});
-	return (
-		<span className={classNames}>
-			<Code>{children}</Code>
-		</span>
-	);
+	return <span className={classNames}>{children}</span>;
```
sapegin: Let's use `<code>` instead of span here.

sapegin:
Thanks!

PR #1215, opened 2018-11-22
Remove <Code> styles from props

gazpachu:
Something I missed in https://github.com/styleguidist/react-styleguidist/pull/1211

### DIFF
--- a/src/rsg-components/Props/PropsRenderer.js
+++ b/src/rsg-components/Props/PropsRenderer.js
@@ -4,7 +4,6 @@ import Group from 'react-group';
 import objectToString from 'javascript-stringify';
 import Arguments from 'rsg-components/Arguments';
 import Argument from 'rsg-components/Argument';
-import Code from 'rsg-components/Code';
 import JsDoc from 'rsg-components/JsDoc';
 import Markdown from 'rsg-components/Markdown';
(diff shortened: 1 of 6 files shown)


src/rsg-components/Props/PropsRenderer.js
```diff
@@ -69,7 +68,7 @@ function renderEnum(prop) {
 	}
 
 	const values = getType(prop).value.map(({ value }) => (
-		<Code key={value}>{showSpaces(unquote(value))}</Code>
+		<code key={value}>{showSpaces(unquote(value))}</code>
```
sapegin: This will use default browser styles or any styles defined in your app stylesheet.

sapegin: I'd actually make [Table](https://github.com/styleguidist/react-styleguidist/blob/master/src/rsg-components/Table/TableRenderer.js) component to support different types of columns — code, type, name, etc. This will simplify this component a bit and reduce number of DOM nodes.

sapegin: So you'll define styles per column here:

```js
export const columns = [
	{
		caption: 'Prop name',
		render: renderName,
                 type: 'code'
	},
```

PR #1233, opened 2018-12-17
Fix basic example to make install instructions work

### LINKED ISSUE #1232: Imports in the basic example don't work
I was trying to convert my repo to 9.0.0-beta3, but couldn't figure out how to import other components in the scope. So I tried to use the example and that gave the same error.

**Current behavior**
The examples where an import takes place error out in the basic example.
errors

**To reproduce**
```
git clone https://github.com/styleguidist/react-styleguidist.git
cd react-styleguidist/examples/basic
npm install
npm install core-js # this seems to be depended on but wasn't in the package.json (other bug)
npx styleguidist server
```

**Expected behavior**
None of the examples have an error.

JelteF:
This fixes #1232 (instead of just closing and locking it). The main thing that fixes the install is deleting package-lock.json and regenerating it. I tried investigating further why that helped and it seems to becease two different acorn versions are installed and the main one should be 6.0.4, not 5.7.1. Otherwise you get this weird import error.

### DIFF
--- a/examples/basic/package-lock.json
+++ b/examples/basic/package-lock.json
@@ -13,18 +13,18 @@
       }
     },
     "@babel/core": {
-      "version": "7.1.6",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.1.6.tgz",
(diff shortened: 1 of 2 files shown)


sapegin:
Thanks!

We use Acorn 6 directly, and it's also a dependency of Buble that should be fixed to the version that users version 6 as well. Other than that Acorn is used by ESLint and probably something else.

PR #1239, opened 2018-12-29
fix: Incorrect current section highlighting on partial match closes #1237 

### LINKED ISSUE #1237: Incorrect current section highlighting on partial match
**Current behavior**

Several sections are highlighted when the current section contains part of a name of a section in the sidebar.

**To reproduce**

image 2018-12-24 at 7 57 53 pm

**Expected behavior**

Only one section is highlighted.

guyius:
### Background
* Usage of the hasInHash function to determine the current section to highlight on the sidebar causes partial matches to be highlighted as well. This is because the hasInHash function uses indexof.

### Changes
* Removed usage of hasInHash and its import in the ComponentsListRenderer.
* Compare string in order to determine the right section to highlight. similar to before I compare the windowHash with the current item href.


sapegin:
Thanks!

PR #1241, opened 2019-01-07
Fix webpack config file name for react-scripts > 2.1.1

oterral:
Since react-scripts 2.1.1, the config file is named webpack.config.js instead of webpack.config.dev.js.

### DIFF
--- a/src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
+++ b/src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
@@ -3,15 +3,16 @@ import findUserWebpackConfig from '../findUserWebpackConfig';
 const cwd = process.cwd();
 afterEach(() => process.chdir(cwd));
 
-it('should return path to Create React App Webpack dev config', () => {
+it('should return path to Create React App Webpack old config (react-scripts <= 2.1.1)', () => {
 	const result = findUserWebpackConfig(a => a);
 	expect(result).toMatch(/^react-scripts\/config\/webpack\.config\.dev/);
 });
 
-it('should return path to Create React App Webpack config', () => {
+it('should return path to Create React App Webpack config (react-scripts > 2.1.1)', () => {
 	const result = findUserWebpackConfig(a => {
 		if (/webpack\.config\.dev/.test(a)) {
-			throw window.Exception();
+			// Simulate an error. For example, if the file doesn't exist.
+			throw new Error();
 		}
 		return a;
 	});
--- a/src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
+++ b/src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
@@ -5,7 +5,7 @@ afterEach(() => process.chdir(cwd));
 
 it('should return path to Create React App Webpack old config (react-scripts <= 2.1.1)', () => {
 	const result = findUserWebpackConfig(a => a);
-	expect(result).toMatch(/^react-scripts\/config\/webpack\.config\.dev/);
+	expect(result).toMatchInlineSnapshot(`"react-scripts/config/webpack.config.dev"`);
 });
 
 it('should return path to Create React App Webpack config (react-scripts > 2.1.1)', () => {
@@ -16,7 +16,7 @@ it('should return path to Create React App Webpack config (react-scripts > 2.1.1
 		}
 		return a;
 	});
-	expect(result).toMatch(/^react-scripts\/config\/webpack\.config\.js/);
+	expect(result).toMatchInlineSnapshot(`"react-scripts/config/webpack.config"`);
 });
 
 it('should return an absolute path to user Webpack config located in project root folder', () => {
--- a/src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
+++ b/src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
@@ -16,7 +16,7 @@ it('should return path to Create React App Webpack config (react-scripts > 2.1.1
 		}
 		return a;
 	});
-	expect(result).toMatch(/^react-scripts\/config\/webpack\.config/);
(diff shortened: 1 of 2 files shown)


joeycozza:
I'm not familiar with this codebase and where findUserWebpackConfig() is called, but an important note is that the new webpack.config.js file exports a function that expects a webpackEnv variable of either "developement" or "production" to be passed in. https://github.com/facebook/create-react-app/commit/30ee52cf3b2cbb6ac70999c02b1196bcaba8d4ca#diff-dc0c4e7c623b73660da1809fc60cf6baR55

ie. webpack.config.js is more of a configFactory than an actual config itself.

Their codebase had to change in how they use webpack.config.js here. https://github.com/facebook/create-react-app/commit/30ee52cf3b2cbb6ac70999c02b1196bcaba8d4ca#diff-d231b8cdf8c6b73068083e851c3a7c5bR48

I don't know if this information would change this PR or not, but I thought it would be worth mentioning.

sapegin:
Please have a look at the CI results.

oterral:
@sapegin The build passed before I add tests. I m not sure how unit tests can breaks integration tests.
I need help on this.

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
```diff
@@ -3,9 +3,19 @@ import findUserWebpackConfig from '../findUserWebpackConfig';
 const cwd = process.cwd();
 afterEach(() => process.chdir(cwd));
 
-it('should return path to Create React App Webpack config', () => {
+it('should return path to Create React App Webpack dev config', () => {
```
sapegin: Dev isn't description — it's an old version of the same config.

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
```diff
@@ -3,9 +3,19 @@ import findUserWebpackConfig from '../findUserWebpackConfig';
...
 	const result = findUserWebpackConfig(a => a);
-	expect(result).toMatch(/^react-scripts\//);
+	expect(result).toMatch(/^react-scripts\/config\/webpack\.config\.dev/);
+});
+
+it('should return path to Create React App Webpack config', () => {
+	const result = findUserWebpackConfig(a => {
+		if (/webpack\.config\.dev/.test(a)) {
+			throw window.Exception();
```
sapegin: What is `window.Exception` and why it's necessary here?

src/scripts/utils/findUserWebpackConfig.js
```diff
@@ -1,7 +1,10 @@
 const fs = require('fs');
 const path = require('path');
 
-const CREATE_REACT_APP_WEBPACK_CONFIG = 'react-scripts/config/webpack.config.dev';
+// react-scripts <= 2.1.1
+const CREATE_REACT_APP_WEBPACK_CONFIG_DEV = 'react-scripts/config/webpack.config.dev';
```
sapegin: Same about `DEV` naming.

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
oterral: yes. I fix

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
oterral: I meant `Error`. Exception doesn't exist ;) .  I force  to to go in the try catch to test the loading of the new webpack.config.

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
oterral: Fixed

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
oterral: Fixed

src/scripts/utils/findUserWebpackConfig.js
oterral: Fixed

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
sapegin: Can you add a comment about that? It's not very obvious.

oterral: already done : https://github.com/styleguidist/react-styleguidist/pull/1241/files#diff-247803435e192ad771c19314144c7ce2R14

src/scripts/utils/__tests__/findUserWebpackConfig.spec.js
```diff
@@ -3,9 +3,20 @@ import findUserWebpackConfig from '../findUserWebpackConfig';
...
+
+it('should return path to Create React App Webpack config (react-scripts > 2.1.1)', () => {
+	const result = findUserWebpackConfig(a => {
+		if (/webpack\.config\.dev/.test(a)) {
+			// Simulate an error. For example, if the file doesn't exist.
+			throw new Error();
+		}
+		return a;
+	});
+	expect(result).toMatch(/^react-scripts\/config\/webpack\.config/);
```
sapegin: This will also match `webpack.config.dev`. Probably should be something like `/^react-scripts\/config\/webpack\.config.js$/`.

sapegin:
Looks like there's no JS at the end. Actually, since it's not a real `path.resolve` we can just use `.toBe()` or `.toMatchInlineSnapshot` — no need for regexps in these tests.

```
    Expected value to match:
      /^react-scripts\/config\/webpack\.config\.js/
    Received:
      "react-scripts/config/webpack.config"
```

oterral:
Didn't know about `.toMatchInlineSnapshot` . Seems a good way to go.

PR #1245, opened 2019-01-11
Docs: Fix webpack css loader example

yhabib:
Following the docs I was trying to load a 3rd party css library by requiring it:
`require: [path.resolve(__dirname, "node_modules/react-dates/lib/css/_datepicker.css")],`

and adding the corresponding loader as it is shown [here](https://react-styleguidist.js.org/docs/webpack.html#custom-webpack-config) but didn't work out.
```js
{
  test: /\.css$/,
  loader: 'style-loader!css-loader?modules'
}
```

Changing `loader` for `use` did the trick.
```js
{
  test: /\.css$/,
  use: ['style-loader', 'css-loader'] 
}
```

### DIFF
--- a/docs/Webpack.md
+++ b/docs/Webpack.md
@@ -69,7 +69,7 @@ module.exports = {
         // Other loaders that are needed for your components
         {
           test: /\.css$/,
-          loader: 'style-loader!css-loader?modules'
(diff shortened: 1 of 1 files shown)


sapegin:
Thanks!

PR #1248, opened 2019-01-13
Chore: Update dependencies

lukeed:
Hey there,

As discussed with @sapegin, here are a few updates to the dependency tree. In many cases, it was a simple find & replace. I also took care of the annoying deprecation message 😉 

Feel free to pick-n-choose and/or reject as desired. I'm happy with whatever~!

The only lingering TODOs are regarding the flow-types for classnames & minimist. I don't use that ecosystem, so I'm not sure where those come from, especially given that they look to be official definitions. I did, however, update the `qss` definition as that one was pretty straight forward.

Thanks!


src/bin/styleguidist.js
```diff
@@ -27,7 +27,11 @@ process.on('uncaughtException', err => {
 			consts.DOCS_CONFIG
 		);
 	} else if (err instanceof StyleguidistError) {
-		console.error(kleur.bold.red(err.message));
+		console.error(
+			kleur()
```
lukeed: This multi-line business was prettier – sorry

lukeed:
Great, no problem! 🙌 

If you switch to TS, both `qss` and `clsx` include their own typings. As of now, `mri` does not but I can probably work on it myself (tho I don't TS) or petition for help.

If you stick with flow, it looks like there's a flow-typed PR open for `clsx` definitions 💪 

Either way, the typings/flow-type for `mri` will be the same as `minimist` (for your usage), and `clsx` would work perfectly with the `classnames` flow-types (for your usage, since not calling dedupe/whatever).

If you need/want something else from me, just lemme know~!

Issue #1201, opened 2018-11-14
React Native example- The node type SpreadProperty has been renamed to SpreadElement


mcconkiee:
**Current behavior**
when running the demo for react-native i get  the following several times over
```
Trace: The node type SpreadProperty has been renamed to SpreadElement
    at Object.isSpreadProperty (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/types/lib/validators/generated/index.js:4192:11)
    at hasSpread (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/babel-plugin-transform-object-rest-spread/lib/index.js:38:13)
    at PluginPass.ObjectExpression (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/babel-plugin-transform-object-rest-spread/lib/index.js:234:14)
    at newFn (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/visitors.js:193:21)
    at NodePath._call (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/path/context.js:88:12)
    at TraversalContext.visitQueue (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/context.js:118:16)
    at TraversalContext.visitSingle (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/context.js:90:19)
    at TraversalContext.visit (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/context.js:146:19)
    at Function.traverse.node (/Users/ericmcconkie/Downloads/react-styleguidist-7.3.11/examples/react-native/node_modules/@babel/traverse/lib/index.js:94:17)
[...]

PR #1260, opened 2019-02-01
Update react-native example

mitsuruog:
This PR is for fixing broken react-native example.

I decided to recreate example code from scratch from Expo. 
Because the previous example had generated by create-react-native-app, but that project had moved to Expo.

I think it would be more worth to transfer to Expo than to keep maintaining old projects for this project.

NOTE: 
Since English is not my first language, so you can freely change README.md when you find any grammar issue or something else.

I am happy if I can make the first contribution!

Related issue:

- React Native example- The node type SpreadProperty has been renamed to SpreadElement #1201


sapegin:
Thanks!

PR #1270, opened 2019-02-07
Upgrade react-group to v3 

### LINKED ISSUE #1265: Upgrade react-group
[New major release](https://github.com/sapegin/react-group/releases/tag/v3.0.0) has a few breaking changes but won't generate an unnecessary DOM node which is good.

- [ ] Update dependency in `package.json`
- [ ] Update and test all usages of `react-group` according to release notes

dougmackenzie:
Closes #1265 

First contribution to open source ever :)

### DIFF
--- a/src/client/rsg-components/Argument/ArgumentRenderer.js
+++ b/src/client/rsg-components/Argument/ArgumentRenderer.js
@@ -18,28 +18,32 @@ export function ArgumentRenderer({ classes, name, type, description, returns, bl
 	if (isOptional) {
 		type = type.expression;
 	}
-	return (
-		<div className={block && classes.block}>
-			<Group>
-				{returns && 'Returns'}
-				{name && (
-					<span>
-						<Name>{name}</Name>
-						{type && ':'}
-					</span>
-				)}
-				{type && (
-					<Type>
-						{type.name}
-						{isOptional && '?'}
-						{!!defaultValue && `=${defaultValue}`}
-					</Type>
-				)}
-				{type && description && ` — `}
-				{description && <Markdown text={`${description}`} inline />}
-			</Group>
-		</div>
+	const content = (
+		<Group>
+			{returns && 'Returns'}
+			{name && (
+				<span>
+					<Name>{name}</Name>
+					{type && ':'}
+				</span>
+			)}
+			{type && (
+				<Type>
+					{type.name}
+					{isOptional && '?'}
+					{!!defaultValue && `=${defaultValue}`}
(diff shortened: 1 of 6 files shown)


src/client/rsg-components/Argument/ArgumentRenderer.js
```diff
@@ -19,24 +19,26 @@ export function ArgumentRenderer({ classes, name, type, description, returns, bl
...
-					{isOptional && '?'}
-					{!!defaultValue && `=${defaultValue}`}
-				</Type>
-			)}
-			{type && description && ` — `}
-			{description && <Markdown text={`${description}`} inline />}
-		</Group>
+		<div className={block && classes.block}>
+			<Group {...props}>
+				{returns && 'Returns '}
```
sapegin: A space after `Returns` shouldn't be necessary.

src/client/rsg-components/Argument/ArgumentRenderer.js
```diff
@@ -19,24 +19,26 @@ export function ArgumentRenderer({ classes, name, type, description, returns, bl
...
-					{type.name}
-					{isOptional && '?'}
-					{!!defaultValue && `=${defaultValue}`}
-				</Type>
-			)}
-			{type && description && ` — `}
-			{description && <Markdown text={`${description}`} inline />}
-		</Group>
+		<div className={block && classes.block}>
+			<Group {...props}>
```
sapegin: And spread here is also unnecessary.

src/client/rsg-components/Argument/ArgumentRenderer.js
```diff
@@ -19,24 +19,26 @@ export function ArgumentRenderer({ classes, name, type, description, returns, bl
 		type = type.expression;
 	}
 	return (
```
sapegin: I think we can restructure it like this:

```js
const content = <Group>...</Group>;

if (block) {
  return <div className={classes.block}>{content}</div>
}
return content;
```

to avoid rendering an unnecessary `div` when it's not `block`.

src/client/rsg-components/Argument/ArgumentRenderer.js
dougmackenzie: Oops, the whole purpose of using <Group>. I've also removed the spaces from the em dash below.

dougmackenzie:
Thanks for the review! Should all be fixed up now.

sapegin:
Cool, thanks!

PR #1273, opened 2019-02-11
Fix typo leading to 404

gnarmedia:


sapegin:
This works when you follow the instruction from the beginning: `styleguidist` is a binary name, but `react-styleguidist` package must be installed in your project.

PR #1277, opened 2019-02-15
WIP: Allow display of shape prop recursively

oterral:
This PR displays  documentation of shapes recursively like this:

recursive_shape_display

 I will be happy to hear if there is a better way to achieve this.

### DIFF
--- a/src/client/rsg-components/Props/PropsRenderer.js
+++ b/src/client/rsg-components/Props/PropsRenderer.js
@@ -78,14 +78,19 @@ function renderEnum(prop) {
 	);
 }
 
-function renderShape(props) {
-	const rows = [];
+function renderShape(props, level = 0) {
+	let rows = [];
 	for (const name in props) {
 		const prop = props[name];
 		const defaultValue = renderDefault(prop);
 		const description = prop.description;
 		rows.push(
-			<div key={name}>
+			<div
+				key={name}
+				style={{
+					paddingLeft: 30 * level + 'px',
+				}}
+			>
 				<Name>{name}</Name>
 				{': '}
 				<Type>{renderType(prop)}</Type>
@@ -95,6 +100,7 @@ function renderShape(props) {
 				{description && <Markdown text={description} inline />}
 			</div>
 		);
+		rows = [...rows, renderExtra({ type: prop }, level + 1)];
 	}
 	return rows;
 }
@@ -174,7 +180,7 @@ function renderDescription(prop) {
 	);
 }
 
-function renderExtra(prop) {
+function renderExtra(prop, level = 0) {
 	const type = getType(prop);
 	if (!type) {
 		return null;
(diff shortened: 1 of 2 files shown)


src/client/rsg-components/Props/PropsRenderer.js
```diff
@@ -78,14 +78,20 @@ function renderEnum(prop) {
...
+	const nextLevel = ++level;
 	for (const name in props) {
 		const prop = props[name];
 		const defaultValue = renderDefault(prop);
 		const description = prop.description;
 		rows.push(
-			<div key={name}>
+			<div
+				key={name}
+				style={{
```
sapegin: Please use JSS instead of inline styles.

src/client/rsg-components/Props/PropsRenderer.js
```diff
@@ -95,6 +101,9 @@ function renderShape(props) {
 				{description && <Markdown text={description} inline />}
 			</div>
 		);
+		// Display more shapes
+		prop.type = prop;
+		rows = [...rows, renderExtra(prop, nextLevel)];
```
sapegin: Use can avoid reassigning here: `rows.push(...renderExtra(prop, nextLevel)`.

src/client/rsg-components/Props/PropsRenderer.js
```diff
@@ -78,14 +78,20 @@ function renderEnum(prop) {
 	);
 }
 
-function renderShape(props) {
-	const rows = [];
+function renderShape(props, level = 0) {
+	let rows = [];
+	const nextLevel = ++level;
```
sapegin: Avoid mutating function arguments. I guess you can use `level + 1` when you call `renderExtra`.

src/client/rsg-components/Props/PropsRenderer.js
sapegin: And instead of mutating the original `props` object you could do:

```js
renderExtra({...prop, type: prop}, level + 1)
```

Or even:

```js
renderExtra({type: prop}, level + 1)
```

oterral: Of course!

src/client/rsg-components/Props/PropsRenderer.js
oterral: I would be happy :) but how ? can you give me a hint? I 'm not sure how to do.

sapegin: There should be enough examples in the codebase:

https://github.com/styleguidist/react-styleguidist/blob/master/src/client/rsg-components/Argument/ArgumentRenderer.js#L9-L13
https://react-styleguidist.js.org/docs/development.html#styles-1
https://cssinjs.org/react-jss/?v=v10.0.0-alpha.10#dynamic-values

oterral: Should I create a ShapeRender ?  or should I pass  `classes` property from the PropsRenderer to the Table renderer and then to rednerShape function ?

oterral: I didn't succeed to use dynamic padding, it always take the same for every component have a look to this [commit](https://github.com/oterral/react-styleguidist/commit/448b87aa322c23e35e7db156813f97cf43060ffb).

If you have an idea.

Issue #1292, opened 2019-02-27
Docs show config options that appear to have been removed


arizonatribe:
**Current behavior**

While working with the Node API for React Styleguidist, I noticed a couple of times where I was receiving the "Something is wrong with your style guide config", but no other info. Tracing that [in the code](https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/config.js#L49) it looks like there's just situations where there is no `exception.message` and `exception.extra` to accompany the message. Locally I added a `console.log` to that file in my `node_modules/react-styleguidist/` location to see if there's anything useful on the `exception` itself, and it gave the full information on the field name that it didn't like.

```
Unknown config option moduleAliases was found, the value is: // rest of output
```

To be clear, I'm not opening the issue for that unsupported field. It looks like there was a hiccup in the support for that field: seems to have been removed from the `schema/config.js` in the version 8.0.6 I was using, but I see it's back now in 9.0.1 (one oddity still is [the docs still have it listed on the page](https://react-styleguidist.js.org/docs/configuration.html#modulealiases), just not among the table-of-contents on [the top of the page](https://react-styleguidist.js.org/docs/configuration.html)).

The reason for the issue is the logging of the error seems like it could be improved (or fixed, depending on your opinion):

```js
if (exception instanceof StyleguidistError) {
  throw new StyleguidistError(
    'Something is wrong with your style guide config',
    exception.message,
    exception.extra
  );
}
```
That's the current code (in the `scripts/config.js`) but `exception.message` doesn't seem to exist on [the `StyleguidistError` object](https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/utils/error.js). The `extra` prop _is_ part of that custom error, but the "message" is just used to feed into the underlying (native) `Error`.

There's really nothing to catch the error and pull the `.extra` prop off of it, unless the user _knows_ to do so. It seems like it makes more sense that if you have the reason for the error, to show that instead as the message.

```js
if (exception instanceof StyleguidistError) {
  throw new StyleguidistError(
    exception.message || 'Something is wrong with your style guide config',
    exception.extra
  );
}
```

Thanks!

sapegin:
The `moduleAliases` was introduced in 9.x, so we need to update the TOC in the docs ;-/

Does it work fine with `exception.toString()`? I'm fine with any solution if it improves the experience.

arizonatribe:
[...]

Issue #1291, opened 2019-02-27
Unable to use Node API in v9.0.1


arizonatribe:
**Current behavior**

I'm not sure what happened between 8.0.6 and 9.0.1, but when I updated today I'm no longer able to import certain files. At first I figured there was just some shuffling of files, but I believe it's more than that.

Following [the docs on using the Node API](https://react-styleguidist.js.org/docs/api.html) I should be able to see (at minimum) this working:

```js
const styleguidist = require('react-styleguidist');
```
But that fails now.

Looking at the `node_modules/react-styleguidist/package.json` in my project is see this for the `main` prop in `package.json`

```json
  "main": "scripts/index.js",
```

That seems to be incorrect (the `scripts/` directory is inside of `lib/` when react-styleguidist is installed into a given app).

To make this work currently I'm having to do this instead:

```js
const styleguidist = require('react-styleguidist/lib/scripts');
```

**To reproduce**

```
mkdir some-app && cd some-app
npm init -y
npm install --save-dev react-styleguidist
node
```
Then - from inside the Node R.E.P.L. - do this:
```
> require.resolve('react-styleguidist')
Error: Cannot find module 'react-styleguidist'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.resolve (internal/modules/cjs/helpers.js:30:19)

```
However, if I know what file to pull in, it's fine:
```
> require.resolve('react-styleguidist/lib/client/rsg-components/Styled')
'/Users/dnunez/projects/carvana/test/node_modules/react-styleguidist/lib/client/rsg-components/Styled/index.js'
```

Alternatively, if you use the (awesome) library called [replem](https://www.npmjs.com/package/replem) you can shorten the steps-to-reproduce:

```
npm install --global replem
replem react-styleguidist
```
Which throws this error:
```
Installed into REPL context:
 - react-styleguidist@9.0.1 as reactStyleguidist
> Cannot find module '/Users/dnunez/.replem/node_modules/react-styleguidist'
```

sapegin:
Feel free to send a pull request with a fix! ;-)

arizonatribe:
Really? That's all you're gonna give me? So is it a bug even, or is that like, my opinion man?

PR #1293, opened 2019-02-27
Fixes for main prop and Node API exception handling

arizonatribe:
Fixed main prop (in package.json) and Node API exception handling (for StyleguidistError instances)

#1292 
#1291

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "url": "https://github.com/styleguidist/react-styleguidist/issues"
   },
   "license": "MIT",
-  "main": "scripts/index.js",
+  "main": "lib/scripts/index.js",
   "bin": {
     "styleguidist": "lib/bin/styleguidist.js"
   },
--- a/src/scripts/__tests__/config.spec.js
+++ b/src/scripts/__tests__/config.spec.js
@@ -46,7 +46,7 @@ it('should throw if config has errors', () => {
 			components: 42,
 		});
 	} catch (err) {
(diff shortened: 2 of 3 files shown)


sapegin:
Please revert version number changes and I'll merge it — looks good otherwise.

arizonatribe:
Okay should be all set now

sapegin:
Could you please test that this code still shows a correct link with an anchor?

https://github.com/styleguidist/react-styleguidist/blob/2bea3d6bd14a06d85ee590d9e08206e2e2f973cf/src/bin/styleguidist.js#L51-L57

arizonatribe:
All your tests passed; what do you mean?

arizonatribe:
I just went through all the source code and I can't even find anywhere where you _are_ passing a link/anchor into a new `StyleguidistError`.

arizonatribe:
Are you trying to indicate that when you're creating a new `StyleguidistError` from another one (in the `scripts/config.js`) that the `.extra` from the original error is supposed to map to the `.anchor` on the new error?

If that's the case, that seems a bit hacky and brittle. This would make more sense:
```js
if (exception instanceof StyleguidistError) {
  throw new StyleguidistError(
    'Something is wrong with your style guide config',
    exception.toString(),
    exception.anchor
  );
}
```

sapegin:
I think I was able fo fix both issues 🤔

arizonatribe:
You asked about the anchor/link, but the fix you did on your own doesn't pass that through to the constructor of the next error. Does that make sense?

PR #1296, opened 2019-02-28
Allow trailing comma in named imports in examples

### LINKED ISSUE #1295: Examples don't support named imports with trailing comma
Hi! I'm in the process of upgrading to version `9.0.1` and encountered this odd issue.

**Current behavior**

An example importing multiple exports from the same file while using a trailing comma in the import statement will break the example.

 ```
import {
  Button,
  Placeholder,
} from 'rsg-example/components';
```

Chrome produces the following error:

`SyntaxError: Unexpected token =`

Firefox produces the following error:

`SyntaxError: missing variable name`

**To reproduce**

I modified the [basic example in a branch on my fork](https://github.com/wkillerud/react-styleguidist/tree/trailing-comma-in-import-breaks-example/examples/basic/src/components) to demonstrate the problem scenario. The [diff is here](https://github.com/wkillerud/react-styleguidist/commit/97b3f0f40f902e19baf20a288afce3beb7387103).

I put a breakpoint in `ReactExample.getExampleComponent` and see the problem in the `compiledCode`:

```
"const react$0 = require('react');
const React = react$0.default || react$0;
const components$0 = require('rsg-example/components');
const Button = components$0.Button;
const Placeholder = components$0.Placeholder;
const  = components$0.;
;React.createElement( Button, null,
  React.createElement( Placeholder, null )
)"
```

`const  = components$0.;`

Removing the trailing comma produces the expected result.

**Expected behavior**

Trailing commas in import statements should not break examples. Some linters and Prettier configs enforce this trailing comma when the number of imports goes above a certain threshold, even in Markdown files (Pretties does this in my current project).

### PR DESCRIPTION
**Pull request by wkillerud**
Fixes #1295

### DIFF
--- a/src/client/utils/__tests__/transpileImports.spec.js
+++ b/src/client/utils/__tests__/transpileImports.spec.js
@@ -57,6 +57,34 @@ const E = snake$0.default || snake$0;
 `);
 	});
 
+	test('transpile multiline named imports without trailing comma', () => {
+		const result = transpileImports(`import {
+  B,
+  C
+} from 'cat'
+`);
+		expect(result).toMatchInlineSnapshot(`
+"const cat$0 = require('cat');
+const B = cat$0.B;
+const C = cat$0.C;
+"
+`);
+	});
+
+	test('transpile multiline named imports with trailing comma', () => {
+		const result = transpileImports(`import {
+  B,
+  C,
+} from 'cat'
+`);
+		expect(result).toMatchInlineSnapshot(`
+"const cat$0 = require('cat');
+const B = cat$0.B;
+const C = cat$0.C;
+"
+`);
+	});
+
 	test('return code if there are no imports', () => {
 		const code = `<Button />`;
 		const result = transpileImports(code);
--- a/src/client/utils/rewriteImports.js
+++ b/src/client/utils/rewriteImports.js
@@ -45,7 +45,7 @@ export default function(str, fn = 'require') {
 	num = 0;
 	return str
 		.replace(NAMED, (_, asterisk, base, req, dep) =>
-			generate(req ? req.split(',') : [], dep, base, fn)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/client/utils/rewriteImports.js [hunk 1]
```diff
@@ -45,7 +49,7 @@ export default function(str, fn = 'require') {
 	num = 0;
 	return str
 		.replace(NAMED, (_, asterisk, base, req, dep) =>
-			generate(req ? req.split(',') : [], dep, base, fn)
+			generate(req ? req.split(',').filter(filterEmptyString) : [], dep, base, fn)
```
sapegin: `filterEmptyString` function doesn't filter anything, `.filter(dep => dep.trim())` would be enough.

src/client/utils/__tests__/transpileImports.spec.js
```diff
@@ -57,6 +57,34 @@ const E = snake$0.default || snake$0;
 `);
 	});
 
+	test('transpile multiline named imports with trailing comma', () => {
```
sapegin: This test isn't about trailing commas.

kopax:
I got myself stuck on this too, thank you for fixing this!

sapegin:
Thanks!

wkillerud:
Thank _you_ 🙏 😄

Spenc84:
@wkillerud I just upgraded to React Styleguidist 9.0.4 and tested out an import statement with a trailing comma and it produced the same error listed above. It's not a big deal since I never use trailing commas anyway, but I just thought it was a little odd since the issue was marked as resolved.

Spenc84:
@wkillerud Actually, please disregard. Yarn was telling me I was on version 9.0.4 but when I opened up the files in Node_Modules they were listed as 9.0.3. I remove the package and reinstalled it and it's working fine now. Thanks for fixing the issue.

PR #1304, opened 2019-03-04
Add more Cypress tests for isolated mode toggling

dougmacknz:
Some work towards https://github.com/styleguidist/react-styleguidist/issues/776

### DIFF
--- a/src/client/rsg-components/Examples/ExamplesRenderer.js
+++ b/src/client/rsg-components/Examples/ExamplesRenderer.js
@@ -7,7 +7,7 @@ const styles = () => ({
 	root: {},
 });
 
-export function ExamplesRenderer({ classes, children, name }) {
+export function ExamplesRenderer({ classes, name, children }) {
 	return (
 		<article className={classes.root} data-testid={name + '-examples'}>
 			{children}
@@ -17,8 +17,8 @@ export function ExamplesRenderer({ classes, children, name }) {
 
 ExamplesRenderer.propTypes = {
 	classes: PropTypes.object.isRequired,
-	children: PropTypes.node,
 	name: PropTypes.string.isRequired,
+	children: PropTypes.node,
 };
 
 export default Styled(styles)(ExamplesRenderer);
--- a/src/client/rsg-components/Examples/ExamplesRenderer.js
+++ b/src/client/rsg-components/Examples/ExamplesRenderer.js
@@ -7,13 +7,18 @@ const styles = () => ({
 	root: {},
 });
 
-export function ExamplesRenderer({ classes, children }) {
-	return <article className={classes.root}>{children}</article>;
+export function ExamplesRenderer({ classes, children, name }) {
+	return (
+		<article className={classes.root} id={name + '-examples'}>
+			{children}
+		</article>
+	);
 }
 
 ExamplesRenderer.propTypes = {
 	classes: PropTypes.object.isRequired,
 	children: PropTypes.node,
+	name: PropTypes.string.isRequired,
 };
 
 export default Styled(styles)(ExamplesRenderer);
--- a/src/client/rsg-components/Examples/ExamplesRenderer.js
+++ b/src/client/rsg-components/Examples/ExamplesRenderer.js
@@ -9,7 +9,7 @@ const styles = () => ({
 
 export function ExamplesRenderer({ classes, children, name }) {
 	return (
-		<article className={classes.root} id={name + '-examples'}>
+		<article className={classes.root} data-testid={name + '-examples'}>
 			{children}
 		</article>
 	);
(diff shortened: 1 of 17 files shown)


test/cypress/integration/component_spec.js
```diff
@@ -80,4 +80,33 @@ describe('Single component', () => {
 				.should('exist');
 		});
 	});
+
+	describe('isolated example mode', () => {
+		beforeEach(() => {
+			cy.get('[class^=rsg--root] article')
```
sapegin: Let's start adding test IDs to elements that can't be accessed by ARIA attributes or labels:

https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

dougmacknz: I've added an ID to this element, however I think it would be better to follow the recommended best practice in the docs and start using `data-*` attributes instead. It would have the benefit of not having to worry about creating unique IDs based on the component name as well. Thoughts?

sapegin: Sure, I meant _testing IDs_ ;-)

dougmacknz:
Alright, I've replaced out some of the other selectors with new test ids as well, hope you don't mind.

A couple of the decisions I made, with the reasoning, let me know if there's anything you disagree with:
- I've used `data-testid` rather than `data-cy`
    - It's library agnostic, in case we want to use it with another library in the future.
    - It's quick and easy to understand its purpose (some people might not know what `cy` is)
    - It's one of the recommended names by Cypress and also used in https://github.com/kentcdodds/dom-testing-library FWIW
* I've made `testid`'s unique (except for my last commit where I don't have some data needed to make it unique) eg. `data-testid="Button-examples"` & `data-testid="CounterButton-examples"` rather than a generic `data-testid="component-examples"`
    - I didn't see any documentation around best practice for this, but I'm guessing since it's called an ID it should probably be unique.
    - We may want to target a particular element in the future- in that case it would be best to have it unique from the start.
    - On the downside this does make selecting a bit more flaky (eg `[data-testid$="-container"]`) rather than being able to use the exact string.

Since you last saw this PR I've also combined tests/assertions that relied on state from the previous test into a single test. This was after reading further down on the Best Practices document - https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests

src/client/rsg-components/Playground/PlaygroundRenderer.js
```diff
@@ -37,7 +37,8 @@ export function PlaygroundRenderer({
 }) {
 	const { className, ...props } = previewProps;
 	return (
-		<div className={classes.root}>
+		// TODO: make the testid below unique eg. ${name}-example-${exampleIndex}
+		<div className={classes.root} data-testid={`${name}-example`}>
```
dougmacknz: Is there an easy way we can get the example index from here? I was able to do something hacky by using `toolbar.props.props.example`, but it's obviously not ideal and also blew up a test.

dougmacknz:
@sapegin would you like me to rebase this and resolve the conflicts?

sapegin:
This would be great!

dougmacknz:
Rebased!

src/client/rsg-components/Playground/PlaygroundRenderer.js
sapegin: Looks like there's `index` in the parent component:

https://github.com/styleguidist/react-styleguidist/blob/34f3c83e769542157c72d0e055ad8850d22b6001/src/client/rsg-components/Playground/Playground.js#L16

src/client/rsg-components/Examples/ExamplesRenderer.js
```diff
@@ -7,12 +7,17 @@ const styles = () => ({
 	root: {},
 });
 
-export function ExamplesRenderer({ classes, children }) {
-	return <article className={classes.root}>{children}</article>;
+export function ExamplesRenderer({ classes, name, children }) {
+	return (
+		<article className={classes.root} data-testid={name + '-examples'}>
```
sapegin: Let's use template literals for consistency here and in other palaces:

```suggestion
		<article className={classes.root} data-testid={`${name}-examples`}>
```

src/client/rsg-components/ToolbarButton/ToolbarButtonRenderer.js
```diff
@@ -74,6 +81,7 @@ ToolbarButtonRenderer.propTypes = {
 	onClick: PropTypes.func,
 	title: PropTypes.string,
 	small: PropTypes.bool,
+	dataTestID: PropTypes.string,
```
sapegin: Could be:

```suggestion
	testId: PropTypes.string,
```

dougmacknz:
Have fixed those up now :)

sapegin:
Thanks!

Issue #1246, opened 2019-01-11
hot reloading broken on windows


dbrelovsky:
**Current behavior**

Hot reloading is not working on a windows machine for all versions after 7.3.7.

I tracked it down to this change: https://github.com/styleguidist/react-styleguidist/pull/1133
On my machine all paths produced by ```getComponentFiles``` contain forward slashes as separator (in v7.3.8) and backslashes in v7.3.7.

If I change 
```js
const componentFiles = getFilesMatchingGlobs(componentGlobs, rootDir, ignore)
```
to 
```js
const componentFiles = getFilesMatchingGlobs(componentGlobs, rootDir, ignore).map(f => path.resolve(f));
```

hot reloading works again.

sapegin:
Feel free to send a pull request with a fix 👾

PR #1305, opened 2019-03-04
Fix: Resolve path of component files (#1246)

MichaelHancock:
Addresses hot reloading issue for windows

### DIFF
--- a/src/loaders/utils/getComponentFiles.js
+++ b/src/loaders/utils/getComponentFiles.js
@@ -1,4 +1,5 @@
 const glob = require('glob');
+const path = require('path');
 const isFunction = require('lodash/isFunction');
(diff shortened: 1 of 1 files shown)


src/loaders/utils/getComponentFiles.js
```diff
@@ -46,6 +47,7 @@ module.exports = function getComponentFiles(components, rootDir, ignore) {
 
 	// Resolve list of components from globs
 	const componentFiles = getFilesMatchingGlobs(componentGlobs, rootDir, ignore);
+	const resolvedComponentFiles = componentFiles.map(file => path.resolve(file));
```
sapegin: Please add a comment explaining why we need to do this line.

sapegin:
Thanks!

PR #1328, opened 2019-04-09
Allow output.globalObject from webpack-config

ernil-labgruppen:
In some cases i.e. https://github.com/webpack/webpack/issues/6642 a user needs to set output.globalConfig in their webpack config. Unfortunately this is thrown away by styleguidist.

This patch allows specifically output.globalObject to be retained from the users webpack config.


sapegin:
You can already do this with https://react-styleguidist.js.org/docs/configuration.html#dangerouslyupdatewebpackconfig

ernil-labgruppen:
Ah, thanks for pointing that out. works perfectly and is a much more elegant solution (even with the 'dangerously' in the title ;) ).

I'll drop this pull request.

PR #1331, opened 2019-04-10
feature(scroll): Scroll section in to view when exiting isolated view

Blightwidow:
Just a small PR to center the view on the section was viewing when leaving the isolated mode.


sapegin:
Thanks!

PR #1335, opened 2019-04-15
Feat: Allow code base color to be customized with theme config

abhishekdev:
#### Change Summary

- Add `theme.color.codeBase` property to config
- Allow syntax highlight default color to set without affecting base color
- The `codeBase` and `codeBackground` can be set to allow finer control of code text contrast ratios. e.g. Create dark themes for code samples.

#### Why is this required?

The `them.color.code*` properties can be used to customized the syntax highlight theme. However, setting a dark color as the codeBackground makes the code almost unreadable (due to low contrast ratios). 

##### Issue Screenshots
Static Code Block
Issue 1 a - Static Code Block
 Editable Code Block
Issue 1 b - Editable Code Block

##### Screenshots with proposed solution
Static Code Block
Sol 1 a - Static Code Block
Editable Code Block
Sol 1 b - Editable Code Block

#### Is there a workaround?
Yes. 

- Using custom style wrappers for `rsg-components/Editor` and `rsg-components/Markdown/Pre` components 

There could be other approaches, but it seems most approaches require much more code change than should be required to customize colors with themes.

### DIFF
--- a/src/client/styles/prismTheme.js
+++ b/src/client/styles/prismTheme.js
@@ -1,4 +1,7 @@
 const prismTheme = ({ color }) => ({
+	'&': {
+		color: color.codeBase,
+	},
 	[`& .token.comment,
 & .token.prolog,
 & .token.doctype,
--- a/src/client/styles/theme.js
+++ b/src/client/styles/theme.js
@@ -26,6 +26,7 @@ export const color = {
 	ribbonBackground: '#e90',
 	ribbonText: '#fff',
 	// Based on default Prism theme
+	codeBase: '#333',
 	codeComment: '#6d6d6d',
 	codePunctuation: '#999',
 	codeProperty: '#905',


abhishekdev:
I could not find any tests around styles, so not sure if this change warrants any other change to test or docs. 

A potential follow up update could be to update the [examples/customised/styleguide.config.js](https://github.com/styleguidist/react-styleguidist/blob/master/examples/customised/styleguide.config.js) to use a dark-theme.

```diff
--- a/examples/customised/styleguide.config.js
+++ b/examples/customised/styleguide.config.js
@@ -10,6 +10,10 @@ module.exports = {
 		linkHover: '#90a7bf',
 		border: '#e0d2de',
 		font: ['Helvetica', 'sans-serif'],
+		color: {
+			codeBase: '#fff',
+			codeBackground: '#000',
+		},
 	},
 	styles: {
 		Playground: {
```

sapegin:
Thanks! This is very good ;-)

Yeah, both follow ups make sense to me. The only possible issue with the last one: test static code snipped without language tag, I guess there will be no Prism theme applied, so maybe instead of removing a color we need to add `codeBase` and `codeBackground`.

abhishekdev:
Thanks for quick review and release 😄 

I'll do some tests around the code cleanup and get back with a PR.

PR #1341, opened 2019-04-18
Add styled-components with typescript example

jorgeortega:
Adding a simple example with styled-components and TypeScript

### DIFF
--- a/examples/styled-components/src/components/Button/Button.tsx
+++ b/examples/styled-components/src/components/Button/Button.tsx
@@ -1,0 +1,56 @@
+import React, { ReactNode } from 'react'
+import styled from 'styled-components';
+import { themeGet } from 'styled-system';
+
+interface ButtonProps {
+	/** Button label */
+	children: ReactNode,
+	/** Button variation */
+	variant: 'primary' | 'secondary',
+	fullWidth: Boolean,
+};
+
+const getColor = (variant: string) => ({ primary: 'bg', secondary: 'primary' }[variant]);
+const getBgColor = (variant: string) => ({ primary: 'primary' }[variant]);
+
+/**
+ * A button.
+ */
+const Button = styled.button<ButtonProps>`
+	display: ${props => props.fullWidth && 'block'};
+	width: ${props => props.fullWidth && '100%'};
+	height: 2.5rem;
+	padding: ${themeGet('space.3')} ${themeGet('space.4')};
+	text-align: center;
+	border: 1px solid ${themeGet('colors.primary')};
+	border-radius: ${themeGet('radii.base')};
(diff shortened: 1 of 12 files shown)


examples/styled-components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import styled from 'styled-components'
+
+interface ButtonProps extends HTMLButtonElement {
+	btnType: 'primary' | 'secondary',
```
sapegin: Let's simplify this a bit: rename a prop to `variant` and pass it to a styled component, so we don't duplicate the whole component markup twice.

Or just copy code here: 
https://github.com/component-driven/component-driven-development/blob/master/src/components/core/Button/Button.js

examples/styled-components/styleguide.config.js
```diff
@@ -0,0 +1,33 @@
...
+				}
+			],
+			noParse: /\.(css|scss)/
+		},
+		resolve: {
+			modules: [
+			  path.resolve(__dirname, 'src'),
+			  path.join(__dirname, 'node_modules')
+			],
+			extensions: ['.js', '.ts', '.tsx']
```
sapegin: Make sense to update the default extensions list:
https://github.com/styleguidist/react-styleguidist/blob/0cd533c18a3ac79bfec2c0e34f65780a748bea0a/src/scripts/make-webpack-config.js#L44

examples/styled-components/styleguide.config.js
```diff
@@ -0,0 +1,33 @@
...
+				{
+					test: /\.(js|ts)x?$/,
+					exclude: /node_modules/,
+					loader: 'babel-loader',
+				}
+			],
+			noParse: /\.(css|scss)/
+		},
+		resolve: {
+			modules: [
```
sapegin: Is this necessary?

examples/styled-components/Readme.md
```diff
@@ -0,0 +1,14 @@
+# React Styleguidist styled-components typescript example style guide
```
sapegin: It's TypeScript.

examples/styled-components/Readme.md
```diff
@@ -0,0 +1,14 @@
...
+
+See [deployed version](https://react-styleguidist.js.org/examples/styled-components/).
```
sapegin: This won't work, we only deploy basic example. But maybe we should deploy other examples too…

jorgeortega:
@sapegin Thanks for your feedback. All changes updated.

sapegin:
Thanks!

PR #1348, opened 2019-04-30
Distinguish active projects from inactive

mccalluc:
This list was really helpful! An active/inactive distinction seems pretty clear... but beyond that which tool is best for the job gets much more subjective.

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -613,14 +613,19 @@ Another important distinction is that Storybook shows only one variation of one 
 
 ## Are there any other projects like this?
 
-- [Atellier](https://github.com/scup/atellier), a React components emulator.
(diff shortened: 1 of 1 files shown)


sapegin:
Thanks for the PR!

I'm wondering do we really need inactive tools anymore? How did you determine which ones are inactive?

mccalluc:
For the inactive ones, I've added a note at the end of the line saying when the last release was... or actually, the last commit to github: that was easier to get.

Your call: For me, having this exhaustive list was really useful, and made me more confident about my choice in the end.

sapegin:
Yeah, that makes sense!

sapegin:
I'll merge as soon as I figure out why CI isn't working. It's not related to your changes 🤷‍♂

PR #1349, opened 2019-04-30
Chore: Replace 'walkes' with 'estree-walker'

### LINKED ISSUE #1210: Remove dependency on "walkes"
Starting in version 8 of the package it looks like "walkes" was added as a project dependency, although I cannot find out where it is used. I am not sure if this is needed?

That library is LGPL-3 licensed and can make this project unusable in certain licensing situations starting in version 8.

eragon512:
@sapegin could you please look at this PR and check if this is acceptable?

I'm concerned that this might break existing functionality, as the AST walking is a core part of the code, and the implemented code technically breaks the tests, however, this seems to be a popular fix, so opening this up for further discussion

Fixes #1210

### DIFF
--- a/src/client/utils/transpileImports.js
+++ b/src/client/utils/transpileImports.js
@@ -1,5 +1,5 @@
 // @flow
-import walkes from 'walkes';
+import { walk } from 'estree-walker';
 import rewriteImports from './rewriteImports';
 import getAst from './getAst';
 
@@ -21,11 +21,11 @@ export default function transpileImports(code: string): string {
 	}
 
 	let offset = 0;
-	walkes(ast, {
+	walk(ast, {
 		// import foo from 'foo'
 		// import 'foo'
-		ImportDeclaration(node) {
-			if (node.source) {
+		enter: node => {
+			if (node.type === 'ImportDeclaration' && node.source) {
 				const start = node.start + offset;
 				const end = node.end + offset;
 
--- a/flow-typed/estree-walker.js.flow
+++ b/flow-typed/estree-walker.js.flow
@@ -1,10 +1,9 @@
-declare module 'walkes' {
+declare module 'estree-walker' {
 	import type { AcornNode } from 'acorn';
 
-	declare function walkes(
+	declare export function walk (
 		node: AcornNode,
 		visitors: { [key: string]: (node: AcornNode) => void }
 	): void;
-
-	declare export default typeof walkes;
+	
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -4840,6 +4840,11 @@
       "resolved": "https://registry.npmjs.org/estraverse/-/estraverse-4.2.0.tgz",
       "integrity": "sha1-De4/7TH81GlhjOc0IJn8GvoL2xM="
     },
+    "estree-walker": {
+      "version": "0.6.0",
+      "resolved": "https://registry.npmjs.org/estree-walker/-/estree-walker-0.6.0.tgz",
+      "integrity": "sha512-peq1RfVAVzr3PU/jL31RaOjUKLoZJpObQWJJ+LgfcxDUifyLZ1RjPQZTl0pzj2uJ45b7A7XpyppXvxdEqzo4rw=="
+    },
     "esutils": {
       "version": "2.0.2",
       "resolved": "https://registry.npmjs.org/esutils/-/esutils-2.0.2.tgz",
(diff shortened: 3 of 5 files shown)


semi-sentient:
It looks like "walkes" is still listed as a dev dependency. Shouldn't that be removed as well?

eragon512:
@semi-sentient yeah, you're right, I'll do that.
Is this PR under consideration? because I had closed it off, due to the fact that estree-walker was a pretty big module.
Is the license issue important enough to increase the bundle size? :o

semi-sentient:
I was just sort of snooping around and noticed that walkes was still listed as a dependency. Since it has a more restrictive license we can't import it for use here @Amazon so just wanted to point that out as I'd very much love to use react-styleguidist on my current project. ;)

richardwestenra:
@eragon512 What, in your view, is necessary to get this PR ready for consideration? This dependency license has been an issue for my team too.

ottis:
Keen to get this resolved too, similar situation as @semi-sentient i.e. we'd love to use this but the license of walkes is restrictive 😢

eragon512:
@richardwestenra @ottis I didn't realise this was such a major issue for you guys 😅 . I'll try to take this up again, seeing the demand for it

My main concern is that the replacement in this PR (`estree-walker`) *might* break existing functionality, but I guess @sapegin would be a better person to weigh in here, as I'm not very familiar with ASTs and such

eragon512:
Re-opening for active consideration and further work

src/client/utils/__tests__/compileCode.spec.js
```diff
@@ -14,10 +14,7 @@ var bar = baz.bar;"
 
 	test('transform imports to require()', () => {
 		const result = compileCode(`import foo from 'bar'`, compilerConfig);
-		expect(result).toMatchInlineSnapshot(`
-"const bar$0 = require('bar');
-const foo = bar$0.default || bar$0;"
-`);
+		expect(result).toMatchInlineSnapshot(`"import foo from 'bar'"`);
```
sapegin: These changes are very suspicious, it means the compilation doesn't happen: the resulting code is the same as the source ;-£

src/client/utils/__tests__/transpileImports.spec.js
```diff
@@ -42,17 +32,13 @@ import E from 'snake'
 "/**
 * Some important comment
 */
-require('dog');
+import 'dog'
```
sapegin: Also regular `require()` should still work.

src/client/utils/transpileImports.js
```diff
@@ -21,7 +21,7 @@ export default function transpileImports(code: string): string {
 	}
 
 	let offset = 0;
-	walkes(ast, {
+	walk(ast, {
```
sapegin: Looks like the API is different and it should be something like this:

```js
walk( ast, {
			enter ( node ) {
				if (node.type === 'ImportDeclaration') {
                                     // do the work
			        }
			},
})


```

src/client/utils/__tests__/compileCode.spec.js
eragon512: My fears exactly

eragon512:
@sapegin please check the code when you get time, `estree-walker` seems to be a perfect drop-in replacement

The earlier suspicious test snapshots were my bad, I hadn't completely replaced the `walkes` dependency before 😅

src/client/utils/transpileImports.js
eragon512: Implemented 👍

sapegin:
That look good! 

Could you please replace Flow types here?

https://github.com/styleguidist/react-styleguidist/tree/master/flow-typed

eragon512:
> That look good!
> 
> Could you please replace Flow types here?
> 
> https://github.com/styleguidist/react-styleguidist/tree/master/flow-typed

Replaced the flow types, as requested

sapegin:
Awesome!

If you merge these two conditions, it should fix the code coverage check:

```
 if (node.type !== 'ImportDeclaration') {
   return;
 }
 if (!node.source) {
  return;
 }
```

eragon512:
done, please confirm

sapegin:
Thanks, merging!

eragon512:
@semi-sentient @richardwestenra @ottis here you go!

Swatinem:
Hm, just out of boredom I was checking all my old npm packages and was like *whaaaaaaa* when I saw this package getting ~50k downloads a week while its been basically unmaintained since 4-6 years.
… and then I landed here and this made my day :-D

Issue #1343, opened 2019-04-24
Support Hooks in examples


sapegin:
**The problem**

Hooks could be a better way to use state in examples, than our custom `initialState`. But they don't work right now because we're wrapping example code in a class component, and hooks only work in a function component.

```jsx
const [primary, setPrimary] = React.useState(false);
<Button isPrimary={primary} onClick={() => setPrimary(true)}>Push me</Button>
```

**Proposed solution**

Wrap examples in a function component and pass state from the current wrapper class component as a prop.

**Alternative solutions**

1. Drop React < 16.8 support and use hooks to manage state: rewrite the current state manager as a function component with state.

2. Consider removing the `initialState` and rely on hooks only. (Breaking change.)

    2.1. Deprecate `initialState` / `setState`, update all docs and examples to hooks. Remove in the next major release.

**Additional context**

```
Invariant Violation: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
```

eragon512:
hey, I would like to work on this issue

sapegin:
@eragon512 That's awesome! 👍

eragon512:
Opened a PR which implements the proposed solution, as it seemed to be the simplest way to support both legacy and new applications. 

If dropping React < 16.8 support is an option, then I would be happy to refactor the wrapper class to implement hooks

eragon512:
@sapegin Closing this issue, as it has been addressed in #1353

hugotox:
Hello! What's the way to document components that use hooks? I'm getting the `Invalid hook call` error mentioned in the OP

foscjos2:
@hugotox did you ever get a working solution to this?

AshleyFang:
I have also come across this problem. Does anyone have any recommended solutions? I am willing to give them a try.

```bash
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
[...]

PR #1353, opened 2019-05-04
Support React Hooks in examples

eragon512:
Implements the proposed solution for #1343

### DIFF
--- a/src/client/rsg-components/ReactExample/ReactExample.js
+++ b/src/client/rsg-components/ReactExample/ReactExample.js
@@ -6,6 +6,15 @@ import splitExampleCode from '../../utils/splitExampleCode';
 
 /* eslint-disable react/no-multi-comp */
 
+// Wrap the example component with a Functional Component to support
+// hooks in examples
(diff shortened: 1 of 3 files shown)


src/client/rsg-components/ReactExample/ReactExample.js
```diff
@@ -6,6 +6,13 @@ import splitExampleCode from '../../utils/splitExampleCode';
 
 /* eslint-disable react/no-multi-comp */
 
+function FunctionComponentWrapper(props) {
```
sapegin: Could you add a comment here, explaining that we need it to support hooks in examples?

eragon512: added the comment, as requested

eragon512:
~~~~
Could you please add some tests and replace initialState / setState in some examples to show how it works?
~~~~

@sapegin I added an extra example, because we still support the old state method, right?

sapegin:
Awesome, thanks!

Issue #1320, opened 2019-03-20
assetsDir as an array breaks styleguidist build


ammmze:
**Current behavior**

When `assetsDir` is configured as an array, the `styleguide:build` script fails like this...

```bash
$ npm run styleguide:build

> react-styleguidist-example@1.0.0 styleguide:build /Users/me/projects/styleguidist-example
> styleguidist build

Building style guide...
 FAIL  Failed to compile

The "path" argument must be of type string. Received type object
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! react-styleguidist-example@1.0.0 styleguide:build: `styleguidist build`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the react-styleguidist-example@1.0.0 styleguide:build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

> Note: Running `npm start` or `npm styleguide` works fine. The `styleguidist server` successfully starts up without issue. And using a string value is also successful.

**To reproduce**

I've forked the example repo and made [this](https://github.com/ammmze/example/commit/310e059cc15bd9a4c8f1cf458cc0bb0fe5391a50) change and then run `npm run styleguide:build` which triggers the error.

**Expected behavior**

The build should be successful as it is when running the server.

sapegin:
Feel free to send a pull request with a fix.

PR #1367, opened 2019-05-17
Fix build if assetsDir is an array

mendrew:
copy-webpack-plugin is not allowing `from` to be as an array.
If we still want to use `from` and support array format for
assetsDir, then we should pass an array of objects {from: ...} to
copy-webpack-plugin

iss: #1320

### DIFF
--- a/src/scripts/make-webpack-config.js
+++ b/src/scripts/make-webpack-config.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const isArray = require('lodash/isArray');
 const webpack = require('webpack');
 const TerserPlugin = require('terser-webpack-plugin');
 const MiniHtmlWebpackPlugin = require('mini-html-webpack-plugin');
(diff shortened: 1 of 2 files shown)


src/scripts/make-webpack-config.js
```diff
@@ -93,11 +94,9 @@ module.exports = function(config, env) {
 				}),
 				new CopyWebpackPlugin(
 					config.assetsDir
-						? [
-								{
-									from: config.assetsDir,
-								},
-						  ]
+						? isArray(config.assetsDir)
+							? config.assetsDir.map(dir => ({ from: dir }))
```
mendrew: I don't like ternary nesting, but here it feels natural.
Other approaches would move logic too far from usage.

But I don't mind to change it 🙂

src/scripts/make-webpack-config.js
```diff
@@ -93,11 +94,9 @@ module.exports = function(config, env) {
 				}),
 				new CopyWebpackPlugin(
 					config.assetsDir
-						? [
-								{
-									from: config.assetsDir,
-								},
-						  ]
+						? isArray(config.assetsDir)
```
sapegin: No need for Lodash here, use `Array​.isArray()`.

But I'd probably rewrite it like this:

```js
_.castArray(config.assetsDir).map(dir => ({ from: dir }))
```

To get rid of the condition.

mendrew: Oh, that's much better!

mendrew: Thank you for suggestion. I've updated PR.

Issue #1361, opened 2019-05-15
Can react-styleguidist support markdown metadata？


tuntuntutu:
I want to get the metadata of markdown docs，For example，

```
---
title: componentName
anthor:  John
---
```
and  I can get the metadata in `SectionRenderer`, just like 
```
{
title: 'componentName',
anthor: 'John'
}
```

sapegin:
This was proposed couple of times before and seems like a useful feature — feel free to send a pull request.

eragon512:
Hey, I'd like to take up this issue, how should the metadata be made available to the user?

sapegin:
There's an undocumented feature to load JSON files which we need to replace with a frontmatter:

https://github.com/styleguidist/react-styleguidist/blob/0ef7ff38a550b7b049bd0b82d87bd2ec97af3fce/src/loaders/utils/processComponent.js#L9-L18

PR #1369, opened 2019-05-19
Support Markdown Metadata for components

eragon512:
Refer: #1361

### DIFF
--- a/src/loaders/utils/processComponent.js
+++ b/src/loaders/utils/processComponent.js
@@ -1,5 +1,6 @@
 const fs = require('fs');
 const path = require('path');
+const frontMatter = require('front-matter');
 const getNameFromFilePath = require('./getNameFromFilePath');
 const requireIt = require('./requireIt');
 const slugger = require('./slugger');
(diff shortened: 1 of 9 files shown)


src/scripts/logger.js
```diff
@@ -1,6 +1,10 @@
 /* eslint-disable no-console */
 
-const _ = require('lodash/fp');
+const _flow = require('lodash/fp/flow');
```
sapegin: Any reason to change Lodash imports everywhere? Seems unrelated to the task but makes reviewing harder.

src/loaders/utils/processComponent.js
```diff
@@ -28,14 +48,13 @@ module.exports = function processComponent(filepath, config) {
...
-	const componentMetadataPath = getComponentMetadataPath(filepath);
 	return {
 		filepath: componentPath,
 		slug: slugger.slug(componentName),
 		pathLine: config.getComponentPathLine(componentPath),
 		module: requireIt(filepath),
 		props: requireIt(`!!${propsLoader}!${filepath}`),
 		hasExamples: examplesFile && fs.existsSync(examplesFile),
-		metadata: fs.existsSync(componentMetadataPath) ? requireIt(componentMetadataPath) : {},
+		metadata: getComponentMetadata(filepath),
```
sapegin: It actually would make more sense to make it part of `props`. The current approach has multiple issues:

1. Metadata won't be hot reloaded on changes in Markdown file
2. We're reading a Markdown file one more time.
3. We're duplicating Markdown file name finding code.
4. We're doing some weird magic on client side, but we shouldn't send frontmatters to the browser at al.

src/scripts/logger.js
eragon512: will remove the commit for this :+1:

src/loaders/utils/processComponent.js
eragon512: @sapegin so, should we keep everything (including the frontmatter) in `props` and remove the `metadata` prop above?

sapegin: Yup! `props` is quite stupid name, in some places we have `props.props.props` 🤦‍♂️ but I wound't date to change it before we migrate all the code to TypeScript ;-)

> Also, does the examplesFile variable have the markdown filepath? Didn't notice that, that's why I duplicated the Markdown file naming code 😅

Yeah, and we have it in props loader too:

https://github.com/styleguidist/react-styleguidist/blob/0ef7ff38a550b7b049bd0b82d87bd2ec97af3fce/src/loaders/props-loader.js#L69-L71

Issue #1371, opened 2019-05-20
Compilation error on using 'async' and 'await' in readme.md


gargroh:
**Current behavior**

When using async functions in readme.md then getting compilation error, before upgrading to 9.x.x it was working fine. 
Error :-
![image](https://user-images.githubusercontent.com/42495927/58002894-32da6980-7afd-11e9-82e2-d67ffd1c69ec.png)

Simplified code :-
![image](https://user-images.githubusercontent.com/42495927/58003423-5eaa1f00-7afe-11e9-9aa4-bece1f55f449.png)

**To reproduce**
Just try to use async functions in readme.md, it can be easily replicated.

**Expected behavior**

It should compile perfectly as it doing for 8.x.x

sapegin:
Feel free to send a pull request with a fix.

mendrew:
### The error explanation
We have the error here because styleguidinst has updated version of [bublé](https://github.com/bublejs/buble) ([from v0.9.4 to v0.9.7](https://github.com/mendrew/react-styleguidist/commit/48f72208d6cfb19e34654f22b47c86d8531e2b71#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R34)) which does actual code transformation.
In latest version of bublé they throw this error because right now [bublé cannot transpile async](https://github.com/bublejs/buble/issues/109#issuecomment-380854164), so they've added a special transform option called `asyncAwait` to pass `async` through as before. Here the code, how they do it: [example](https://github.com/bublejs/buble/blob/086d895cdd616de065bf82339567ed4f0bc44b75/src/program/types/ArrowFunctionExpression.js#L6-L12)

It has been working before because bublé just passed `async` through.
Right now it throws the error, but you are able to confgure bublé to pass async through without transformation by using [compilerConfig](https://react-styleguidist.js.org/docs/configuration.html#compilerconfig) in your `styleguide.config.js`.
Like this: 
```js
  compilerConfig: {
    transforms: { asyncAwait: false }
  }
```

It also makes sence to pass default styleguidinst options for `compilerConfig` ([code](https://github.com/mendrew/react-styleguidist/blob/010761b5752ac43841aa54837c5562eddd8f9d82/src/scripts/schemas/config.js#L34-L45)), as documentation about default options is outdated.

### Questions
@sapegin Should we add ` transforms: { asyncAwait: false }` for default compilerConfig to make async/await work by default?
[...]

PR #1379, opened 2019-06-05
Disable asyncAwait transform to make async/await work by default

mendrew:
iss: #1371

Add `transforms: { asyncAwait: false }` for default compilerConfig to make async/await work by default
Update default value of compilerConfig in the configuration documentation.

### Error explanation: ([issue discussion](https://github.com/styleguidist/react-styleguidist/issues/1371#issuecomment-496003017))
We have the error here because styleguidinst has updated version of [bublé](https://github.com/bublejs/buble) ([from v0.9.4 to v0.9.7](https://github.com/mendrew/react-styleguidist/commit/48f72208d6cfb19e34654f22b47c86d8531e2b71#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R34)) which does actual code transformation.
In latest version of bublé they throw this error because right now [bublé cannot transpile async](https://github.com/bublejs/buble/issues/109#issuecomment-380854164), so they've added a special transform option called `asyncAwait` to pass `async` through as before. Here the code, how they do it: [example](https://github.com/bublejs/buble/blob/086d895cdd616de065bf82339567ed4f0bc44b75/src/program/types/ArrowFunctionExpression.js#L6-L12)

It has been working before because bublé just passed `async` through.
Right now it throws the error, but you are able to confgure bublé to pass async through without transformation by using [compilerConfig](https://react-styleguidist.js.org/docs/configuration.html#compilerconfig) in your `styleguide.config.js`.
Like this: 
```js
  compilerConfig: {
    transforms: { asyncAwait: false }
  }
```

It also makes sence to pass default styleguidinst options for `compilerConfig` ([code](https://github.com/mendrew/react-styleguidist/blob/010761b5752ac43841aa54837c5562eddd8f9d82/src/scripts/schemas/config.js#L34-L45)), as documentation about default options is outdated.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -59,7 +59,24 @@ Your application static assets folder, will be accessible as `/` in the style gu
 
 #### `compilerConfig`
 
-Type: `Object`, default: `{ objectAssign: 'Object.assign' }`
+Type: `Object`, default:
+
+```javascript
+{
+  // Don't include an Object.assign ponyfill, we have our own
+  objectAssign: 'Object.assign',
+  // Transpile only features needed for IE11
+  target: { ie: 11 },
+  transforms: {
+    // Don't throw on ESM imports, we transpile them ourselves
+    modules: false,
+    // Enable tagged template literals for styled-components
+    dangerousTaggedTemplateString: true,
+    // configure bublé to pass async through without transformation and error
+    asyncAwait: false,
+  },
+}
+```
 
 Styleguidist uses [Bublé](https://buble.surge.sh/guide/) to run ES6 code on the frontend. This config object will be added as the second argument for `buble.transform`.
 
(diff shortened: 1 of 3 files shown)


mendrew:
This is how updated `compilerConfig docs` look like 
updated-compiler-config-doc

Maybe it makes sense to specify that if you use async/await then it won't be transform.

And I don't know where to add a test for it new compierConfig default value..

docs/Configuration.md
```diff
@@ -59,7 +59,24 @@ Your application static assets folder, will be accessible as `/` in the style gu
...
+  // Don't include an Object.assign ponyfill, we have our own
+  objectAssign: 'Object.assign',
+  // Transpile only features needed for IE11
+  target: { ie: 11 },
+  transforms: {
+    // Don't throw on ESM imports, we transpile them ourselves
+    modules: false,
+    // Enable tagged template literals for styled-components
+    dangerousTaggedTemplateString: true,
+    // configure bublé to pass async through without transformation and error
```
sapegin: “configure bublé to” → “”

(The whole thing is to configure Buble ;-)

mendrew:
Thank you for suggestions. 

I've updated the PR. Added test and changed comments.

sapegin:
Thanks!

Issue #1281, opened 2019-02-20
Named exports are not supported


smashercosmo:
I don't know if it's a bug or by design, but...

**Current behavior**

Named component exports are not supported even with ```@component``` annotation

**To reproduce**

try to load component with named export in styleguidist

**Expected behavior**

It used to work before. And should still work, considering that named exports are gaining more popularity over default exports (https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/)

Upd: another example of named exports becoming new default (😄): https://github.com/este/este/issues/1684

SammyM:
I've also encountered this issue. A workaround is to import the component into each example in the markdown file like so:
```
import { Component } from 'components';

<Component />
```
But this isn't ideal, and also wouldn't work for people who use the `defaultExample: true` setting.

adriankremer:
since i've updated to 9.x importing on the same module name as the Markdown file produces: SyntaxError: Identifier 'Button' has already been declared.

korziee:
> since i've updated to 9.x importing on the same module name as the Markdown file produces: SyntaxError: Identifier 'Button' has already been declared.

I'm getting the same error, without looking at the source code I'd assume that the identifier was previously set using `var`/`let` and since has been changed to `const`.

We've been using this method of importing the component into every readme through a named export. Has anyone had any success in mitigating this?

It would be great if we could turn off auto import by default.

quentez:
We're also stuck on v8 because of the above.

MattAndDev:
Hi there, stuck in a smiliar not-upgradable condition.

What we have named exports for functional components:
`components/component-name.jsx`
```
const ComponentName = () => (<p></p>)
ComponentName.propTypes = {}
```

This works beautifully in `8.x.x` but breaks in resolving correctly the components in `9.x`
I played around a bit with propsParser and resolver options, but it does not seem a react-docgen issue (it resolves the correct docs for the given named export).

With a little bit of direction (where to look) I'd be happy to contriibute back.

Amazing work 👍

mendrew:
### Possible problematic place

I've been experimenting with this and found that these lines are responsible for example code imports
https://github.com/styleguidist/react-styleguidist/blob/f43cb8ae6eab0b2b5a016bcdab2430a1f2c3b3af/src/loaders/examples-loader.js#L74-L90

And comments are valid. 
First we do
> // const name$0 = require(path); 

and then 
> // const name = name$0.default || name$0;

[...]

PR #1380, opened 2019-06-05
Fix named export for example components

mendrew:
iss: #1281

Updated examples-loader to import component with named export
Named export of Basic button is now working

I tried suggestion `const name = name$0[name] || name$0.default || name$0;`
CounterButton component is now falling with an error:
```
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.
Check the render method of `FunctionComponentWrapper`.

```
CoutnerButtons is falling

Placeholder from my previous [report](https://github.com/styleguidist/react-styleguidist/issues/1281#issuecomment-498433301) is now working.

### DIFF
--- a/src/loaders/examples-loader.js
+++ b/src/loaders/examples-loader.js
@@ -82,8 +82,11 @@ export default function examplesLoader(source) {
 				b.variableDeclaration('const', [
 					b.variableDeclarator(
 						b.identifier(name),
-						b.logicalExpression('||', b.identifier(`${name}$0.default`), b.identifier(`${name}$0`))
-					),
+						b.logicalExpression('||',
+							b.identifier(`${name}$0['${name}']`),
+							b.logicalExpression('||', b.identifier(`${name}$0.default`), b.identifier(`${name}$0`)),
+						),
+					)
 				]),
 			])
 		)
--- a/src/loaders/examples-loader.js
+++ b/src/loaders/examples-loader.js
@@ -78,15 +78,20 @@ export default function examplesLoader(source) {
 				b.variableDeclaration('const', [
 					b.variableDeclarator(b.identifier(`${name}$0`), requireIt(requireRequest).toAST()),
 				]),
-				// const name = name$0.default || name$0;
+				// const name = name$0[name] || name$0.default || name$0;
 				b.variableDeclaration('const', [
 					b.variableDeclarator(
 						b.identifier(name),
-						b.logicalExpression('||',
+						b.logicalExpression(
+							'||',
 							b.identifier(`${name}$0['${name}']`),
-							b.logicalExpression('||', b.identifier(`${name}$0.default`), b.identifier(`${name}$0`)),
-						),
-					)
+							b.logicalExpression(
+								'||',
+								b.identifier(`${name}$0.default`),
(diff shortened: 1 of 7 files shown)


mendrew:
Ok, I finally get the idea why with this fix we have problems with other components from example.
The reason is simple but I've been debugging a lot till I get it.

### First example: CounterButton. 
Markdown code for CounterButton looks like this 
```js
import Button from '../Button'
let ref
;<div>
  <CounterButton ref={r => (ref = r)} />
  <Button size="small" onClick={() => ref.set(0)}>
    Reset
  </Button>
</div>
```

The error is about this import `import Button from '../Button'`. Button is no longer has default export in my PR and Button should be imported using named import.

### Second example: was tricked by it 
```jsx
import React from 'react'
import Button from 'rsg-example/components/Button'
import Placeholder from 'rsg-example/components/Placeholder'
;<Button>
  <Placeholder />
</Button>
```

In this case I had error with `import Button from 'rsg-example/components/Button'`, I thought it is absolutely different button, but then I looked at the styleguidist config and found this
https://github.com/mendrew/react-styleguidist/blob/1ed431d693af196c8c4d8f84839c60ac06274f2b/examples/basic/styleguide.config.js#L8

It's **the same** button component, and we should import it in the same way with named import.

mendrew:
I thinks that ast bulder code update `const name = name$0[name] || name$0.default || name$0;` is enough.

I'm going to remove WIP code and update tests.

mendrew:
I've finally updated the PR. It's ready for review.

I'm wondering does it make sense to add named export to some basic example code, just to have this case in examples?

sapegin:
Haha, this all sounds very familiar! ;-) And thanks for fixing so many things in Styleguidist!

> I'm wondering does it make sense to add named export to some basic example code, just to have this case in examples?

Yeah, that would make sense — tests aren't always enough to make sure things are still working.

mendrew:
> Haha, this all sounds very familiar! ;-) And thanks for fixing so many things in Styleguidist!

You are welcome! I found the code pretty interesting and clear. So it's a kind of pleasure to debug 🙂.   I haven't used it yet. But definitely will. And all this design system area looks inspiring.

> Yeah, that would make sense — tests aren't always enough to make sure things are still working.

I've updated the PR. Added such kind of import into RandomButton, so it likely hurt anyone who uses somehow example/basic components.
Did update for RadmonButton markdown to be explicit in a way we import this button.

Any other suggestions? @sapegin

src/loaders/examples-loader.js
```diff
@@ -82,8 +82,11 @@ export default function examplesLoader(source) {
 				b.variableDeclaration('const', [
 					b.variableDeclarator(
 						b.identifier(name),
-						b.logicalExpression('||', b.identifier(`${name}$0.default`), b.identifier(`${name}$0`))
-					),
+						b.logicalExpression('||',
```
sapegin: Let's update a comment above to:

```
// const name = name$0[name] || name$0.default || name$0;
```

mendrew: Oh, you are right! I was thinking about it but forgot! Updated.

sapegin:
Thanks! 🦄

Issue #1332, opened 2019-04-11
pagePerSection links break when section name includes a "/"


stonebk:
**Current behavior**

If you have a forward slash in your section name, the `pagePerSection` links are all broken, resulting in "Page not found". Removing the forward slashes in the names fixes the issue.

**To reproduce**

Simple `styleguide.config.js`:

```
module.exports = {
    pagePerSection: true,
    sections: [{
        name: '@foo/components'
    }, {
        name: '@bar/components'
    }]
};
```

**Expected behavior**

Links should work for any text in a section name field. Probably just need to urlencode the string.

sapegin:
Feel free to send a pull request with a fix.

stonebk:
This appears to have fixed the top level section link, but any sub-links within that section are still unencoded and result in "Page not found".

mendrew:
@stonebk could you please give an example of the config file to get an exact idea of sub-links you are talking about?
I was able to reproduce your problem by applying your config from the description on Styleguidist sections example. I'd be glad to fix this as well.

stonebk:
@mendrew it looks something like this:

```
{
    "webpackConfig": { ... },
    "styleguideComponents": { ... },
    "title": "Foundation Team APIs",
    "usageMode": "expand",
    "require": [ ... ],
    "pagePerSection": true,
    "sections": [
        {
            "name": "Introduction",
            "content": "docs/introduction.md"
        },
        {
            "name": "@zillow/zrs-activation",
            "sections": [
                {
                    "name": "Components",
                    "components": "/Users/stonebk/foundation-site/node_modules/@zillow/zrs-activation/src/components/**/[A-Z]*.{js,jsx,ts,tsx}"
                }
            ]
        },
        {
            "name": "@zillow/zrs-profile",
            "sections": [
                {
                    "name": "Components",
                    "components": "/Users/stonebk/foundation-site/node_modules/@zillow/zrs-profile/src/components/**/[A-Z]*.{js,jsx,ts,tsx}"
                }
            ]
        },
    ]
}
```

mendrew:
Awesome, thank you, I'll take a look.

mendrew:
Hi @stonebk. The fix has been released in [v9.1.10](https://github.com/styleguidist/react-styleguidist/releases/tag/v9.1.10).
Could you please take a look?

stonebk:
@mendrew, all better now! Thanks so much!

mendrew:
[...]

PR #1384, opened 2019-06-15
Fix section pages if section name has inappropriate symbols

mendrew:
If you have a forward slash in your section name, the pagePerSection links are all broken,
resulting in "Page not found".

For example:
  pagePerSection links break when section name includes a "/", like `name: '@foo/components'`

The solution is just to encode a name in the getUrl and then correctly
decode it in the router hash logic.

iss: https://github.com/styleguidist/react-styleguidist/issues/1332

Fix for pagePerSection links break when section name includes a  :


sapegin:
Thanks!

Issue #1381, opened 2019-06-06
Component name capitalization issue


okonet:
**Current behavior**

I've named my component `ButtonTS` but in the styleguide I see `ButtonTs`. It also affects auto-import of this component. In the `ButtonTS.md` I have

```jsx
<ButtonTS>Test</ButtonTS>
```

and this isn't working with the error `ReferenceError: ButtonTS is not defined` being shown.

If I change the example to

```diff
- <ButtonTS>Test</ButtonTS>
+ <ButtonTs>Test</ButtonTs>
```

then it works.

**To reproduce**

https://codesandbox.io/s/github/component-driven/component-driven-development/tree/typescript

**Expected behavior**

It should not change component name from `ButtonTS` to `ButtonTs`

mendrew:
I think that I found the reason, it is this function. 
https://github.com/styleguidist/react-styleguidist/blob/5ff72d8c82d11ce70c6449a77343e2cc5462c037/src/loaders/utils/getNameFromFilePath.js#L4-L11

It is used here to define `displayName`
https://github.com/styleguidist/react-styleguidist/blob/5ff72d8c82d11ce70c6449a77343e2cc5462c037/src/loaders/utils/getProps.js#L136-L139

We do `_.camelCalse` for the whole string. So `ButtonTS` becomes `ButtonTs`.

I'm looking at test cases:
https://github.com/styleguidist/react-styleguidist/blob/5ff72d8c82d11ce70c6449a77343e2cc5462c037/src/loaders/utils/__tests__/getNameFromFilePath.spec.js#L4-L24

### Questions
1. What if we just do camelCase for `-` cases?

2. At the same time this is happening because `doc.displayName` is not defined.
I see warning about this which is comming from propsParser (which is in the mentioned repo is default `react-docgen`)
```
Warning: src/components/core/ButtonTS/ButtonTS.tsx matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.
```

So, react-docgen is probably failed here. What if try `react-docgen-typescript` as a [propsParser](https://react-styleguidist.js.org/docs/webpack#create-react-app-with-typescript) 
I've tried but failed. Not really good at TS and it's configuration..

okonet:
Why do even need to modify the component's name? Is it when the parser failed to infer it from the component? I'd say in this case it should be less obtrusive and only convert if necessary.

Probably we should warn users about that behavior in the console?

mendrew:
[...]

PR #1388, opened 2019-06-21
Fix component name capitalization issue

mendrew:
iss: https://github.com/styleguidist/react-styleguidist/issues/1381

It is happening in a case when `react-docgen` (any propsParser) is failed
and we are trying to guess displayName based on the file path.

We capitalized every case in the string instead of just words separated
by `-` or started with a small letter.

So, by mistake `ButtonTS` in file name becomes `ButtonTs` display name for
component when we expect `ButtonTS`

### DIFF
--- a/src/loaders/utils/getNameFromFilePath.js
+++ b/src/loaders/utils/getNameFromFilePath.js
@@ -1,5 +1,4 @@
 const path = require('path');
-const _ = require('lodash');
 
 module.exports = function getNameFromFilePath(filePath) {
 	let displayName = path.basename(filePath, path.extname(filePath));
@@ -7,5 +6,9 @@ module.exports = function getNameFromFilePath(filePath) {
 		displayName = path.basename(path.dirname(filePath));
 	}
 
-	return _.upperFirst(_.camelCase(displayName));
+	return displayName
+		.charAt(0)
+		.toUpperCase()
+		.concat(displayName.slice(1))
+		.replace(/-([a-z])/, (_, match) => match.toUpperCase());
 };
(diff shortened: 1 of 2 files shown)


mendrew:
As a follow up of our discussion here: https://github.com/styleguidist/react-styleguidist/issues/1381#issuecomment-502829790
where the topic is 
> display an explicit warning for this case.

The only error we are getting is this one https://github.com/styleguidist/react-styleguidist/blob/5b9b517ac3a3ee6e1360ccc9acef60d3f36727a6/src/loaders/props-loader.js#L42-L46
It's pretty common and doesn't give us an idea of what is happening with displayName.

But it's might be an overhead for user to see one more warnings when it might be just an expected behavior. (when propsParser is failed and displayName is based on file path)

src/loaders/utils/getNameFromFilePath.js
```diff
@@ -1,11 +1,14 @@
...
-const _ = require('lodash');
 
 module.exports = function getNameFromFilePath(filePath) {
 	let displayName = path.basename(filePath, path.extname(filePath));
 	if (displayName === 'index') {
 		displayName = path.basename(path.dirname(filePath));
 	}
 
-	return _.upperFirst(_.camelCase(displayName));
+	return displayName
```
sapegin: Let's extract to a function (at the top of this files is fine) with some meaningful name. Ideally with a comment with some examples of what it does.

mendrew: Sure, fair enough, will extract with a comment. Thank you for taking a look!)

mendrew: I've updated the code. Could you please take a look?
I chose a bit different method which looks clear ([startCase](https://lodash.com/docs/4.17.11#startCase)). Added a comment with example.

If example is not in the right format I can change. And if previous logic is better then I can revert, no problem.

sapegin:
Thanks, merging!

PR #1389, opened 2019-06-22
Encode sublinks in url to fix broken routes

mendrew:
Follow up of https://github.com/styleguidist/react-styleguidist/issues/1332
This appears https://github.com/styleguidist/react-styleguidist/pull/1384 has fixed the top level section link,
but any sub-links within that section are still unencoded and result
in "Page not found".

fix-sublinks-routing


sapegin:
Thanks!

PR #1393, opened 2019-06-25
ci: Use a monorepo to host styleguidist

elevatebart:
Objective: enable splitting react-styleguidist into multiple packages without losing any developer experience.

- [x] Move all code into packages/react-styleguidist
- [x] Use yarn workspaces to manage dependencies
- [x] Make compile + jest tests pass
- [x] Make sure all scripts are still working the same
- [x] Fix dangerfile
- [x] cypress should pass green
- [x] Review ci to use lerna
- [x] Adapt documentation for change of log convention

NOTA: I voluntarily left out all the changes regarding extracting the core of styleguidist from the rest. I feel they would confuse proofreading.


elevatebart:
@sagepin , why is react-styleguidist supporting node v6 at unit test time, but not on integrations?

elevatebart:
Hello @sagepin , 

Once this PR is merged, the semantic release will not work. 
It is incompatible with monorepos for now. lerna has its own set of tools to do automated delivery.
By doing `lerna version --create-release=github --changelog-preset=tamia` we would be releasing exactly as before. 

Someone needs to create `conventional-changelog-tamia` as a package. and make it respect the Tamia rules. How do you want to go about doing it?

Another changed induced by this PR is the use of `yarn` for its [workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) feature. By default, yarn respects the `engine` field in package.json. It raises the question: Should we care for node 6?

There is a lot more that could be done but I think this is a safe first step.

Please tell me if you have questions

sapegin:
Thanks! 🤗 

> Once this PR is merged, the semantic release will not work.
> It is incompatible with monorepos for now. lerna has its own set of tools to do automated delivery.
> By doing lerna version --create-release=github --changelog-preset=tamia we would be releasing exactly as before.

That's the thing that bothers me the most: if releases aren't automated and can't be done from the GitHub UI, there will be no patch releases at all, because I can't do them manually.

> Someone needs to create conventional-changelog-tamia as a package. and make it respect the Tamia rules. How do you want to go about doing it?

I guess we can switch to Angular convention. It's absolutely horrible but people like it for some reason.

> Another changed induced by this PR is the use of yarn for its workspaces feature. By default, yarn respects the engine field in package.json. It raises the question: Should we care for node 6?

It's planned for the next major release.

okonet:
I’m wondering if this Pr should cover the release automation process. Can this be done on Travis CI?

elevatebart:
> I’m wondering if this Pr should cover the release automation process. Can this be done on Travis CI?

If @sagepin agrees to have a time when nothing is delivering automatically, I could see it work yes.
I can as well start a draft of `conventional-changelog-tamia`, if you are interested.

okonet:
I don’t understand. We’re talking about switching to lerna delivery process which works same way as semantic-release so why are we going to have time without ability to release?

elevatebart:
2 scenarios are possible:
1. We abandon tamia-semantic-release and use angular conventional commits. In this case no need to have time without automated delivery. I only need the e-mail of the bot to be put in an encoded variable in travis. This way the commits and releases that are sent back to github have a nice picture with it and it looks clean. (I do not need it but it would be nicer).
2. We try to keep the tamia convention in which case we have to create a package `conventional-changelog-tamia` that would be like [conventional-changelog-eslint](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-eslint) but with tamia rules.

[EDIT]: We can actually do scenario 1 with another convention from [this list](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog#preset)

sapegin:
> I don’t understand. We’re talking about switching to lerna delivery process which works same way as semantic-release so why are we going to have time without ability to release?

I don't understand that, in my experience Lerna doesn't work with semantic-release and there's no way of automating release process, as far as I know.

> We abandon tamia-semantic-release and use angular conventional commits. In this case no need to have time without automated delivery. I only need the e-mail of the bot to be put in an encoded variable in travis. This way the commits and releases that are sent back to github have a nice picture with it and it looks clean. (I do not need it but it would be nicer).

I'm fine with Angular convention, since it's the most popular one. The bot email is `artem+styleguidist@sapegin.ru`.

okonet:
I’d go with scenario 1 and Angular convention as well. It isn’t the most beautiful one but the most adopted and we should consider that if we want more people to contribute.

sapegin:
Yep, many folks use it in PRs already.

elevatebart:
@sapegin I use only lerna for vue-styleguidist. If you want an example of how lerna release works, you can have a look at Travis.yml there. 

I am using the command lerna version with the —release flag

I will finish the PR next Monday morning.

sapegin:
I'll have a look, thanks!

Could you please update docs here? https://github.com/styleguidist/react-styleguidist/blob/master/docs/Maintenance.md

elevatebart:
Hey @sapegin , 
How do you avoid publishing features while still publishing patches? 
Do you have multiple branches for features? 
Do you merge all pull requests together? 

As the system is set up now (with my update), every PR merged into master with a `feat: my nice new feature` commit message will trigger a minor release. 
Should I avoid this release trigger?
Should we have a "next" branch for the next minor release?

In vue-styleguidist I deliver minor and patches as often as I need. That does not seem to bother anyone.

okonet:
I’d do the same. To me, the process should be as simple as possible.

elevatebart:
@okonet I tend to agree. 

@sapegin I would update the maintenance docs with your permission to say exactly this.
What do you think ?

elevatebart:
Hello @sapegin @okonet

I want to check with you guys what blocked this PR.
Tell em if you have questions or concerns.  
We did not have time, or this is not valuable are completely valid answers :)

Just want to clarify if I should invest time to do the merge ;)

See you soon

sapegin:
Sorry for late reply ;-/ We didn't have time and also we had some arguments on how to make releases and use semantic-release, which I believe we've solved by now with @okonet and we have a new process now that is just default semantic-release with ESLint config.

https://github.com/styleguidist/react-styleguidist/blob/master/docs/Maintenance.md#releases

Also it would be super cool to split it into smaller pieces because I'm really struggling to review such a gigantic PR. For example, if we want to use Yarn, let's have a separate PR that only switches npm to Yarn.

What do you think @elevatebart @okonet?

sapegin:
I'm going to reopen it for better visibility.

elevatebart:
I think we should get on this when the typescript migration is done.
- The first step is indeed to switch for yarn.
- We can then have a short hangout call to see if leran release would work for react-styleguidist
- Then do the actual migration
- Finally start extracting the pieces

sapegin:
I did a bit of experimentation with my other project that uses Lerna and was able to publish it automatically on CI. A few observations:

* Conventional changelog preset doesn't work in Lerna, and looks like it never worked. So the Angular convention is the only option.
* No way to use sane casing for changelog file (`Changelog.md` instead of `CHANGELOG.md`) or better remove it entirely, since it's duplicating GitHub release notes.

And now I'm confused about Yarn workspaces. What problem does it solve if we still need Lerna?

elevatebart:
@sapegin @okonet Should I start over on this one?
I believe it might be simpler.

okonet:
If you have time I’d love someone to work on this since I have no idea how to tackle it

sapegin:
That would be awesome! Let's split it into smaller tasks, otherwise, it's impossible to review ;-/

elevatebart:
I close this PR for clean up, I will start over in a new one.

PR #1395, opened 2019-06-27
docs(webpack): improve typescript usage

cdaringe:
# problem

as discussed in https://github.com/styleguidist/react-docgen-typescript/issues/71#issuecomment-362370750, some compiler options break component discovery.  the safest strategy is to always include the tsconfig from user space, which improves component detection.

# solution

add `.withCustomConfig` call, as it's quite commonplace to for consumers to tweak their compiler settings


sapegin:
Thanks!

Issue #1189, opened 2018-10-30
Inline JavaScript comments don’t work in template.body.raw


jaller94:
Trimming template’s white spaces messes with JavaScript inline comments.

https://github.com/styleguidist/react-styleguidist/blob/e59671d4d4f1d57e89e5d731942d616458f492b4/scripts/make-webpack-config.js#L29

The line above removes all white spaces, including new lines. This comments every line after `//` inside a `<script />`.

Demo: https://github.com/jaller94/demo-react-styleguidist-js-inline-comment

**Expected behavior**
Inline comments in `script` node should not break the script.

My suggestions of how to fix this:
* Remove (inline) comments before trimming white spaces.
* Don’t trim new lines.
* Don’t trim any white spaces.
* Transform them to `/* comment */` before trimming white spaces.

sapegin:
> Don’t trim any white spaces.

The templates are usually to small to care about whitespace I think. Feel free to send a pull request!

dougmackenzie:
I've managed to reproduce this with the basic example, but changes to the make-webpack-config.js file don't seem to have any effect (so removing the `trimWhitespace: true` doesn't fix it)- what am I missing? 🤔

tonyjmartinez:
@sapegin This looks like it's actually a problem with [mini-html-webpack-plugin](https://github.com/styleguidist/mini-html-webpack-plugin). Am I on the right track?

tonyjmartinez:
Here's what I have so far. I moved the options to remove whitespace to the **html-minify** config.

```
const template = isFunction(config.template) ? config.template : MiniHtmlWebpackTemplate;
const templateContext = isFunction(config.template) ? {} : config.template;
const htmlPluginOptions = {
	context: Object.assign({}, templateContext, {
		title: config.title,
		container: config.mountPointId,
		// trimWhitespace: true,
	}),
	template: context =>
		minify(template(context), {
			collapseWhitespace: true,
			preserveLineBreaks: true,
		}),
	// template,
};
```

Here's what the result from your example looks like.

```
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<meta name="viewport" content="width=device-width,initial-scale=1">
<meta http-equiv="X-UA-Compatible" content="ie=edge">
<title>React Styleguidist Style Guide</title>
</head>
<body>
<div id="rsg-root"></div>
<script>
function works() {
    		/* Opening and closing comments are fine */
[...]

PR #1402, opened 2019-07-19
[WIP] Template inline js comments

tonyjmartinez:
#1189 
PR to fix a bug with inline JavaScript comments in template script tags. The comments were causing the rest of the script to be commented out when whitespace got removed.

My fix follows an example from [mini-html-webpack-plugin](https://github.com/styleguidist/mini-html-webpack-plugin) that uses [html-minifier](https://github.com/kangax/html-minifier). I'm using this to set options on how the HTML is minified.

I see two possible issues with what I have so far.

1. Needing to add **html-minifier** to package.json.
2. Whether or not I should set the **minifyJS** option to true. It produces cleaner minified HTML but removes comments from the JavaScript within script tags.

Here's an example where I minify the JS.

image

And here's an example where I don't.

image

### DIFF
--- a/src/scripts/make-webpack-config.js
+++ b/src/scripts/make-webpack-config.js
@@ -27,7 +27,6 @@ module.exports = function(config, env) {
 		context: Object.assign({}, templateContext, {
 			title: config.title,
 			container: config.mountPointId,
-			trimWhitespace: true,
 		}),
 		template,
 	};


docs/Webpack.md
```diff
@@ -97,7 +97,9 @@ If you’re using [Create React App](https://github.com/facebook/create-react-ap
 
    ```javascript
    module.exports = {
-     propsParser: require('react-docgen-typescript').withCustomConfig('./tsconfig.json').parse
+     propsParser: require('react-docgen-typescript').withCustomConfig(
```
tonyjmartinez: Not sure why this file keeps getting formatted. Does prettier run on commit? I don't think I changed this file.

package.json
```diff
@@ -48,6 +48,7 @@
     "github-slugger": "^1.2.1",
     "glob": "^7.1.3",
     "glogg": "^1.0.2",
+    "html-minifier": "^4.0.0",
```
tonyjmartinez: Needed to minify the html with JS in script tags. Used in example I found on mini-html-webpack-plugin

src/scripts/__tests__/make-webpack-config.spec.js
```diff
@@ -187,6 +187,29 @@ it('should pass template context to HTML plugin', () => {
 	});
 });
 
+it('should minify html with JavaScript in script tag', () => {
```
tonyjmartinez: Test that HTML minification works correctly.

src/scripts/make-webpack-config.js
```diff
@@ -21,13 +22,20 @@ module.exports = function(config, env) {
...
 	const isProd = env === 'production';
 
-	const template = isFunction(config.template) ? config.template : MiniHtmlWebpackTemplate;
+	const template = isFunction(config.template)
+		? config.template
+		: context =>
+				minify(MiniHtmlWebpackTemplate(context), {
+					collapseWhitespace: true,
+					preserveLineBreaks: true,
+					minifyJS: true,
```
tonyjmartinez: This is the option I'm not sure about. Minifying JS makes the minified HTML look better but removes comments.

sapegin:
Thanks for the pull request!

Do we really need to minify five lines of HTML code? Wouldn't be easier to disable minification?

tonyjmartinez:
@sapegin I assumed that minifying the HTML was desirable since the original code trimmed white spaces. I'll update and just remove the **trim whiteSpace: true** line from **make-webpack-config.js**. I confirmed that doing this fixes the bug with inline JS comments.

sapegin:
I think it was just a default behavior ;-) I don't see much benefit here.

tonyjmartinez:
Gotcha. I updated the commit. Let me know if anything else needs to be changed.

sapegin:
Thanks! 🍡

PR #1404, opened 2019-07-23
Docs: Add clarification to non-webpack project docs

Lasvad:
add clarification when adding webpack configs to a non-webpack project to clearly indicate what should be added to styleguide.config.js or to webpack.config.js


sapegin:
Cool, thanks!

PR #1409, opened 2019-07-26
Add example of using React Context in the docs.

jkarttunen:
Yeah, it's not perfect but it's a start. Please comment.


jkarttunen:
Sorry, pushed too many commits. Feel free to comment though. The diff should be OK, since those commits are merged already.

Issue #1321, opened 2019-03-27
Unexpected token in code example


Sly36:
I had to install react-styleguidist with yarn (vs npm) to have PROPS & METHODSi working.

Now it works but i have Syntax errors for all my components (with defaultExample set to true and components with their own examples too):

in the .md file
```
``` jsx
<Card>test</Card>
```

results in

```
SyntaxError: Unexpected token (1:6)
1 : <React.Fragment><Card>test</Card></React.Fragment>;
          ^
```

and
```
``` jsx
// comment
<Card>test</Card>
```

results in

```
SyntaxError: Unexpected token (2:11)
1 : // comment
2 : <Card>test</Card>
               ^
```

here is an example of a simple component code :

```
import React from 'react';
import PropTypes from "prop-types";
import "../../../styles/elements/card.less"

const Card = ({children, id}) =>
    <div id={id} className={"Card"}>
        {children}
    </div>;

Card.propTypes = {
    id: PropTypes.string
};

export default Card;
```

vollmerr:
I am getting this same error. Everything works fine on version 9.0.4, but get the above error on 9.0.5 for all my components.

Sly36:
Thank you, downgrading to 9.0.4 fix the problem

JonathanLehner:
downgrading works, but doesn't seem like a nice solution. what is the underlying issue?

karsai5:
Having the same issue. Also 9.0.4 gets picked up as having vulnerabilities by npm audit which doesn't make it a super viable alternative.

Betree:
It seems that this issue doesn't affect everyone so maybe we can get more info by sharing our configs. Below is ours, we use it with the following libraries (see the full list [here](https://github.com/opencollective/opencollective-frontend/blob/master/package.json)):
- `@babel/cli`: `7.1.2`
- `@babel/core`: `7.1.2`
- `@babel/node`: `7.0.0`
- `next`: `8.1.0`
- `react`: `16.8.6`

---

> https://github.com/opencollective/opencollective-frontend/blob/master/styleguide.config.js

```es6
const fs = require('fs');
const path = require('path');
const fileExistsCaseInsensitive = require('react-styleguidist/lib/scripts/utils/findFileCaseInsensitive');

module.exports = {
  assetsDir: 'styleguide',
  getExampleFilename(componentPath) {
    const parsedPath = path.parse(componentPath);
    const parentDirName = parsedPath.dir.split('src/components/')[1] || '';
[...]

PR #1412, opened 2019-07-29
Possible fix for #1321 

jkarttunen:
Replaces React.fragment short notation `<>` with long notation `<React.Fragment>`

Apparently not all the tooling supports the short notation.

### DIFF
--- a/examples/styled-components/src/StyleGuideWrapper.tsx
+++ b/examples/styled-components/src/StyleGuideWrapper.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, Fragment } from 'react';
 import ThemeProvider from './ThemeProvider';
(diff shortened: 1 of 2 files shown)


sapegin:
Could you say more about the problem you're trying to fix here? `<>` is going to be transpiled to `React.Fragment` anyway, so looks like this won't change anything.

jkarttunen:
Apparently this does cause problems for some people, with certain dependencies. In one of my projects too this prevents me from upgrading to latest styleguidist.  No idea what’s the real cause, but this looks like a probable fix. I guess I can verify it by upgrading to fixed package with npm link first. https://github.com/styleguidist/react-styleguidist/issues/1321

jkarttunen:
... looking again at the first error message, this might not help at all, it’s the same error message I get, but with long notation.

PR #1414, opened 2019-08-05
Fix: Single sections OK to contain no component

elevatebart:
If a section contains only one subsection,
styleguidist will try to optimize.
It will make the section disapear and show
its components instead.

If this section only contain sections, no components,
it should optimize to sections and not to components.

CF: https://github.com/vue-styleguidist/vue-styleguidist/issues/492

### DIFF
--- a/src/client/rsg-components/TableOfContents/TableOfContents.js
+++ b/src/client/rsg-components/TableOfContents/TableOfContents.js
@@ -41,7 +41,11 @@ export default class TableOfContents extends Component {
 		const { sections, useRouterLinks } = this.props;
 		// If there is only one section, we treat it as a root section
 		// In this case the name of the section won't be rendered and it won't get left padding
-		const firstLevel = sections.length === 1 ? sections[0].components : sections;
+		// Since a section can contain only other sections,
+		// we need to make sure not to loose the subsections.
+		// We will treat those subsecttions as the new roots.
+		const firstLevel =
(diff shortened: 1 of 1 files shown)


src/client/rsg-components/TableOfContents/TableOfContents.js
```diff
@@ -41,7 +41,8 @@ export default class TableOfContents extends Component {
 		const { sections, useRouterLinks } = this.props;
 		// If there is only one section, we treat it as a root section
```
sapegin: I guess this comment should be updated? Also this use case should still work: a style guide without any sections, just `components` field in the config — there should be no section in the sidebar.

elevatebart: Absolutely agree, that a single section containing no subsection nor content should be ignored/treated as root.
What if this top section has `content`? Should we still ignore it?

As I am adding the comments right now I am wondering.

elevatebart: I confirm that a single section containing only components is still displaying its set of components instead of its title.

sapegin: > What if this top section has content? Should we still ignore it?

I'd keep it as is for now.

sapegin:
Thanks, merging! 🌝

PR #1423, opened 2019-08-16
Fix: If components only styleguide, build TOC

elevatebart:
On the last commit, I did not take into account that the sections array could be empty and not null.
When there was components only and no sections we had the behavior described in the issue attached below.

This is the fix needed.

Sorry again

closes https://github.com/vue-styleguidist/vue-styleguidist/issues/561

### DIFF
--- a/src/client/rsg-components/TableOfContents/TableOfContents.js
+++ b/src/client/rsg-components/TableOfContents/TableOfContents.js
@@ -46,7 +46,7 @@ export default class TableOfContents extends Component {
 		// We will treat those subsecttions as the new roots.
 		const firstLevel =
 			sections.length === 1
-				? // only use subsections if there actually is subsections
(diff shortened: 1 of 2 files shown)


src/client/rsg-components/TableOfContents/TableOfContents.js
```diff
@@ -45,7 +45,12 @@ export default class TableOfContents extends Component {
 		// we need to make sure not to loose the subsections.
 		// We will treat those subsecttions as the new roots.
 		const firstLevel =
-			sections.length === 1 ? sections[0].sections || sections[0].components : sections;
+			sections.length === 1
+				? // only use subsections if there actually is subsections
```
sapegin:
```suggestion
				? // only use subsections if there actually are subsections
```

Issue #1350, opened 2019-05-03
Use Styled Components themeprovider markdown


dysticus:
**Current behavior**

I'm not sure if it's a bug, feature request or undocumented feature, but here goes.

Our styleguide will serve as a library for multiple brands. These brands share the same components, but their design tokens differ.

According to [the docs](https://react-styleguidist.js.org/docs/thirdparties.html#styled-components) you can use styled-components ThemeProvider to provide a theme to your components. I implemented as such (wrapping wrapper in the ThemeProvider) and this works (for components).

However, in my styleguide, I also have .md files for displaying design tokens. I'd like to show only those tokens, linked to that brand. To do so, I'd like to access the theme variable from styled-components ThemeProvider props, in my markdown files. Currently I haven't found a way to do so.

Actually, it's not necessary for me to stick with this approach, if it can be accomplished another way, I'm very happy to hear so. 

**To reproduce**
(Not sure if this is really needed, as I don't know how to reproduce it 😄)
- Create any react-styleguidist project with styled components.
- Overwrite your wrapper component so you can wrap it in a ThemeProvider from styled components
- Try accessing the theme in your markdown files

**Expected behavior**

I'd like to be able to access my theme from styled-components ThemeProvider or any another way to access a variable containing my theme inside the markdown files.

Thanks!

driedl:
I'd like this too. In my case, I would like to add a dropdown listing all available themes, then pass this theme to the ThemeProvider so all my components then render using the whatever theme is selected.

The theme dropdown is easy to add by implementing a custom StyleGuideRenderer that wraps the default and just adds the ddl component... but i dont know how to pass the value as a prop to the themeProvider dynamically with this new value...

tisgro:
I've proposed a general solution to the problem of multiple theme support in PR #1434 – @dysticus @driedl let me know if it works for you.

PR #1434, opened 2019-09-14
Feat: Add support for multiple themes

tisgro:
This change addresses issue #1350 by adding support for style guides with multiple themes. I had the same needs, and was able to implement multiple theme support by customizing several components, but it felt a bit hacky and brittle. I thought I'd suggest a general solution:

- add support for styling internal components per theme
- provide the current theme to external components
- implement a basic UI for switching themes

The idea is to extend the config with a "themes" options, while maintaining support for the existing "theme" option. When multiple themes are configured, a simple theme switcher is rendered in the sidebar (which can be easily customized). The current theme is stored in the StyleGuide context, so it can easily be picked up in the Wrapper component and passed to external components.

Demo: https://fancy-sg.surge.sh

If you think this is a viable direction, I'll add tests, update the docs, and move the changes to the basic example into a proper "themes" examples.


sapegin:
@tisgro thanks for the pull request! This looks pretty cool from the first sight and I think the general direction is the right one. UI also looks fine for the MVP (we can style the select later I think). For the external component theme is just a string and they can do whatever they want, right?

🦄

tisgro:
@sapegin Great! We can definitely style the select later (and, as mentioned, it's pretty easy for users to customize the component). And yes, the "theme" exposed to external components is a string (ie, the "id" values specified in the "themes" config), so users control it completely.

Since you're onboard with the change, I'll push on with adding a proper example, tests, and docs (in that order). In the meantime, if you have any feedback on the code changes to the internal components, it would be most welcome 😉

examples/basic/styleguide.config.js
```diff
@@ -11,6 +11,34 @@ module.exports = {
 		url: 'https://github.com/styleguidist/react-styleguidist',
 	},
 	version,
+	themes: [
```
sapegin: The scheme looks a bit too complex, unless there's a reason to have this much nesting.

Also it could be a good time to adopt [System UI Theme Specification](https://system-ui.com/theme/), I'm fine with breaking changes here, if necessary.

So it could look like this:

```
themes: {
  dark: {
    colors: {
					base: '#ccc',
					baseBackground: '#000',
					sidebarBackground: '#111',
					border: '#222',
    }
  }
}
```

So in the end `theme` and `themes` will have the same API. Probably we could even deprecate the `theme` and only keep `themes`.

src/client/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -47,23 +47,30 @@ export default class StyleGuide extends Component {
...
 
 	static defaultProps = {
 		displayMode: DisplayModes.all,
 	};
 
-	state = {
-		error: false,
-		info: null,
-	};
+	constructor(props, context) {
```
sapegin: I don't think you need a constructor here, `this.getDefaultTheme()` call should work with a class property too.

src/client/rsg-components/StyleGuide/StyleGuide.js
```diff
@@ -74,6 +81,24 @@ export default class StyleGuide extends Component {
...
+		} = this.props;
+
+		if (defaultTheme) {
+			return defaultTheme;
+		}
+
+		return themes && themes[0] && themes[0].id;
+	}
+
+	handleThemeSwitch(theme) {
```
sapegin: Just a preference:

```suggestion
	handleThemeChange = currentTheme => {
```

And then `this.setState({currentTheme})` and `onThemeSwitch={this.handleThemeChange}`.

src/client/rsg-components/StyleGuide/StyleGuideRenderer.js
```diff
@@ -76,10 +80,19 @@ export function StyleGuideRenderer({
 			</main>
 			{hasSidebar && (
 				<div className={classes.sidebar} data-testid="sidebar">
-					<header className={classes.logo}>
+					<header className={classes.sidebarSection}>
 						<Logo>{title}</Logo>
 						{version && <Version>{version}</Version>}
 					</header>
+					{themes && (
+						<header className={classes.sidebarSection}>
```
sapegin: We shouldn't have two headers here.

```suggestion
						<section className={classes.sidebarSection}>
```

src/client/rsg-components/Styled/Styled.js
```diff
@@ -8,15 +8,33 @@ export default styles => WrappedComponent => {
...
+
 		constructor(props, context) {
 			super(props, context);
-			this.sheet = createStyleSheet(styles, context.config || {}, componentName);
-			this.sheet.update(props).attach();
+
+			this.createSheet(context.currentTheme, props);
+		}
+
+		componentWillUpdate(nextProps, nextState, nextContext) {
```
sapegin: Let's use `componentDidUpdate` since `componentWillUpdate` is deprecated.

src/client/rsg-components/ThemeSwitcher/ThemeSwitcherRenderer.js
```diff
@@ -0,0 +1,39 @@
...
+		alignItems: 'center',
+		justifyContent: 'space-between',
+	},
+});
+
+export function ThemeSwitcherRenderer({ classes, themes, currentTheme, onThemeSwitch }) {
+	const handleThemeSwitch = e => onThemeSwitch(e.target.value);
+
+	return (
+		<span className={classes.themeSwitcher}>
```
sapegin:
```suggestion
		<label className={classes.themeSwitcher}>
```

src/client/rsg-components/ThemeSwitcher/ThemeSwitcherRenderer.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Styled from 'rsg-components/Styled';
+import map from 'lodash/map';
+
+const styles = ({ color, fontFamily, fontSize }) => ({
+	themeSwitcher: {
```
sapegin:
```suggestion
	root: {
```

examples/basic/styleguide.config.js
tisgro: Makes sense – I actually started with the scheme you suggested, but switched from an object to an array because of sorting: I thought users should be able to rely on stable sorting, which object keys don't provide. If you feel strongly that a simplified structure is more important than stable sorting, I can switch it back.

(note: in all cases, `theme` and `themes` have the same API)

I'd be happy to look into System UI and deprecating `theme`, but I'd personally prefer to put those changes into a separate PR, so these changes can get in sooner 😉

src/client/rsg-components/Styled/Styled.js
tisgro: I tried that first, but as stated here:

https://reactjs.org/docs/legacy-context.html

>Note: As of React 16, componentDidUpdate no longer receives prevContext.

It's obviously not ideal to use `componentWillUpdate`, but I think the real solution is to upgrade to the new context API… 😬

src/client/rsg-components/StyleGuide/StyleGuide.js
tisgro: Yeah, that's nicer.

examples/basic/styleguide.config.js
sapegin: I'd order themes alphabetically actually ;-/

Sure, we can add `themes` with the current API as an experimental feature.

src/client/rsg-components/Styled/Styled.js
sapegin: Oh, I guess we'll need to migrate to the new context API at some point ;-/

examples/basic/styleguide.config.js
tisgro: Haha, ok – touché ;)

But yeah, let's go with the current structure and see what the feedback is from the users.

src/client/rsg-components/Preview/Preview.js
```diff
@@ -34,12 +37,18 @@ export default class Preview extends Component {
...
+			this.state.error !== nextState.error ||
+			this.props.code !== nextProps.code ||
+			this.context.currentTheme !== nextContext.currentTheme
+		);
 	}
 
-	componentDidUpdate(prevProps) {
-		if (this.props.code !== prevProps.code) {
+	componentDidUpdate(prevProps, prevState) {
+		// Avoid executing the code when the error has changed - it will cause
```
tisgro: @sapegin this seems a bit brittle, but I found the error handling here a bit difficult to reason about, so I wanted to minimize any changes. Let me know if you prefer a different approach.

sapegin: Does it mean that the code will never update when the user fixes the error?

src/client/rsg-components/StyleGuide/StyleGuide.spec.js
```diff
@@ -146,10 +146,90 @@ it('renderer should render logo, version, table of contents, ribbon and passed c
...
+			styles: {
+				color: {
+					base: '#333',
+				},
+			},
+		},
+	];
+
+	it('renderer should render theme switcher', () => {
+		const actual = shallow(
```
sapegin: Can we avoid shallow rendering and inspecting component internals? These tests aren't testing any behaviour observable by the user. Maybe print the theme name and check the output in the test?

src/client/rsg-components/StyleGuide/__snapshots__/StyleGuide.spec.js.snap
```diff
@@ -140,3 +141,84 @@ exports[`should render components list 1`] = `
 `;
 
 exports[`should render welcome screen 1`] = `<Styled(Welcome) />`;
+
+exports[`when multiple themes are configured renderer should render theme switcher 1`] = `
```
sapegin: Let's drop this snapshot — it's not testing anything.

src/client/rsg-components/ThemeSwitcher/__snapshots__/ThemeSwitcher.spec.js.snap
```diff
@@ -0,0 +1,24 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`should render a theme switcher 1`] = `
```
sapegin: And this one too.

src/client/rsg-components/Preview/Preview.js
tisgro: The code does update when the user fixes the error, since both the previous error and the next error are the same when the code prop changes, triggering an update.

src/client/rsg-components/StyleGuide/StyleGuide.spec.js
tisgro: Ok, fair enough – I suppose it's only testing the boolean logic to render the ThemeSwitcher or not, while the user-facing behavior is tested in the ThemeSwitcher tests.

I'm not sure what you mean by printing the theme name, however… that logic is already tested in the ThemeSwitcher tests (as mentioned), right?

sapegin: Let's remove it then, if we test the functionality in the non-renderer component test.

src/client/rsg-components/ThemeSwitcher/__snapshots__/ThemeSwitcher.spec.js.snap
tisgro: I think this component should be tested in some way... I've tried to make it more meaningful – let me know what you think.

docs/Cookbook.md
```diff
@@ -226,13 +226,15 @@ See the [Preact example style guide](https://github.com/styleguidist/react-style
...
 ## How to change styles of a style guide?
 
-There are two config options to change your style guide UI: [theme](Configuration.md#theme) and [styles](Configuration.md#styles).
+There are three config options to change your style guide UI: [theme](Configuration.md#theme), [themes](Configuration.md#themes), and [styles](Configuration.md#styles).
 
-Use [theme](Configuration.md#theme) to change fonts, colors, etc.
+Use [theme](Configuration.md#theme) to apply a single set of changes to fonts, colors, etc.
 
-Use [styles](Configuration.md#styles) to tweak the style of any particular Styleguidist component.
+Use [themes](Configuration.md#themes) to configure additional themes by id. When used, this takes precedent over `theme`.
```
sapegin: I feel “by id” isn't super clear in this context, maybe drop this piece.

sapegin:
> I think this component should be tested in some way... I've tried to make it more meaningful – let me know what you think.

I was converting some tests to React Testing Library recently and realized there's no reason to test renderers separately, they are an implementation detail of their main components. 

But one snapshot less is already something so let's keep it as is for now ;-)

apennell:
@tisgro @sapegin What happened to this pr? It's exactly what I'm looking for right now! 😢  Is there anything we can do to reopen and take this live?

sapegin:
@apennell There are still some unresolved comments — feel free to finish this pull request.

PR #1445, opened 2019-10-13
Fix: Running the customized example results in an error

tonyjmartinez:
I encountered the following error running the **customized** example.

Screen Shot 2019-10-12 at 8 31 48 PM

Upgrading the packages in package.json fixed that problem. Updating the babel config fixed a problem that the package upgrade introduced with core-js.


tonyjmartinez:
I found that the same issue was happening in a few of the other examples so I went a similar route. I didn't upgrade in examples where the upgrade introduced a different problem.

elevatebart:
Hey @sapegin, 
Would it be a good time to merge this PR?
It would trigger a patch release (that failed in the previous commit since the commit message did not contain `Fix: `).

sapegin:
@elevatebart Yup, looks good now!

PR #1447, opened 2019-10-14
Docs: Fixed typos/grammar mistakes in the docs

tonyjmartinez:
I went through and fixed a handful of typos/grammar issues in the docs. Some of the changes are pretty obvious typos/grammar problems. Others are a bit more debatable, so let me know if you disagree with any of the changes.

Also, it looks like Maintenance.md uses "change log" instead of "changelog" throughout the doc. From what I've seen online, "changelog" is the preferred term so let me know if I'm okay to go through and change it there.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -55,7 +55,7 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 
 Type: `String` or `Array`, optional
 
-Your application static assets folder, will be accessible as `/` in the style guide dev server.
+Your application static assets folder will be accessible as`/`in the style guide dev server.
 
 #### `compilerConfig`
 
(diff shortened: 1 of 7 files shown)


docs/Configuration.md
```diff
@@ -55,7 +55,7 @@ By default, Styleguidist will look for `styleguide.config.js` file in your proje
 
 Type: `String` or `Array`, optional
 
-Your application static assets folder, will be accessible as `/` in the style guide dev server.
+Your application static assets folder will be accessible as`/`in the style guide dev server.
```
sapegin: Something is wrong with spaces here ;-/

```suggestion
Your application static assets folder will be accessible as `/` in the style guide dev server.
```

sapegin:
> Also, it looks like Maintenance.md uses "change log" instead of "changelog" throughout the doc. From what I've seen online, "changelog" is the preferred term so let me know if I'm okay to go through and change it there.

I did some googling and I guess you're right, let's change it too ;-)

docs/Configuration.md
tonyjmartinez: That was on purpose lol but I changed it back https://www.grammarly.com/blog/slash/

sapegin: That's a very different use case ;-)

tonyjmartinez: D'oh gotcha.

Issue #1454, opened 2019-10-25
Isolated mode example shows duplicates when component name is the same as another


chrisdopuch:
**Current behavior**

If there are two or more components with the same name in a Styleguidist project, when you click on the isolate example button the isolated page will show _all_ of the components with that name (instead of just the one you clicked on to isolate). 

Demo:
![demo](https://zappy.zapier.com/duplicate-isolated%202019-10-2510%20at%2010.43.23.gif)

**To reproduce**

My fork: https://github.com/chrisdopuch/react-styleguidist
See the `examples/sections` example, where I added a `MyLabel/Label.js` component.
Run the example, then try to click into the isolated mode like in the gif above. 

Edit: sorry, forgot to mention the repro case is on branch `isolate-duplicate-components`
I'm pushing up a fix to `fix-duplicate-isolated-components` branch right now 👍 

**Expected behavior**

I expected only the example that I clicked the isolate button to be rendered, but instead it rendered all the components with that name. 
This is a problem for larger codebases where there might be many components with the same name under different sections (e.g. teams a, b, and c all have a `Button` component). 

**Possible Solution** 

I noticed that if you preface the component name in the hash path in the URL it actually fixes the problem:
Compare http://localhost:6060/#!/Label/1 vs http://localhost:6060/#!/Components/Label/1

I believe if we update the `getUrl` util function to prefix the section name before the component name, it will resolve this problem.

Maybe there's another way to fix this via configuration, but I couldn't figure it out after pouring over the configuration docs.

Update: I submitted a PR for this here: https://github.com/styleguidist/react-styleguidist/pull/1455 

Thanks for looking! 🙏

PR #1455, opened 2019-10-25
Fix duplicate isolated components

chrisdopuch:
This PR is a suggested fix for https://github.com/styleguidist/react-styleguidist/issues/1454

The issue is that if you have more than one component with the same name, the isolated mode will incorrectly show all of the components of the same name (instead of just the one you clicked on).

This fix solves the issue by prepending the existing hash path from the page the user came from. By prepending this hash path, Styleguidist correctly scopes the isolated example to just the desired component. 

You can test this by doing `npm run start:sections` on my branch and trying to click into the isolated mode for either of the duplicate `Label` components.

Fix demo:
fixed

Open questions:
- I'm not familiar with the `noChrome` feature, I _think_ this is the correct behavior there too but I'm not 100% sure. Open to advice on how to properly manually test that piece 🙏

### DIFF
--- a/examples/sections/src/components/MyLabel/Label.js
+++ b/examples/sections/src/components/MyLabel/Label.js
@@ -1,0 +1,8 @@
+import React from 'react';
+
+/**
+ * Simple text label.
+ */
+export default function Label() {
+	return <span>I am a duplicate Label component!</span>;
+}
--- a/examples/sections/styleguide.config.js
+++ b/examples/sections/styleguide.config.js
@@ -53,6 +53,12 @@ module.exports = {
 					usageMode: 'expand', // 'hide' | 'collapse' | 'expand'
 				},
 				{
+					name: 'Labels',
+					components: () => ['./src/components/MyLabel/Label.js'],
+					exampleMode: 'expand', // 'hide' | 'collapse' | 'expand'
+					usageMode: 'expand', // 'hide' | 'collapse' | 'expand'
+				},
+				{
 					name: 'Others',
 					components: () => ['./src/components/RandomButton/RandomButton.js'],
 					exampleMode: 'collapse', // 'hide' | 'collapse' | 'expand'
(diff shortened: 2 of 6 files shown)


examples/sections/src/components/MyLabel/Label.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+/**
+ * Simple text label.
+ */
+export default function Label() {
```
chrisdopuch: This component is required to demonstrate the problem.
I am happy to remove it, or leave it in to continue to protect against this bug resurfacing in the example sections styleguide. Just let me know which you'd prefer.

examples/sections/styleguide.config.js
```diff
@@ -52,6 +52,12 @@ module.exports = {
 					exampleMode: 'expand', // 'hide' | 'collapse' | 'expand'
 					usageMode: 'expand', // 'hide' | 'collapse' | 'expand'
 				},
+				{
+					name: 'Labels',
```
chrisdopuch: Same thing here: https://github.com/styleguidist/react-styleguidist/pull/1455/files#r339249591

chrisdopuch:
👀 working on fixing the cypress tests, not sure why it's not finding the isolate button

src/client/utils/getUrl.js
```diff
@@ -34,7 +32,11 @@ export default function getUrl(
 	if (anchor) {
 		url += `#${slug}`;
 	} else if (isolated || nochrome) {
-		url += `#!/${encodedName}`;
+		const currentHashPath =
```
sapegin: Could you add some comments on what's happening here, or better extract this into a function with a meaningful name?

chrisdopuch: Good suggestion, will do

 - [ ] add comments

chrisdopuch: OK this should be resolved now, thanks @sapegin

chrisdopuch:
Cool, thanks for the approval @sapegin ! 👍 
Looks like you'll have to be the one to merge it, as I don't have write permissions to do so.

PR #1460, opened 2019-10-30
Feat: Add serverPort command line option

### LINKED ISSUE #936: Launching with command that auto configures serverPort?
Hi, 

I am working with your library more and more and its awesome! So much that i am working on multiple modules that have styleguidist. 

### My issue is that:

Following the documentation it seems that the port used by styleguidist seems to be `6060` by default and that it can only be changed with `serverPort` , in the `styleguide.config.js`.

This is a pain as I do not want to change the config just for a temporary need.

### Question / Request

Is it possible to launch the dev server by setting the port **in** the command?
ex:
 ```bash
$ npm run styleguide --port=6061
````

jKatt:
This PR adds the ability to specify the serverPort via cli and fixes #936

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -7,11 +7,12 @@
 
 ## Options
 
-| Option            | Description                              |
-| ----------------- | ---------------------------------------- |
-| `--config <file>` | Specify path to a config file            |
-| `--open`          | Open Styleguidist in the default browser |
-| `--verbose`       | Print debug information                  |
+| Option            | Description                                   |
+| ----------------- | --------------------------------------------- |
+| `--config <file>` | Specify path to a config file                 |
+| `--port <port>`   | Specify port to run the development server on |
+| `--open`          | Open Styleguidist in the default browser      |
+| `--verbose`       | Print debug information                       |
 
 ## Usage
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -537,9 +537,9 @@ Dev server hostname.
 
 #### `serverPort`
 
-Type: `Number`, default: `6060`
+Type: `Number`, default: `process.env.NODE_PORT` or `6060`
 
-Dev server port.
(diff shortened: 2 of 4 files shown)


jKatt:
Seems like it has been remove by accident with PR #1040 and got lost in the revert of that PR.

If the `NODE_PORT` option would be added again the cli option would not break old configs/scripts.
But if you specify it, it will win over `NODE_PORT` and the serverPort set in the config.

kopax:
Yes, could you please restore it. Settings such as `port` should be parameterizable from the environment for ops. It's a convention good for service in general to follow. It is required that environment priority is higher than parameter.

jKatt:
I added the option to set the serverPort via `NODE_PORT` again, but the behavior is a bit different to before, when `NODE_PORT` would only change the default value.

Now if you specify a serverPort in your config, it will not be used if you also specified `NODE_PORT`.
That can be a bit confusing, but is the easiest way to prioritize it over the command line parameter.

I think it would be better if `NODE_PORT` will be replacement of the default value for serverPort again and can be overwritten by the command line option.

sapegin:
Thanks for the PR! This is a good feature. I have two thought:

1. Probably just `--port` is enough, it's clear in the context of the CLI what kind of port it is.

2. I'm a bit unsure about the priority. I feel like the CLI parameter should take the priority even if the `NODE_PORT` happens to be in the environment, because it could be set up for another tool, and we'll have no way to explicitly define the port. Unless I'm missing something ;-)

kopax:
> 2. I'm a bit unsure about the priority. I feel like the CLI parameter should take the priority even if the `NODE_PORT` happens to be in the environment, because it could be set up for another tool, and we'll have no way to explicitly define the port. Unless I'm missing something ;-)

Sorry it was my mistake you are right. I wanted to tell that the cli parameter should be higher priority.

sapegin:
@kopax Are you happy with the logic now? ;-)

PR #1461, opened 2019-11-01
chore: add typescript

pavandv:
added Typescript Config and changed babel configuration to transpile TS, JS files.


package.json
```diff
@@ -151,7 +159,7 @@
     "start:ui": "node lib/bin/styleguidist.js server",
     "lint": "eslint . --cache --fix",
     "flow": "flow",
-    "compile": "babel --delete-dir-on-start --ignore '**/*.spec.js','**/__*__/*' -d lib/ src/",
+    "compile": "babel --delete-dir-on-start --ignore '**/*.spec.js','**/__*__/*' --extensions '.js','.ts' -d lib/ src/",
```
sapegin: We need .tsx as well.

```suggestion
    "compile": "babel --delete-dir-on-start --ignore '**/*.spec.js','**/__*__/*' --extensions '.js,.ts,.tsx' -d lib/ src/",
```

tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "es2015",
+    "jsx": "react",
```
smashercosmo: Why not 'preserve'? JSX compilation is handled by babel.

package.json
pavandv: Done 👍

tsconfig.json
pavandv: Preserve converts `.ts/.tsx` files to `.jsx`. And as we are using tsconfig only for type-checking, 'preserve/react' doesn't matter.

sapegin: What's the default value? Can we just drop it?

pavandv: Default is `preserve`. If we drop it, all files should be renamed to `.tsx` instead of `.ts`.

PR #1463, opened 2019-11-04
Feature: progress on support of typescript

elevatebart:
- [x] Add eslint validation of typescript
- [x] Set up ts-jest to start converting the tests to typescript
- [x] Generate typings in `types` folder
- [x] Update travis.yml for ci integration
- [x] Convert a few files to make sure it works


tsconfig.json
```diff
@@ -2,19 +2,25 @@
...
     "moduleResolution": "node",
     "declaration": true,
     "strict": true,
     "esModuleInterop": true,
+    "noUnusedLocals": true,
     "allowSyntheticDefaultImports": true,
     "declarationDir": "types",
-    // TODO: Remove 'allowJs' once all files are converted to TS.
-    // This is required now, for accessing types and to do typechecking for JS files.
-    "allowJs": true,
```
pavandv: This is needed until we convert every file to TS. without this check, TSC will throw errors for TS files those who imported the JS files.

tsconfig.json
```diff
@@ -2,19 +2,25 @@
...
+      "dom",
+      "esnext"
+    ],
   },
   // Remove the 'include' section if we want entire project to be Typescript.
   "include": [
     "src"
+  ],
+  "exclude": [
+    "node_modules"
```
pavandv: I believe excluding the `node_modules` here will tell the TSC not to pick any `.d.ts` files that a library/package exports.. With this we will not get typing suggestions from any 3rd Party libraries we have installed.

jsconfig.json
```diff
@@ -1,13 +0,0 @@
-{
```
pavandv: For the time being, we still need this.. Not every file will be converted in a single PR (assuming the conversation is incremental approach). We need this file for VS Code to show us IntelliSense.

.travis.yml
```diff
@@ -38,12 +38,13 @@ jobs:
 
     - stage: typecheck
       script:
-        - npm run flow check
+        - npm run typecheck
```
pavandv: For Time being, we still need this.. Once the `@flow` annotated files are converted to TS, then we can safely remove this.

.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules/
 /examples/**/styleguide/index.html
 /.publish/
 /lib
+/types
```
pavandv: Don't we need this? All the `flow types` should be converted to `TS types` and will be placed in this folder.

jsconfig.json
sapegin: If it's only for VS Code and not used for actual type check, I'd remove it.

.eslintignore
```diff
@@ -1,4 +1,5 @@
 node_modules
+examples/styled-components
```
sapegin: Any reason to ignore this one?

.eslintrc
```diff
@@ -57,6 +65,17 @@
         "rules": {
           "compat/compat": 0
         }
+    },
+    {
+      "files": ["*.ts", "*.tsx"],
+      "parser": "@typescript-eslint/parser",
+      "plugins": ["@typescript-eslint"],
+      "extends": [
```
sapegin: Let's use `tamia/typescript-react` instead of all four to match the JS rules.

package.json
```diff
@@ -102,6 +102,9 @@
     "@babel/preset-typescript": "^7.6.0",
     "@testing-library/jest-dom": "^4.2.0",
     "@testing-library/react": "^9.3.0",
+    "@types/estree": "^0.0.39",
```
sapegin: Where do we use this?

package.json
```diff
@@ -141,13 +144,16 @@
     "react-test-renderer": "^16.11.0",
     "strip-shebang": "^1.0.2",
     "style-loader": "^1.0.0",
+    "ts-jest": "^24.1.0",
```
sapegin: Could you elaborate a bit what this thing does? I've never used it. How is it different from using babel-jest to transpile TypeScript files?

package.json
```diff
@@ -141,13 +144,16 @@
...
     "typescript": "^3.6.4",
     "url-loader": "^2.2.0",
     "webpack": "^4.41.2",
     "webpack-cli": "^3.3.9"
   },
   "scripts": {
     "pretest": "npm run lint",
+    "typings": "tsc -p ./tsconfig.build.json --emitDeclarationOnly",
```
sapegin: I'm not so sure we need this here. What would be the use cases for this?

src/client/utils/transpileImports.ts
```diff
@@ -21,10 +20,16 @@ export default function transpileImports(code: string): string {
 	}
 
 	let offset = 0;
-	walk(ast, {
+	// estree walkers type is incompatible with acorns output
+	// it i sworking here out of luck and typescript is demonstrating it
```
sapegin:
```suggestion
	// it is working here out of luck and typescript is demonstrating it 
```

.travis.yml
elevatebart: You are right, let us keep them both

.gitignore
elevatebart: The types folder will contain every generated file that comes out of tsc. I usually avoid tracking build artefacts. Tu
At is a decision to be taken by @sapegin ultimately. Artem, what do you think ?

.eslintignore
elevatebart: the Typescript files in this project have never been linted as they were typescript. There is a couple of type resolving issues. I wanted to pace the changes and avoid having to deal with updating the dependencies for this example in this Pull Request. 

I would prefer this example to be felt with in it’s own Pull Request. It will be easier to review.

.eslintrc
elevatebart: Sounds great

package.json
elevatebart: This is a weird one... 
I believe one of the dependencies did not include this dependency but was referencing it. Might be estree-walker. If you remove this, you get a lot of compilation errors.

.gitignore
sapegin: I doubt we need to generate types but if we need then, yes, this should be ignored.

.eslintignore
sapegin: Could you add a comment above this line explaining that?

package.json
elevatebart: This command will generate type files. It will generate .d.ts files from the typescript. This is useful for code that plans to use rsg as an API instead of a conmand line tool. I was initially thinking of adding it to the compile command. It looked cleaner as a separate command so I thought to make it there.

Just so we are clear, the two options, noEmit and emitDeclarationOnly will throw an error if you call them together. So there was no way to combine typecheck and typings cleanly.

elevatebart: This command will generate type files. It will generate .d.ts files from the typescript. This is useful for code that plans to use rsg as an API instead of a conmand line tool. I was initially thinking of adding it to the compile command. It looked cleaner as a separate command so I thought to make it there.

Just so we are clear, the two options, noEmit and emitDeclarationOnly will throw an error if you call them together. So there was no way to combine typecheck and typings cleanly.

.eslintignore
elevatebart: Will do.

package.json
sapegin: Okay, let's keep it then if it works ;-)

sapegin: But we'll need to publish this folder to npm and add a link to `package.json` I guess.

package.json
elevatebart: I have never tried compiling typescript with Babel-jest before. For some reason it never crossed my mind. Will experiment.

package.json
elevatebart: Indeed, but for now, since we do not expose the files that are converted to ts, no typings would be useful to expose. I would almost remove anything related to typings declarations until refactoring is complete. What do you think ?

tsconfig.json
elevatebart: Did you try it ?

package.json
sapegin: Good idea! I'll add typing to the list of things to do.

pavandv: I think this is contradictory to a change made above in `.gitignore`. All the declaration files that TSC generated goes to `types` folder, right? If that is the case, what benefit are we getting by generating `.d.ts` files if we are ignoring the folder from the check-in.

.gitignore
pavandv: Please refer to this comment made below.
https://github.com/styleguidist/react-styleguidist/pull/1463#discussion_r342043092

tsconfig.json
pavandv: Not with this particular repo... but with a project I have been working in my org.

package.json
sapegin: These are type definitions for the consumers of the Styleguidist Node API, they won't be used by Styleguidist itself. And since they are generated on publish, there's no need to keep them in the repo.

elevatebart: @parandv, I believe there is a misunderstanding here. 
I just want to make sure we understand the same thing.

What is tracked on GitHub is not exactly what is published in the npm package.
In Github, we want all the code that can be modified by people and nothing more. This avoids having changes twice in each commit and makes it easier to merge conflicts.
In NPM we publish all files necessary for the usage of the package. In extenso, the contents of the `lib` directory and the contents of the `types` directory. 

If you have seen `types` folders tracked on Github on another repo, it is because this is where the types are developed. Actual people code the types in the `types` folder. 

In those repositories, there is usually no typescript files and only `.js` and `.d.ts` declaration files.

In the current PR, I will remove all typing generation from the build for now since they will benefit no-one. As @sapegin mentioned, let's have it as a TODO item.

pavandv: OK 👍 sounds good.

.eslintrc
elevatebart: done

package.json
elevatebart: all removed ;)

.travis.yml
elevatebart: done ;-)

jsconfig.json
elevatebart: @pavandv, I am a heavy user of vscode and I did not find any issues here. 
Could you be more specific on what we loose by use tsconfig instead of jsconfig?

I would gladly llok into it.

tsconfig.json
elevatebart: I agree that it is useful for vscode typings but it makes type checking impossible.
I have found a way to keep both vscode helpers and typechecking only ts files.

Thank you for the remark. 
BTW, we cannot typecheck flow files with typescript.

.gitignore
elevatebart: removed gitignore from ther as I removed typings... for now

jsconfig.json
pavandv: I was just referring that `jsconfig` would help us show IntelliSense for JS files that are not yet been converted to TS, that is all. we could delete this file once every file is converted to TS. 

I'm not sure if tsconfig could work for both JS & TS. If it does, then I'm good.

tsconfig.json
pavandv: Yes, you are right. This will make type checking impossible. I missed it. Thanks for pointing it out. 👍

elevatebart: But we still need to be able to load js files in typescript because loading a ts file in a flow checked file will fail. the no-unresolved rule of flow will break...

elevatebart: I simplified the build process with only one `tsconfig` that specifies that only `.ts` and `.tsx` files  and their dependencies should be type checked. This way we can resolve js files during the refactoring and not do it all at once.

jsconfig.json
elevatebart: It does work with js files. jsconfig is essetially a tsconfig with `"allowJs":true` ;-)

src/loaders/.eslintrc
```diff
@@ -2,6 +2,6 @@
 	"extends": "tamia",
 	"rules": {
 		"valid-jsdoc": 0,
```
sapegin: Looks like this file was / is mixing tabs and spaces ;-/

src/scripts/config.js
```diff
@@ -1,11 +1,12 @@
+import schema from './schemas/config';
```
sapegin: Could you please convert other requires to imports here too?

src/scripts/schemas/config.ts
```diff
@@ -364,3 +386,6 @@ module.exports = {
 		},
 	},
 };
+
+// export config here
```
sapegin:
```suggestion
```

Not very useful comment I think ;-)

src/typings/Component.ts
```diff
@@ -0,0 +1,13 @@
+import { Example } from './Example';
```
sapegin: What are these types?

tsconfig.json
```diff
@@ -1,20 +1,23 @@
...
-    "declarationDir": "types",
-    // TODO: Remove 'allowJs' once all files are converted to TS.
-    // This is required now, for accessing types and to do typechecking for JS files.
-    "allowJs": true,
-  },
-  // Remove the 'include' section if we want entire project to be Typescript.
-  "include": [
-    "src"
-  ]
+	"compilerOptions": {
```
sapegin: Let's keep spaces here as defined in EditorConfig.

package.json
```diff
@@ -115,7 +120,8 @@
     "cheerio": "^1.0.0-rc.3",
     "css-loader": "^3.2.0",
     "deabsdeep": "^1.0.6",
-    "deepfreeze": "^2.0.0",
+    "deep-freeze-es6": "^1.0.0",
+    "deepfreeze": "2.0.0",
```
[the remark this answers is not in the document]
elevatebart: It is still used in `.js` tests

src/loaders/.eslintrc
elevatebart: fixed

elevatebart:
Go ahead and merge it please !

src/typings/Component.ts
sapegin: Merging it now but still have this question ;-)

sapegin:
Merging, thank for the awesome work! 🦄

PR #1465, opened 2019-11-06
Chore: add typescript to tsx components

elevatebart:


src/client/rsg-components/Name/Name.spec.tsx
```diff
@@ -1,5 +1,8 @@
 import React from 'react';
+import { shallow } from 'enzyme';
 import { NameRenderer, styles } from './NameRenderer';
+// TODO: simplify this path
```
sapegin: Rewriting this test to use React Testing Library will make this line unnecessary, there's no need to test Renderers separately. (Not suggesting to do it right now ;-)

src/client/rsg-components/Name/NameRenderer.tsx
```diff
@@ -15,17 +15,26 @@ export const styles = ({ fontFamily, fontSize, color }) => ({
...
+	deprecated,
+}) => {
 	const classNames = cx(classes.name, {
 		[classes.isDeprecated]: deprecated,
 	});
 	return <code className={classNames}>{children}</code>;
-}
+};
 
 NameRenderer.propTypes = {
```
sapegin: Do you have any particular reasons to keep PropTypes in this component but remove them from other components?

src/client/rsg-components/Styled/Styled.tsx
```diff
@@ -0,0 +1,36 @@
...
+export default function StyleHOC<P extends JssInjectedProps>(
+	styles: Styles
+): (WrappedComponent: ComponentType<P>) => ComponentType<Omit<P, keyof JssInjectedProps>> {
+	return (WrappedComponent: ComponentType<P>) => {
+		const componentName = WrappedComponent.name.replace(/Renderer$/, '');
+		return class extends Component<Omit<P, keyof JssInjectedProps>> {
+			public static displayName = `Styled(${componentName})`;
+			public static contextType = Context;
+			public sheet: StyleSheet;
+			public constructor(props: Omit<P, keyof JssInjectedProps>, context: any) {
```
sapegin: Can we have types for context as well? This could prevent a lot of bugs in the future.

src/client/rsg-components/Styled/Styled.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { Component, ComponentType } from 'react';
```
sapegin: You're a hero, you did typings for the HoC! 😲

src/typings/RecursivePartial.ts
```diff
@@ -0,0 +1,6 @@
+export type RecursivePartial<T> = {
```
sapegin: Could you add some comments, what this thing does?

src/client/rsg-components/Name/NameRenderer.tsx
elevatebart: PropTypes are useful for developers that still customize RSG using javascript. It provides them with warnings if they pass arguments that they should not.

I can remove it if you prefer. I wanted to avoid loss of functionality.

src/client/rsg-components/Styled/Styled.tsx
elevatebart: I believe we can indeed, just not sure how to set it up.

src/client/rsg-components/Styled/Styled.tsx
elevatebart: Thank you

src/typings/RecursivePartial.ts
elevatebart: Ah yes... I put some comments in the Theme.ts file but forgot to comment here. 
Will do.

This is used for configuration objects. 
When using configuration objects, most of the time, all members have been given their defaults. The config object members can be considered as always defined.

User facing configs though are another item each member is non-required so instead of having 2 different intefaces, using Partial is often the proper way to go.

elevatebart:
I believe you would be ready to merge unless you prefere removing the prop types

src/client/rsg-components/Name/NameRenderer.tsx
sapegin: Make sense! We can think about generating them from TS types later.

src/client/rsg-components/Styled/Styled.tsx
sapegin: I guess similar to theme?

src/client/rsg-components/Styled/Styled.spec.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { render } from '@testing-library/react';
-import Styled from './Styled';
+import '@testing-library/jest-dom/extend-expect';
```
sapegin: Will this help to avoid importing the module in every test that uses it?

https://github.com/testing-library/jest-dom#usage

elevatebart: No, I forgot to transform jestsetup into a ts file.
It will soon though.

elevatebart: Correction, I tried doing it and TypeScript refuses to let me set up the globals for nodejs.
`global.render = render` for instance, gets the red squiggles and breaks compilation.
I believe we should simplify this once the whole applicaion is written in TS.

elevatebart: I actually cannot yet. Moving it to typescript would force me to define all the globals and set up a different tsconfig.json. I think we should focus on that once the whole codebase is typed.

sapegin: Okay, sounds good ;-)

elevatebart:
Absolutely !! Go ahead

Thank you

Bart Ledoux

Issue #1352, opened 2019-05-04
Migrate codebase to TypeScript


sapegin:
**The problem**

Recently I've added Flow to some of the files (not many) but it looks like TypeScript is more favourable now among React developers. Also full codebase conversion would help us find bugs.

**Proposed solution**

Steps to do it, all can be done one by one as separate pull requests:

- [x] Add TS as an option.
    - [x] Add `typescript` as a dependencey
    - [x] Add `tsconfig.json`
    - [x] Add `@babel/preset-typescript` to Babel config (`babel.config.js`)
    - [x] Add `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin` to ESLint config (`.eslintrc`)
    - [x] Update [Travis CI config](https://github.com/styleguidist/react-styleguidist/blob/4e0bd70d4cb19b944f787ac78390a0bb5e3136a0/.travis.yml#L39-L41) to run TypeScript type check as well.
- [ ] Replace existing Flow types with TS: update files with `@flow` comment and replace types inside `flow-types` with `@types/*` modules.
- [ ] Remove Flow from dependencies and configs.
- [ ] Convert all files to TS.

How I think it should work:

* Use TS only for type checking, keep Babel as a compiler.
* Keep sane imports syntax: `import React from 'react'`, instead of `import * as React from 'react'`.

Use [Tâmia](https://github.com/tamiadev/tamia) as an example: look at `.babelrc`, `tsconfig.json` and `package.json`.

~~Can we use [react-javascript-to-typescript-transform](https://github.com/lyft/react-javascript-to-typescript-transform) for mass conversion?~~ Not really: it needs too many manual changes after and removes all formatting including empty lines.

smashercosmo:
Do you want to use TS only as a type checker?

sapegin:
Yup, I'd like to keep Babel setup as is for now.

goku321:
I would be happy to help :)

sapegin:
@goku321 That would be great! I've updated he issue with some more details. Let me know if you have any more questions.

kakadiadarpan:
@sapegin Maybe [flow-to-ts](https://github.com/khan/flow-to-ts) can help a bit in this migration. I can try it out and see how it goes 😄

nathanredblur:
I can help too. I have experience with typescript and react. Tell me how can I start.

sapegin:
[...]

PR #1471, opened 2019-11-15
Refactor: move on the typescript transpilation of the components

elevatebart:
- Markdown
- Editor
- Argument
- Errors

#1352


src/client/rsg-components/Editor/Editor.tsx
```diff
@@ -35,16 +35,26 @@ const styles = ({ fontFamily, fontSize, color, borderRadius }) => ({
...
+export class Editor extends Component<EditorProps> {
+	public static propTypes = {
 		code: PropTypes.string.isRequired,
 		onChange: PropTypes.func.isRequired,
-		classes: PropTypes.object.isRequired,
+		classes: PropTypes.objectOf(PropTypes.string.isRequired).isRequired,
 	};
 
-	state = { code: this.props.code, prevCode: this.props.code };
+	public state = { code: this.props.code, prevCode: this.props.code };
```
sapegin: Do we need to make everything public because of the bad tests we have?

src/client/rsg-components/Message/__snapshots__/Message.spec.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`renderer should render message 1`] = `
 <div>
   <Markdown
-    text="Hello *world*!"
+    text=""
```
sapegin: Any idea why the text disappeared here? This is suspicious ;-)

src/typings/stripHtmlComments.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'strip-html-comments' {
+	function getLogger(text: string): string;
```
sapegin: Is this type correct? Why `getLogger`?

src/client/rsg-components/Message/__snapshots__/Message.spec.tsx.snap
elevatebart: Good catch, I thought that children was always an array for some reason...

src/client/rsg-components/Editor/Editor.tsx
elevatebart: @sapegin , tests have nothing to do with it
All properties that are not explicitely used by the component should be public as this is the only way React runtime can access it.

src/typings/stripHtmlComments.d.ts
elevatebart: hehe, bad copy paste.

src/client/rsg-components/Editor/Editor.tsx
sapegin: Holy cow 🐄

sapegin:
Do you want to do anything or I can merge it? 🕺

elevatebart:
Go right ahead and merge it ;-)

Bart Ledoux

sapegin:
🦄

PR #1481, opened 2019-11-22
Refactor: make client utils typescript

### LINKED ISSUE #1352: Migrate codebase to TypeScript

elevatebart:
- Made typings ambient to avoid having to import them into every component
- All files in `src/client` are now typescript

closes #1352


src/client/rsg-components/Examples/Examples.tsx
```diff
@@ -5,32 +5,24 @@ import Markdown from 'rsg-components/Markdown';
...
+function isTypedExample(ex: any): ex is Rsg.CodeExample | Rsg.MarkdownExample {
+	return !!ex.type;
+}
+
 const Examples: React.FunctionComponent<ExamplesRenderer> = ({ examples, name, exampleMode }) => {
 	const { codeRevision } = useStyleGuideContext();
 	return (
 		<ExamplesRenderer name={name}>
 			{examples.map((example, index) => {
+				if (!isTypedExample(example)) {
```
sapegin: What this piece does?

src/client/utils/__tests__/filterComponentExamples.spec.ts
```diff
@@ -1,12 +1,11 @@
-import deepfreeze from 'deepfreeze';
 import filterComponentExamples from '../filterComponentExamples';
 
-const component = deepfreeze({
+const component = {
```
sapegin: Can we use `as const` instead of deepfreeze to make sure nothing is mutating this values? We had some issues with that before ;-/

src/client/utils/__tests__/__snapshots__/getRouteData.spec.ts.snap
```diff
@@ -71,20 +71,8 @@ Object {
...
-          "module": 1,
-          "name": "Label",
-        },
-      ],
-      "exampleMode": "collapse",
-      "slug": "buttons",
-      "usageMode": "collapse",
-    },
-  ],
+  "displayMode": "notFound",
```
sapegin: Are these changes correct?

src/client/utils/filterSectionExamples.ts
```diff
@@ -0,0 +1,20 @@
+import { ExampleSectionViewModel } from '../rsg-components/Section';
```
sapegin: Isn't it `Rsg.Section` or something now?

elevatebart: Yes you are right this type should be in the `Section.d.ts` file. I need to finish moving ll types there.

src/client/utils/__tests__/filterComponentExamples.spec.ts
elevatebart: I will create typings for `deepfreeze` as I could not find any in definitelyTyped. That should allow me to keep all the `deepfreeze` in `src/client`.

src/client/utils/__tests__/__snapshots__/getRouteData.spec.ts.snap
elevatebart: no they are not (you see that snapshot testing is useful :-))

src/client/utils/filterSectionExamples.ts
elevatebart: done

PR #1484, opened 2019-11-25
Refactor: Make script folder typescript

elevatebart:


src/scripts/index.ts
```diff
@@ -1,21 +1,22 @@
...
 /**
  * Initialize Styleguide API.
  *
  * @param {object} [config] Styleguidist config.
  * @returns {object} API.
  */
-module.exports = function(config) {
-	config = getConfig(config, conf => {
-		setupLogger(conf.logger, conf.verbose, {});
+module.exports = function(configArg?: Rsg.StyleguidistConfig | string) {
```
sapegin: Does this work inside modules?

src/scripts/make-webpack-config.ts
```diff
@@ -1,22 +1,27 @@
...
-const MiniHtmlWebpackPlugin = require('mini-html-webpack-plugin');
-const MiniHtmlWebpackTemplate = require('@vxna/mini-html-webpack-template');
-const { CleanWebpackPlugin } = require('clean-webpack-plugin');
-const CopyWebpackPlugin = require('copy-webpack-plugin');
-const merge = require('webpack-merge');
-const forEach = require('lodash/forEach');
-const isFunction = require('lodash/isFunction');
-const mergeWebpackConfig = require('./utils/mergeWebpackConfig');
-const StyleguidistOptionsPlugin = require('./utils/StyleguidistOptionsPlugin');
+/* eslint-disable @typescript-eslint/camelcase */
```
sapegin: I don't see where we need this ;-/

elevatebart: I will add it next to the lines so that you can see. but overall it has to do with configuration of the Terser plugin.

src/scripts/index.ts
elevatebart: If we do not use this syntax and use `export default` instead, we break the API. 
Instead of writing

```js
const sg = require('react-styleguidist');
sg().buid()
```
users would write

```js
const sg = require('react-styleguidist').default;
sg().buid()
```

It would probably mean another Major. I prefered to wait.

elevatebart: If you want to check that it works, look at the `test/run.server.js` file that has not changed.

sapegin: Could you add a comment about that here? ;-)

elevatebart: done.

src/scripts/index.ts
```diff
@@ -11,6 +11,9 @@ import setupLogger from './logger';
 /**
  * Initialize Styleguide API.
  *
+ * NOTE: here we export an es5 module to avoid breaking the API
```
sapegin: Let's call it CommonJS module, there were no such thing as ES5 modules I think.

elevatebart:
I changed a bit the compile script as well since this was needed to avoid double mocks
Babel is annoying with globs:

https://stackoverflow.com/questions/54915206/how-do-i-tell-babel-cli-to-ignore-any-path-with-in-it

elevatebart:
I believe you can merge.

sapegin:
Awesome, thanks!

PR #1489, opened 2019-12-05
New: Add hot module reloading for theme and styles

### LINKED ISSUE #1475: Hot Module Reloading for theme and styles
Whenever I want to update the styles/theme of my style guide, I have to not only reload the page but stop and restart styleguidist entirely... not perfect.

What about passing styles and themes as references to files to require?
This way we could load them using webpack and leverage HMR on a very standard way.

We could have one file for styles and one file for themes. 
It has been my experience that one wants to reference one in the other. 
We could as well have one file for both exporting theme and style as members,

elevatebart:
closes #1475

This Pull Request allows users of `react-styleguidist` to iterate faster with the styling of their guides.
It avoids having to stop and restart the webpack-dev-server every time a change is made to the theme or the styles.

To achieve this goal, I allow the `styles` and `theme` parameters to be passed as File paths instead of an object. Webpack can then use the new files as modules and hot replace them.

- [X] create a `themed` example
- [X] adapt the `schema/config.ts`
- [x] Add documentation about new feature
- [X] change `styleguide-loader.ts` to resolve the style files using `requireIt` and set them up for hmr using adddependency and  the `hmrStyles` flag
  - [x] add tests
- [X] update `client/index.ts` to resolve es6 & es5 exports of style files and pass them on to `renderStyleguide`
- [X] add a key to Styleguide on account of the calculated `cssRevision` to force its refresh and recalc the styles
- [X] use the new context to cache-bust `createStyleSheet` memoization when styles have
 changed
  - [X] add tests


elevatebart:
I thought that I could find a simple way to decouple components from the styles and themes object. I could not without big changes to the `rsg-components` files. 

To have Webpack recompile each component's style independently (and therefore hot replace only the components impacted) the requests to the `Styled` component would have to be different per component. 

This is too bigger change than what I wanted for this PR. I will come up with another one.

This one is ready for review.

docs/Cookbook.md
```diff
@@ -260,10 +260,16 @@ module.exports = {
 }
 ```
 
-> **Note:** See available [theme variables](https://github.com/styleguidist/react-styleguidist/blob/master/src/client/styles/theme.ts).
+You can as well store those styles and theme variables like below. This has one big advantage, it avoid needing our styleguide to restart at every modification of the styles or theme.
```
sapegin: Below where? I'd be more explicit and say that you can store the styles in a separate files to allow hot reload or something like that.

examples/themed/styleguide.config.js
```diff
@@ -0,0 +1,31 @@
...
+module.exports = {
+	components: 'src/components/**/[A-Z]*.js',
+	defaultExample: true,
+	moduleAliases: {
+		'rsg-example': path.resolve(__dirname, 'src'),
+	},
+	ribbon: {
+		url: 'https://github.com/styleguidist/react-styleguidist',
+	},
+	theme: path.resolve(__dirname, './styleguide.theme.js'),
```
sapegin:
```suggestion
	theme: path.resolve(__dirname, 'styleguide.theme.js'),
```

package.json
```diff
@@ -175,7 +175,7 @@
     "start:sections": "node lib/bin/styleguidist.js server --config examples/sections/styleguide.config.js",
     "start:ui": "node lib/bin/styleguidist.js server",
     "lint": "eslint . --cache --fix --ext .js,.ts,.tsx",
-    "compile": "babel --delete-dir-on-start --ignore **/__mocks__/*,**/__tests__/*,**/*.spec.*,**/*.d.ts --extensions '.js,.ts,.tsx' -d lib/ src/",
+    "compile": "babel --delete-dir-on-start --ignore **/__mocks__/*,**/__tests__/*,**/*.spec.*,**/*.d.ts --extensions .js,.ts,.tsx -d lib/ src/",
```
sapegin: Any reason for this change?

src/client/styles/__tests__/createStyleSheet.spec.ts
```diff
@@ -1,11 +1,28 @@
+import { Styles } from 'jss';
 import * as theme from '../theme';
 import createStyleSheet from '../createStyleSheet';
 
+// eslint-disable-next-line no-var
+var mockCustomTheme: RecursivePartial<Rsg.Theme>;
```
sapegin: Would `let` work here too?

package.json
elevatebart: No js file was getting compiled for some reason. After removing the quotes it worked better. It is probably due to my OS (windows), I will remove the change.

sapegin: Ha, I'd expect it no to work on Windows without quotes ;-) Would `\".js,.ts,.tsx\"` work? But if it works without quotes, then fine.

.eslintrc
```diff
@@ -1,78 +1,74 @@
...
-    {
-      "files": ["*.ts", "*.tsx"],
-      "parser": "@typescript-eslint/parser",
-      "plugins": ["@typescript-eslint"],
-      "extends": [
-        "tamia/typescript-react"
-      ]
-    }
-  ]
+	"parser": "babel-eslint",
```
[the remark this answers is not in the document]
elevatebart: Done

docs/Configuration.md
```diff
@@ -625,16 +627,20 @@ A function that returns an HTML string, see [mini-html-webpack-plugin docs](http
...
 #### `theme`
 
-Type: `object`, optional
+Type: `Object` or `String`, optional
 
 Customize style guide UI fonts, colors, etc.
 
+Contain a theme object or a file path to a file exporting such object.
+
+The path is relative to the cwd or absolute.
```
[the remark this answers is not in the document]
elevatebart: There was a problem in my code for `schema/config.ts` I fixed it and fixed the docs

docs/Cookbook.md
elevatebart: I finally finished writing the documentation and reviewing.

src/client/styles/__tests__/createStyleSheet.spec.ts
sapegin: 🤔

elevatebart: I thought I posted something about this one, but I guess I bailed before responding.
Jest hoists all file mocks before the imports (so that they are taken into account).
If want to mock the file but still be ble to check what is inside, I need to use var instead of let or I will get a undeclared variable issue when running.

elevatebart: I guess I was looking for Kent C. Dodds tweet about jest and hoisting mocks

elevatebart: did not find it

elevatebart:
Hi @sapegin ,

I finally got around to explore the HMR part deeper.
I came up with a different approach. Tell me what you think.

This PR might seem impressive because of the number of files. Let me break it for usability's sake.
- 14 of them are a new example
- 5 are a general improvement to the system that I should probably extract
- 15 are in the src and need careful review.

I will try to remove all the changes that do not pertain to this enhancement and bring them to another PR.

elevatebart:
I removed all non-relevant changes from this PR

examples/themed/src/Components/Button/Readme.md
```diff
@@ -0,0 +1,88 @@
...
+import React from 'react'
+```
+
+Fenced blocks with other languages are rendered as highlighted code:
+
+```html
+<h1>Hello world</h1>
+```
+
+Each example has its own state that you can access at the `state` variable and change with the `setState` function. Default state is `{}`:
```
sapegin: Let's remove everything from this line — we're deprecating custom state management.

src/client/utils/getCssRevision.ts
```diff
@@ -0,0 +1,14 @@
+function hashCode(input: string): number {
```
sapegin: Could you add a comment, what's happening here?

src/client/utils/renderStyleguide.tsx
```diff
@@ -45,6 +47,7 @@ export default function renderStyleguide(
 	return (
 		<StyleGuide
 			codeRevision={codeRevision}
+			cssRevision={module.hot && styleguide.hmrStyles ? getCssRevision(styleguide.config) : 0}
```
sapegin: A comment would be nice here as well.

src/loaders/__tests__/styleguide-loader.spec.ts
```diff
@@ -184,5 +184,52 @@ it('should add common parent folder of all components to context dependencies',
...
+			theme: 'path/to/theme',
+		},
+		addDependency,
+	} as any);
+	expect(addDependency).toHaveBeenCalledWith('path/to/styles');
+	expect(addDependency).toHaveBeenCalledWith('path/to/theme');
+});
+
+it.each([
+	{ hot: true, styles: {}, RE: /'hmrStyles': false/ },
```
sapegin:
```suggestion
	{ hot: true, styles: {}, regExp: /'hmrStyles': false/ },
```

for example, `RE` is  too obscure.

src/client/index.ts
```diff
@@ -37,6 +37,18 @@ const scrollToOrigin = () => {
...
 	const styleguide = require('!!../loaders/styleguide-loader!./index.js');
+
+	// make sure the hmrStyles value does not get lost after each HMR
+	hmrStyles = hmrStyles || styleguide.hmrStyles;
+	styleguide.hmrStyles = hmrStyles;
+
+	// account for es6 exports of styles and theme files
+	styleguide.config = {
+		...styleguide.config,
+		theme: (styleguide.config.theme as any).default || styleguide.config.theme,
```
sapegin: Can we do this in `styleguide-loader.ts`?

src/loaders/styleguide-loader.ts
```diff
@@ -63,11 +64,26 @@ export function pitch(this: Rsg.StyleguidistLoaderContext) {
 		this.addContextDependency(commonDir(allComponentFiles));
 	}
 
+	let hmrStyles = false;
+
+	if (typeof config.styles === 'string') {
```
sapegin: I like this more than webpack aliases! `hmrStyles` variable is still obscure to me.

elevatebart: `hmrStyles` allowed for a way to turn off all calculations linked to hot replacing style components when `module.hot` was off or the styles were still in `styleguide.config.js`.

Unless you consider it necessary I prefer to remove it entirely. It is too confusing for a first draft. If it becomes a problem, we can assess it then.

...in two words: premature optimization

src/client/index.ts
elevatebart: With more code I can and did. I believe it would be best to do this in `requireIt()` as an option though.

elevatebart:
The whole trick of this is to calculate a hash of the styles to force the application to re-render every-time styles files change.

src/client/index.ts
elevatebart: I actually found that it is actually a lot more code to be added... like in examples-loader. I really do not like it since it is going to end up on the client anyway.
I will do as you ask, while advising against it for weight reasons.

examples/themed/src/Components/Button/Readme.md
elevatebart: agreed, done

src/client/utils/getCssRevision.ts
elevatebart: Absolutely, it is a simple md5 hash on the object

src/client/utils/renderStyleguide.tsx
elevatebart: Yes, I removed most of that complexity, I hope you like the simplifications

elevatebart:
Hello @sapegin I believe I am now done with this pass of the review and you can have another go.
To me it is ready to merge unless you see something that needs fixing.

src/loaders/styleguide-loader.ts
```diff
@@ -64,29 +69,51 @@ export function pitch(this: Rsg.StyleguidistLoaderContext) {
...
+
+			// then create a variable to contain the value of the theme/style
+			styleContext.push(resolveES6(configMember, varName));
+
+			// Finally assign the calcultaed value to the member of the clone
+			// NOTE: if you are mutating config without cloning it, it changes
+			// the value for all config untill the process is stopped.
+			// This means `config` should not be mutated during this process.
+			const variableAst = {};
+			// we trick the toAst() function here because we want to force the value in it
```
sapegin: I don't understand this comment ;-/

src/loaders/styleguide-loader.ts
```diff
@@ -64,29 +69,51 @@ export function pitch(this: Rsg.StyleguidistLoaderContext) {
...
-	}
-	if (typeof config.theme === 'string') {
-		this.addDependency(config.theme);
-		// TODO: need to do the same as in examples-loader for managing default export
-		config.theme = requireIt(config.theme) as any;
-	}
+	const configClone = { ...config };
+	const styleContext: t.VariableDeclaration[][] = [];
+
+	const setVariableValueToObjectInFile = (
```
sapegin: This function will be much easier to understand with an example of code it generates. Though in general it's a big improvement over the previous iteration.

src/client/index.ts
sapegin: What's going to end up in the client? I'm confused here ;-/

src/loaders/styleguide-loader.ts
sapegin: Let's get rid of it!

src/client/utils/getCssRevision.ts
sapegin: I'd rather use something like https://github.com/bevacqua/hash-sum instead of a custom implementation.

src/loaders/styleguide-loader.ts
```diff
@@ -27,6 +30,9 @@ const CLIENT_CONFIG_OPTIONS = [
 	'version',
 ];
 
+const StylesVariableName = '__rsgStyles';
```
sapegin: Let's not use class name convention for constants.

```suggestion
const STYLE_VARIABLE_NAME = '__rsgStyles';
```

src/loaders/utils/resolveES6.ts
```diff
@@ -0,0 +1,26 @@
+import { builders as b } from 'ast-types';
```
sapegin: `resolveESModule` could be a better name, ES6 isn't specific enough.

src/client/utils/getCssRevision.ts
elevatebart: I was looking for a package like that, but did not find one. I guess I was not looking for the right keywords. Thank you very much this simplifies things a ton.

src/client/index.ts
elevatebart: In one case we have `theme.default || theme` ending up in the final bundle sent to the client.
If we do the resolving of default in the loader we end up with a code like this, 

```js
var theme$0 = require("path/to/theme");
var themeVar = theme$0.default || theme$0;
//...
module.export = {
  theme: themeVar, // instead of theme: require('path/to/theme')
}
```

But having the resolution fully in the loader will bring less confusion.

src/loaders/styleguide-loader.ts
elevatebart: agreed

src/loaders/utils/resolveES6.ts
elevatebart: done

src/client/index.ts
sapegin: Yeah, I think the difference is too tiny to worry about it. And it's good to have related code in the same place.

src/client/utils/renderStyleguide.tsx
```diff
@@ -45,6 +47,13 @@ export default function renderStyleguide(
 	return (
 		<StyleGuide
 			codeRevision={codeRevision}
+			// only caclulate css revisions in dev when hot is on to avoid
+			// stringifying the styles in production
+			cssRevision={
+				module.hot
+					? hashSum({ theme: styleguide.config.theme, styles: styleguide.config.styles })
```
sapegin: We're already destructuring config above, so we can add `theme` and `styles` there too.

```suggestion
					? hashSum({ theme, styles })
```

src/loaders/utils/__tests__/resolveES6.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { generate } from 'escodegen';
+import { builders as b } from 'ast-types';
+import resolveESModule from '../resolveESModule';
+
+it('resolveES6 should return an array of AST', () => {
```
sapegin:
```suggestion
it('should return an array of AST', () => {
```

And we need to rename the file too.

elevatebart:
Finally done all the last fixes, we should be closer to the finish line

src/client/utils/renderStyleguide.tsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
+import hashSum from 'hash-sum';
 import slots from 'rsg-components/slots';
 import StyleGuide from 'rsg-components/StyleGuide';
 import getPageTitle from './getPageTitle';
 import getRouteData from './getRouteData';
 import processSections from './processSections';
 
 interface StyleguideObject {
+	hmrStyles: boolean;
```
sapegin: Looks like we should remove this now?

```suggestion
```

src/loaders/utils/__tests__/resolveESModule.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { generate } from 'escodegen';
+import { builders as b } from 'ast-types';
+import resolveESModule from '../resolveESModule';
+
+it('should return an array of AST', () => {
+	const result = resolveESModule('path/to/module', 'NameOfVar');
+
+	expect(generate(b.program(result))).toMatchInlineSnapshot(`
```
sapegin: That's a good way to test AST! 👍

src/loaders/styleguide-loader.ts
elevatebart: I tried explaining it better it might not be enough. hope it will be

src/loaders/styleguide-loader.ts
elevatebart: I added a comment to explain and a unit test to present it

elevatebart:
@sapegin I did it all now, finally, I am open to questions.

src/loaders/styleguide-loader.ts
```diff
@@ -73,6 +73,25 @@ export function pitch(this: Rsg.StyleguidistLoaderContext) {
...
+	 * }
+	 * ```
+	 * into this code
+	 * ```
+	 * {
+	 *  param: require('test/path')
+	 * }
+	 * ```
+	 * but because we have to account for ES module exports,
+	 * we add an extra step and transform it into something else
```
sapegin: “something else” — we don't know what this is? 😕

elevatebart: Agreed and fixed

elevatebart: @sapegin I have been a bit more thorough on the comments.
Thanks for the constructive criticism. 
We will make user styling incredible together.

elevatebart:
@sapegin do you think I can merge this one?

sapegin:
Yup, go for it!

PR #1490, opened 2019-12-06
New: Accept styles config as a function

### LINKED ISSUE #1474: `Styles` option as a function
When we use `styles` customization, we cannot reference anything from the theme. This means that everything that is changed has to be changed in multiple places.

It could be useful to allow developers to use a function to set styles in their config instead of a pure object. My idea was something like this:

```js
// styledguide.config.js
module.exports = {
  styles({color}){
    Table: {
      tableHead: {
        borderBottom: [[1, 'solid', color.border]]
      }
    }
  }
}
```

elevatebart:
closes #1474

- [X] styles can now accept a function
- [X] tests are written and are passing
- [X] added documentation for the feature in `Configuration.md` and `Cookbook.md`
- [x] update the `customised` example to illustrate the use of the new styles

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -605,7 +605,7 @@ Customize styles of any Styleguidist’s component.
 
 See example in the [cookbook](Cookbook.md#how-to-change-styles-of-a-style-guide).
 
-> **Note:** Using a function allows access to theme variables as seen in the example below. See available [theme variables](https://github.com/styleguidist/react-styleguidist/blob/master/src/client/styles/theme.ts). The returned object remains the same as the styles.
+> **Note:** Using a function allows access to theme variables as seen in the example below. See available [theme variables](https://github.com/styleguidist/react-styleguidist/blob/master/src/client/styles/theme.ts). The returned object folows the same format as when configured as a litteral.
 
 ```javascript
 module.exports = {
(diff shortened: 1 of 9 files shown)


docs/Configuration.md
```diff
@@ -599,12 +599,29 @@ Folder for static HTML style guide generated with `styleguidist build` command.
...
 #### `styles`
 
-Type: `object`, optional
+Type: `Object` or `Function`, optional
 
 Customize styles of any Styleguidist’s component.
 
 See example in the [cookbook](Cookbook.md#how-to-change-styles-of-a-style-guide).
 
+> **Note:** Using a function allows access to theme variables as seens in the example below. See available [theme variables](https://github.com/styleguidist/react-styleguidist/blob/master/src/client/styles/theme.ts). The returned object remains the same as the styles.
```
sapegin:
```suggestion
> **Note:** Using a function allows access to theme variables as seen in the example below. See available [theme variables](https://github.com/styleguidist/react-styleguidist/blob/master/src/client/styles/theme.ts). The returned object remains the same as the styles.
```

elevatebart:
should I merge?

sapegin:
Sure! We're temporarily on ESLint commit convention (before I change it again to Angular :-) so it should be `New: ....`.

sapegin:
https://github.com/styleguidist/react-styleguidist/blob/master/docs/Maintenance.md#commit-message-conventions

PR #1491, opened 2019-12-09
update copy-webpack-plugin

narehart:
updates `copy-webpack-plugin` which includes latest version of `serialize-javascript` that fixes a security vulnerability: https://github.com/advisories/GHSA-h9rv-jmmf-4pgx

### DIFF
--- a/src/scripts/make-webpack-config.ts
+++ b/src/scripts/make-webpack-config.ts
@@ -96,15 +96,17 @@ export default function(
 					root: config.styleguideDir,
 					verbose: config.verbose === true,
 				} as any),
-				new CopyWebpackPlugin(
-					config.assetsDir ? castArray(config.assetsDir).map(dir => ({ from: dir })) : []
-				),
 			],
 			optimization: {
 				minimize: config.minimize === true,
 				minimizer: [minimizer],
 			},
 		});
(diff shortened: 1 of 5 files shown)


src/scripts/make-webpack-config.ts
```diff
@@ -96,15 +96,17 @@ export default function(
...
-				new CopyWebpackPlugin(
-					config.assetsDir ? castArray(config.assetsDir).map(dir => ({ from: dir })) : []
-				),
 			],
 			optimization: {
 				minimize: config.minimize === true,
 				minimizer: [minimizer],
 			},
 		});
+		if (config.assetsDir && webpackConfig.plugins) {
```
sapegin: Could you explain these changes and also the test removal?

narehart: @sepegin Version 5.1.0 is now using `schema-utiles` to validate passed in options and throws a `ValidationError` when passing an empty array. Refactored to reflect this and removed test because it was asserting that the plugin was added when the array was empty which is no longer the case in this PR.
https://github.com/webpack-contrib/copy-webpack-plugin/pull/419/files#diff-1fdf421c05c1140f6d71444ea2b27638R13
https://github.com/webpack/schema-utils

PR #1504, opened 2019-12-26
Honor default exports over named exports

Lazyuki:
When you have a component like 

```tsx
import React from 'react';
import styled from 'styled-components';

export const C: React.FC = () => {
  ...
}

const StyledC = styled(C)``;

export default StyledC;
```
so that you can use non-styled (name-exported) components for testing, styleguidist picks up the named export instead of the default export and breaks the styling in examples. 

If I understand [this part of the doc](https://react-styleguidist.js.org/docs/components.html#default-vs-named-exports) correctly, isn't this supposed to be the case? As in it will try to find
1. default exports 
2. if not found => single named export
3. if not found => named export with the same name

I'm aware there are a few workarounds to get my example above working, but just in case the current behavior isn't intended.

### DIFF
--- a/src/loaders/__tests__/examples-loader.spec.ts
+++ b/src/loaders/__tests__/examples-loader.spec.ts
@@ -180,7 +180,7 @@ it('should prepend example code with React require()', () => {
 	expect(result).toBeTruthy();
 	expect(() => new Function(result)).not.toThrowError(SyntaxError);
 	expect(result).toMatch(
-		`const React$0 = require('react');\\nconst React = React$0['React'] || (React$0.default || React$0);`
+		`const React$0 = require('react');\\nconst React = React$0.default || (React$0['React'] || React$0);`
 	);
 });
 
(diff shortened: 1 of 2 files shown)


sapegin:
Thanks for the pull request!

I guess you're right, this looks more reasonable. You need to fix the CI though ;-)

elevatebart:
@sapegin this could be a breaking change and we have 11 around the corner, do you think it should be part of this new major version?

sapegin:
@elevatebart Yeah, make sense! If it's going to be ready soon.

Lazyuki:
@sapegin thanks for the merge. Do you have a rough ETA on v11?

sapegin:
There's a beta version:

https://github.com/styleguidist/react-styleguidist/releases/tag/v11.0.0-beta2

Lazyuki:
oh thanks!

Lazyuki:
Hey @sapegin, I see that the 11 branch is updated with current master+new v11 features. Could you release v11.0.0-beta3 with these updated changes on master? Much appreciated!

sapegin:
I think I'll try to make a final release with what we have in the 11 branch.

Lazyuki:
Oh that's great news!

PR #1510, opened 2020-01-03
Fix crashes when DefaultProp isn't a string

Lazyuki:
Styleguidist crashes when default prop isn't a string, e.g. a number.
image

### DIFF
--- a/src/client/rsg-components/Props/renderDefault.tsx
+++ b/src/client/rsg-components/Props/renderDefault.tsx
@@ -9,18 +9,19 @@ export default function renderDefault(prop: PropDescriptorWithFlow): React.React
 	// Workaround for issue https://github.com/reactjs/react-docgen/issues/221
 	// If prop has defaultValue it can not be required
 	if (prop.defaultValue) {
+		const defaultValueString = typeof prop.defaultValue.value === 'string' ? showSpaces(unquote(prop.defaultValue.value)) : showSpaces(String(prop.defaultValue.value));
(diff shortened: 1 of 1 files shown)


src/client/rsg-components/Props/renderDefault.tsx
```diff
@@ -9,18 +9,19 @@ export default function renderDefault(prop: PropDescriptorWithFlow): React.React
 	// Workaround for issue https://github.com/reactjs/react-docgen/issues/221
 	// If prop has defaultValue it can not be required
 	if (prop.defaultValue) {
+		const defaultValueString = typeof prop.defaultValue.value === 'string' ? showSpaces(unquote(prop.defaultValue.value)) : showSpaces(String(prop.defaultValue.value));
```
sapegin: I guess we don't really need a condition here?

```suggestion
		const defaultValue = showSpaces(unquote(String(prop.defaultValue.value)));
```

Lazyuki: Yeah that works too. Although I just realized if the defaultValue isn't a primitive type like an object it would probably show `[object Object]` as the default prop?

Lazyuki:
BTW if this gets merged to `master` can you also update the `11` branch if possible?

Lazyuki:
@sapegin sorry to bother you but would it be possible to merge `master` into `11` and release `11.0.0-beta3`? I see that there's another commit on master so I'm not sure if you are willing to merge that just yet.

Issue #1508, opened 2020-01-03
v11 automatic Typescript support does not play well with 3rd party components


dysticus:
**Current behavior**

RSG v11's automatic Typescript support does not seem to play well with 3rd party components. 
In <v11, in combination with react-docgen-typescript, props and methods were correctly parsed after setting the propsparser property in the config. From > v11 react-docgen was upgraded so react-docgen-typescript was no longer required.

**To reproduce**

Clone https://github.com/dysticus/example . It contains 2 branches: notworking and working. Not working is with v11, working is with <v11 and react-docgen-typescript.

Start the styleguide. You'll see in the the non-functional version no props and no methods, in the working branch they do appear.

**Expected behavior**

Props and methods should be extracted for 3rd parties to, like in versions before v11.

Thanks!

thecodejack:
I see this warning in non working branch which means `react-docgen` unable to parse the component. Can you please raise the issue with `react-docgen`?

![image](https://user-images.githubusercontent.com/1515160/71775091-c7b4a900-2fa0-11ea-8066-c0c137c12a74.png)

thecodejack:
Did a bit a of code walkthrough and realized, `react-docgen` never supported thirdparty lib parsing. As mentioned in https://react-styleguidist.js.org/docs/thirdparties.html#how-styleguidist-works , it always parses static files we produce. But with `react-docgen-typescript`, it checks for types / interfaces and them for prop parsing. 

I don't see `react-docgen` fixing the issue anytime soon. Probable solution I can see is, suggest `react-docgen-typescript` for third party component parsing like before in documentation with the release.

What do you suggest @sapegin ?

sapegin:
Yeah, I guess properly document what's not supported by react-docgen and supported by react-docgen-typescript would be good start.

PR #1514, opened 2020-01-07
doc update for typescript props config

thecodejack:
documentation update for #1508

### DIFF
--- a/docs/Thirdparties.md
+++ b/docs/Thirdparties.md
@@ -54,7 +54,7 @@ module.exports = {
 
 Please see our [examples](https://github.com/styleguidist/react-styleguidist/tree/master/examples) and refer to [react-docgen](https://github.com/reactjs/react-docgen) documentation for more information about what types of syntax are supported.
 
-But while writing components using typescript and extending typescript based third party component like below example, You can still be able to fetch prop definitions of the third party component using `react-docgen-typescript`.
+But while writing components using typescript and re-exporting a component from node_modules, You can still be able to fetch prop definitions of the parent component using `react-docgen-typescript`.
 
 ```javascript
 import Button from 'antd/es/button'
@@ -62,7 +62,7 @@ import Button from 'antd/es/button'
 export default Button
 ```
 
-In the above example, we were trying to extend a third party component. Now if we want prop definitions of third party component, `react-docgen` won't be able to fetch them. For same you can setup `react-docgen-typescript` with `styleguidist` by following below steps.
+In the above example, we were trying to extend a component which is from node_modules. Now if we want prop definitions of the parent component, `react-docgen` won't be able to fetch them. For same you can setup `react-docgen-typescript` with `styleguidist` by following below steps.
 
 1. Install [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript).
 2. Create a `styleguide.config.js`, see [configuration](Configuration.md) reference.
--- a/docs/Thirdparties.md
+++ b/docs/Thirdparties.md
@@ -54,7 +54,15 @@ module.exports = {
 
 Please see our [examples](https://github.com/styleguidist/react-styleguidist/tree/master/examples) and refer to [react-docgen](https://github.com/reactjs/react-docgen) documentation for more information about what types of syntax are supported.
 
-But if components are written with typescript and extending other components (which are also written using typescript), You may use `react-docgen-typescript` module for fetching extended component prop definitions automatically. Follow the steps to configure same.
+But while writing components using typescript and extending third party component like below example, You can still be able to fetch prop definitions of the third party component using `react-docgen-typescript`.
+
+```javascript
+import Button from 'antd/es/button'
+
+export default Button
+```
+
+In the above example, we were trying to extend a third party component to our component library. Now if we want prop definitions of third party component, `react-docgen` won't be able to fetch them. For same you can setup `react-docgen-typescript` with `styleguidist` by following below steps.
 
 1. Install [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript).
 2. Create a `styleguide.config.js`, see [configuration](Configuration.md) reference.
--- a/docs/Thirdparties.md
+++ b/docs/Thirdparties.md
@@ -54,7 +54,7 @@ module.exports = {
 
 Please see our [examples](https://github.com/styleguidist/react-styleguidist/tree/master/examples) and refer to [react-docgen](https://github.com/reactjs/react-docgen) documentation for more information about what types of syntax are supported.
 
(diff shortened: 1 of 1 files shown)


docs/Thirdparties.md
```diff
@@ -54,6 +54,20 @@ module.exports = {
 
 Please see our [examples](https://github.com/styleguidist/react-styleguidist/tree/master/examples) and refer to [react-docgen](https://github.com/reactjs/react-docgen) documentation for more information about what types of syntax are supported.
 
+But if components are written with typescript and extending third party library components (which are also written using typescript), You may use `react-docgen-typescript` module for fetching third party component prop definitions. Follow the steps to configure same.
```
sapegin: Does third-party mean they are imported from another file or specifically installed from npm?

thecodejack: it mainly with `components installed from npm`.

thecodejack: *Update* just checked. it's actually both.

sapegin: Then I'd make it more explicit — “third-party components” can mean different things, and here just the fact of importing from another file is important.

thecodejack: yea...should I mention "props of extended components"

thecodejack: How about this?

```
But if components are written with typescript and extending another component (which is also written using typescript), You may use `react-docgen-typescript` module for fetching third party component prop definitions. Follow the steps to configure same.
```

thecodejack: done

docs/Thirdparties.md
```diff
@@ -54,6 +54,20 @@ module.exports = {
 
 Please see our [examples](https://github.com/styleguidist/react-styleguidist/tree/master/examples) and refer to [react-docgen](https://github.com/reactjs/react-docgen) documentation for more information about what types of syntax are supported.
 
+But if components are written with typescript and extending other components (which are also written using typescript), You may use `react-docgen-typescript` module for fetching extended component prop definitions automatically. Follow the steps to configure same.
```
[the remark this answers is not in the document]
thecodejack: context was set before. Example of how it doesn't work probably unnecessary.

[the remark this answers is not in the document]
sapegin: It is very much "related". I still don't understand what's the problem exactly, and, looking at the comments, you probably don't understand either. We can't expect users to understand it.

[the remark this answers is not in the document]
thecodejack: Have done the changes. Sorry for resolve bit late.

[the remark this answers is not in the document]
sapegin: Thanks, but it still doesn't answer my question: what exactly isn't working — importing any component or specifically from `node_modules`. From what you've said before I understood it's the former, but from the text it looks like the latter.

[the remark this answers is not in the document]
thecodejack: actually issue seems to be from both any component and node_modules. React doc gen doesn't support. But with `react-docgen-typescript`, both will work.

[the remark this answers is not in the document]
sapegin: So something like this also won't work?

```javascript
import Button from './AnotherButton'
export default Button
```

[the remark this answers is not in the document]
thecodejack: My bad. I was initially discussing with what I am aware of. So I sent sometime around this. So here is a detailed info.

First it's surely **not working with extended third party components**.

Example: Please check https://github.com/styleguidist/react-styleguidist/issues/1508#issue-544978200

Second, components which are extended from internal ones are working fine.

Example: https://github.com/styleguidist/react-styleguidist/tree/11/examples/styled-components/src/components/Heading

**But** I was checking wrong example which didn't work.

Example: https://github.com/styleguidist/react-styleguidist/tree/11/examples/styled-components/src/components/Button

This one didn't work most probably coz it is using style component declaration which is not a standard React component creating pattern. 

Now checked above directly using `react-docgen-typescript` as well. Apart from displaying the 3 declared props, it also displayed props of styled component which is again 3rd party component.

There is lot of wiring and ifs/buts around this. But what I can summarize from this, is to inform user that **for fetching props of third party components only, we should recommend users to use `react-docgen-typescript`.** 

Let know your thoughts around this.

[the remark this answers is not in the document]
sapegin: > First it's surely not working with extended third party components.

That's clear now, but let's not call then "third-party", since it might be ambiguous. "Components rexeported from node_modules" would be more precise.

> Second, components which are extended from internal ones are working fine.
> Example: https://github.com/styleguidist/react-styleguidist/tree/11/examples/styled-components/src/components/Heading

This component doesn't extend another component.

> Now checked above directly using react-docgen-typescript as well. Apart from displaying the 3 declared props, it also displayed props of styled component which is again 3rd party component.

There's a way to filter props in react-docgen-typescript: 

https://github.com/tamiadev/tamia/blob/038390016cbe62480dc385c1656c6f5b4e1fe53f/styleguide.config.js#L10-L18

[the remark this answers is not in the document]
thecodejack: >This component doesn't extend another component.

`index.ts` of [Heading Component](https://github.com/styleguidist/react-styleguidist/tree/11/examples/styled-components/src/components/Heading) is importing component from `Heading.tsx`. So something like below will also work.

```js
import Button from './AnotherButton'
export default Button
```

Have done the changes as suggested by you.

docs/Thirdparties.md
```diff
@@ -54,6 +54,28 @@ module.exports = {
 
 Please see our [examples](https://github.com/styleguidist/react-styleguidist/tree/master/examples) and refer to [react-docgen](https://github.com/reactjs/react-docgen) documentation for more information about what types of syntax are supported.
 
+But while writing components using typescript and re-exporting a component from node_modules, You can still be able to fetch prop definitions of the parent component using `react-docgen-typescript`.
```
sapegin:
```suggestion
While Styleguidist supports TypeScript out of the box, thanks to `react-docgen`, this support is limited. Consider this example:```

File: docs/Thirdparties.md [hunk 4]
```diff
@@ -54,6 +54,28 @@ module.exports = {
...
+
+```javascript
+import Button from 'antd/es/button'
+
+export default Button
+```
+
+In the above example, we were trying to extend a component which is from node_modules. Now if we want prop definitions of the parent component, `react-docgen` won't be able to fetch them. For same you can setup `react-docgen-typescript` with `styleguidist` by following below steps.
```
sapegin:
```suggestion
Here we’re reexporting a third-party component from `node_modules`. Styleguidist won’t be able to render prop types of this component, unless we’re using `react-docgen-typescript`:
```

thecodejack: Yep. Suggestions totally right. Updated with suggestions.

sapegin:
Thanks, merging!

PR #1515, opened 2020-01-08
Build: Avoid compiling spec files when publishing

elevatebart:
When running `npm run compile` all spec.ts and tsx files are compiled into the lib folder which result in publishing tests to npm.

I remove this compilation, to make the command faster and avoid conflicts with other libraries.


elevatebart:
Hey @sapegin, do you mind if I merge this one ?

PR #1522, opened 2020-01-17
Build: Make sure compile works on windows

elevatebart:
the npm run compile fails on windows 10. 
The single quote are not something that the cmd terminal accepts easily.
Since the single quotes are not necessary on linux or macos, we could get rid of them.


elevatebart:
This does not appear like  big deal but working on windows is very common for me.

sapegin:
🏙

PR #1537, opened 2020-02-12
Fix: Open all sections when searching

elevatebart:
fix the Error reported in the issue below 

https://github.com/vue-styleguidist/vue-styleguidist/issues/750

When searching with the option `tocmode: 'collapse'` the sections do not open automatically (because they keep their state).
The found section or page is therefore not visible nor clickable.


sapegin:
Thanks! 🌮

Issue #1542, opened 2020-02-24
Code example does not render when the code block has `tsx padded`


mitsuruog:
**Current behavior**

Code example does not render when the code block has `tsx padded`.

![inspiring-cdn-v779f - CodeSandbox](https://user-images.githubusercontent.com/1703219/75200002-f0d70600-57a7-11ea-9444-7c9cf4bf182f.png)

**To reproduce**

On the TypeScript project, I put some example code in `the README.md` that has the below code block.

<pre><code>```tsx padded

example here

```</code></pre>

(NOTE: I got the same result with `typescript padded`)

please see this sandbox.

https://codesandbox.io/s/inspiring-cdn-v779f

dependencies

    "react-styleguidist": "10.6.2",
    "react-docgen-typescript": "1.16.2",
    "typescript": "3.8.2",


**Expected behavior**

<!-- A clear and concise description of what you expected to happen -->

It should render as a real example. or the `tsx` and `typescript` both do not support officially.

It might be related or the same to this issue anyway.

- Code examples with language set to `typescript jsx` does not render the example #1540

Thanks.

mitsuruog:
It will be fixed in the next major version 11.

PR #1543, opened 2020-02-27
Add typescript/ts/tsx extension as a playground languages

mitsuruog:
Resolve:

- Code example does not render when the code block has `tsx padded` #1542

I added `typescript`, `ts`, `tsx` as a playground language.

It might make breaking current behavior. 
it's up to you whichever it goes to the current version or next major version.


src/loaders/utils/__tests__/chunkify.spec.ts
```diff
@@ -45,6 +45,39 @@ This should be highlighted too:
...
+<h3>Hello playground!</h3>
+\`\`\`
+
+\`\`\`tsx
+<h3>Hello playground!</h3>
+\`\`\`
+`;
+
+	const actual = chunkify(markdown);
+	expect(actual).toMatchSnapshot();
```
sapegin: Let's avoid a snapshot here: it's hard to see if it's correct or not.

src/loaders/utils/__tests__/chunkify.spec.ts
```diff
@@ -45,6 +45,39 @@ This should be highlighted too:
 	expect(actual).toMatchSnapshot();
 });
 
+it('should some custom extensions render as a playground', () => {
```
sapegin: Maybe?

```suggestion
it('should render some extensions as a playground', () => {
```

src/loaders/utils/__tests__/chunkify.spec.ts
mitsuruog: so should I write the test something like this?

```
const expected = [
  // expected result here
];
const actual = chunkify(markdown);
expect(actual).toEqual(expected);
```

sapegin: Here you can check all of the chunks, except the first one, have the `code` type — that's the intent of the test. With a snapshot, it's impossible to tell what was the intent and whether it's correct.

mitsuruog:
@sapegin I updated my test code. please review me again.

src/loaders/utils/__tests__/__snapshots__/chunkify.spec.ts.snap
```diff
@@ -146,4 +146,4 @@ This should be highlighted too:
     "type": "markdown",
   },
 ]
-`;
+`;
```
mitsuruog: ah... should I revert it?

sapegin: It's not merged yet, so just update the test ;-)

mitsuruog:
@sapegin I updated my test code. please review me again.

src/loaders/utils/__tests__/chunkify.spec.ts
```diff
@@ -50,32 +50,69 @@ it('should render some extensions as a playground', () => {
...
+		{
+			type: 'code',
+			content: '<h3>Hello tsx playground!</h3>',
+			settings: {},
+		},
+	];
+
 	const actual = chunkify(markdown);
-	expect(actual).toMatchSnapshot();
+	expect(actual).toEqual(expected);
```
sapegin: Why not something like, so the test intent is actually clear?

```suggestion
	expect(actual.slice(1).every(chunk => chunk.type === 'code')).toBe(true)
```

Not it's exactly the same as it was with a snapshot.

mitsuruog: I don't copy as is what the snapshot did. I thought the current tests also can test for not playground extensions.
but I would like to follow your suggestion.

sapegin: I didn't say you've copied but essentially it does the same.

mitsuruog:
@sapegin I updated my test code. please review me again.

sapegin:
The idea is _not to have_ a snapshot. Please have a look at my suggestion in my previous comment.

mitsuruog:
I was misunderstood what you said. but now it is clear. thank you.
Previous tests compared objects with each other. That didn't tell exactly which point was testing.
but now I can see that your suggestion clearly checks only the `code` part.

@sapegin thank you for many times review.

sapegin:
Merged manually to 11 branch: 5975c7cc722bf057231ad14115e47600a1fd98b2

PR #1544, opened 2020-03-01
fix: specify webpack as a peerDependency

KraigWalker:
* Add `webpack` to `peerDependencies`

The [Getting Started Documentation](https://react-styleguidist.js.org/docs/getting-started.html) explains that the User should install webpack separately as a `devDependency` in order to successfully run `react-styleguidist`.

The actual `package.json` for styleguidist doesn't actually list `webpack` as a peer dependency. It is, however a dev dependency (after all, we build styleguidist with webpack)

This is fine. However, when installing styleguidist as a User, some webpack plugins that styleguidist uses complain that styleguidist isn't providing `webpack` from anywhere.

Screenshot 2020-03-01 at 15 55 45

This is obviously because we've quite rightly delegated `webpack` installation to the user, and therefore we should list it as a `peerDependency` so that the plugins can look further up the dependency tree to resolve the `webpack dependency`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5157,9 +5157,9 @@
       "integrity": "sha512-qerSRB0p+UDEssxTtm6EDKcE7W4OaoisfIMl4CngyEhjpYglocpNg6UEqCvemdGhosAsg4sO2dXJOdyBifPGCg=="
     },
     "events": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/events/-/events-3.0.0.tgz",
-      "integrity": "sha512-Dc381HFWJzEOhQ+d8pkNon++bk9h6cdAoAj4iE6Q4y6xgTzySWXlKn05/TVNpjnfRqi/X0EpJEJohPjNI3zpVA==",
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/events/-/events-3.1.0.tgz",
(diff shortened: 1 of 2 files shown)


sapegin:
Webpack isn't in peer dependencies because we support tools like Create React App, that have their own webpack, so it won't be in user's `package.json`.

jasonwilliams:
@KraigWalker is correct, `webpack` needs to be specified as a peerDependency in this package. I've also been getting the same errors.

```
➤ YN0002: │ react-styleguidist@npm:11.1.5 [42b1e] doesn't provide webpack (pcb22d), requested by clean-webpack-plugin
➤ YN0002: │ react-styleguidist@npm:11.1.5 [42b1e] doesn't provide webpack (p1bd3b), requested by copy-webpack-plugin
➤ YN0002: │ react-styleguidist@npm:11.1.5 [42b1e] doesn't provide webpack (p73280), requested by mini-html-webpack-plugin
➤ YN0002: │ react-styleguidist@npm:11.1.5 [42b1e] doesn't provide webpack (pd3d03), requested by terser-webpack-plugin
➤ YN0002: │ react-styleguidist@npm:11.1.5 [42b1e] doesn't provide webpack (pe6867), requested by webpack-dev-server
➤ YN0060: │ web@workspace:web provides webpack (p46935) with version 5.21.2, which doesn't satisfy what pdfjs-dist and some of its descendants request
```

Even if you support create-react-app (which provides webpack) you still need to specify it as a peer, because you're using it without directly installing it (somebody else is, regardless of whether its the user or CRA). NPM will still choose CRA's version of webpack (provided you're compatible with it)

The second issue is that react-styleguidist is not signalling itself as compatible with React v17 (which should be an easy fix)

jasonwilliams:
I've created a new PR https://github.com/styleguidist/react-styleguidist/pull/1744

PR #1549, opened 2020-03-07
Chore: Remove remnants of Flow

sapegin:
Remove pieces of Flow left after the TypeScript migration.


elevatebart:
I believe we have to keep `@babel/preset-flow` for some tests.
Apart from that and the small alert, I see no pb with this one.

sapegin:
We'll need to updates the docs then: https://react-styleguidist.js.org/docs/api.html

elevatebart:
I agree, I will take care of it before tomorrow.

Bart Ledoux

Join me on LinkedIn
Or as @elevatebart on Discord, Twitter and Github

elevatebart:
I updated the documentation using imports instead of requires. I hope you find it useful.
It might be useful to add a note about NodeJs not accepting es6 imports. There is a non obvious need for transpilation.

sapegin:
I think we can also reexport it as CommonJS in a separate module to avoid breaking changes.

elevatebart:
I did my best. I hope you like it.

sapegin:
I'm not sure the `module` field is useful here, I think it's only used by bundlers, but good enough for now ;-) Thanks for the changes!

PR #1553, opened 2020-03-12
[WIP] iframe proof of concept

sapegin:
@okonet @elevatebart hey folks, I'm experimenting with rendering example in iframes. So far separate iframes works fine, but inside the style guide it's incredibly slow. I have a feeling it's not because of iframes but because of webpack-dev-server, but don't know more yet.

Let me know if you have any ideas how to make it faster. It doesn't make sense to continue, unless I solve the performance issue.

The UI loads pretty fast but then, there are bunch of websocket requests that are pending for a very long time.

image

Another issue for now is that when you change the code, all iframes are refreshing, not just for the component you've changed.


sapegin:
I'll try to merge webpack configs again and use https://github.com/ryanseddon/react-frame-component to render iframes.

Though having a separate page with just a component without any UI is a good thing too.

okonet:
> Though having a separate page with just a component without any UI is a good thing too.

I think it's not as valuable in dev mode as for automations etc. I can imagine the URL of this iframe can be retrieved for automation purposes anyways so I would not block this branch with such requirement.

src/bin/styleguidist.js
```diff
@@ -141,7 +142,7 @@ function commandServer() {
 				openBrowser(urls.localUrlForBrowser);
 			}
 		}
-	}).compiler;
+	}).compiler.compilers[1];
```
okonet: hmm?

src/client/iframe.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable import/first */
+import './polyfills';
+import ReactDOM from 'react-dom';
+import renderIframe, { ExamplesMap } from './utils/renderIframe';
+
+// eslint-disable-next-line compat/compat
+const url = new URL(location.href);
+const file = url.searchParams.get('file') || undefined;
```
okonet: what is this trying to do ` || undefined`? I would either make it explicit with `null` or remove since it will be undefined if not set by the first expression.

src/client/iframe.ts
```diff
@@ -0,0 +1,28 @@
...
+const exampleIndexRaw = url.searchParams.get('exampleIndex');
+const exampleIndex = typeof exampleIndexRaw === 'string' ? parseInt(exampleIndexRaw) : undefined;
```
okonet: Same here, let's use `null` to make it explicit.

okonet:
Do you think we could make the rendering part pluggable as well so you could choose between single document and iframe rendering? I'm wondering if supporting both could be a best of both worlds even. iframe in isolation mode and document for the styleguide itself.

elevatebart:
Another source of inspiration: https://github.com/seek-oss/playroom

src/bin/styleguidist.js
sapegin: It's too early for nitpicking, most of this code will be rewritten because it doesn't solve the problem yet.

okonet:
> Nope, support for inline components is insanely complex and has many issues with isolation. Adding another complex solution (iframes) without removing something is too much complexity.

Okay I agree. Also things like media queries aren’t even possible right now afaik (maybe with a super dirty trick?) so I think iframes FTW.

Maybe @bebraw can help with the webpack part?

bebraw:
@okonet I might know someone that has a clue (he wrote wds of his own for webpack). I'll ping to get better idea. Hopefully we can find a solution.

sapegin:
Now this actually seems to work! I'm not making a network request for each iframe, but render a React app inside an iframe.

sapegin:
No global styles yet ;-)

image

sapegin:
But now there's another problem with webpack: css-loader adds all styles to the host document, and I need them in each iframe instead.

bebraw:
> But now there's another problem with webpack: css-loader adds all styles to the host document, and I need them in each iframe instead.

In one case, I solved a global style issue with [react-shadow](https://www.npmjs.com/package/react-shadow) though I am not sure if that would help here at all since we're going with iframes for isolation.

Another direction to check is to figure out how css-loader would have to change for it to work in this use case and potentially fork it.

okonet:
I agree. We could check style-loader (since that’s the one that is injecting styles I to document) and either add an option or just write our own (could be based on JSS injection since it’s already in the bundle).

bebraw:
> I agree. We could check style-loader (since that’s the one that is injecting styles I to document) and either add an option or just write our own (could be based on JSS injection since it’s already in the bundle).

Ah, yeah. Of course. **style-loader** exposes what they call an `insert` method. In the docs they mention iframes specifically, https://www.npmjs.com/package/style-loader#insert .

sapegin:
Yup, it's style-loader, not css-loader. And it already has an option to redirect the injection to an iframe. So we can create an iframe for styles, inject styles to this iframe, and then clone them to each example iframe. There's also an option to inject a link instead of `style` tags. This should make it easier: we would need to only inject styles once on example iframe creation I think (not sure how hot reload works in this case).

The main difference with their example is that we have multiple iframes.

elevatebart:
It might be a simple matter of webworkers implementations. It fails to load on firefox for now.

elevatebart:
Should we render the components inside a div instead of reiectly on the body? 
I believe it would avoid the weird warning:
`Rendering components directly into document.body is discouraged`

sapegin:
I was able to render styles inside iframes with a custom style-loader config and cloning the style tag. Not sure how to do hot reload here: style-loader doesn't trigger the `insert` method again, so I guess using `MutationObserver` is the only way.

okonet:
I’d keen trying out if you can point me to the code

sapegin:
@okonet 

1. The style-loader config is here: https://github.com/styleguidist/react-styleguidist/pull/1553/files#diff-f50ed279a2aff4fa7dc04aa11f530b96 — we'll need to patch the user's config later.

2. The iframe rendering is here: https://github.com/styleguidist/react-styleguidist/pull/1553/files#diff-51a5b483b47eb9188e34acc3d7fd1171 — I already use `MutationObserver` here to resize the iframe, we can try something similar to track style updates.

sapegin:
I have CSS hot reload with custom style loader! It's a bit messy but looks like it works ;-)

elevatebart:
Plus now it works on FireFox !!!

How did you do that?

sapegin:
I guess not rendering to `body` is what fixed Firefox.

okonet:
😭

sapegin:
As usual — lack of feedback ;-( But let's keep it open in case anyone wants to continue working on this.

Issue #1551, opened 2020-03-11
Property types don't show up


smashercosmo:
After upgrading to latest styleguidist (the one with native TS support) property types stopped showing up

![Screenshot 2020-03-11 at 14 06 17](https://user-images.githubusercontent.com/273283/76415867-ac03bf80-63a2-11ea-91dd-2cea45813f25.png)

**Current behavior**

With react-docgen-typescript you could see property types

**To reproduce**

Try to use react-styleguidist with TS

**Expected behavior**

You should be able to see property types

mitsuruog:
I have the same issue. version is`v11.0.1`.

![react-styleguidist issues 1551 - CodeSandbox](https://user-images.githubusercontent.com/1703219/76431298-b9499a00-63f4-11ea-95dd-1f4650fb2bdb.png)

Here is my sandbox.

- https://codesandbox.io/s/react-styleguidistissues1551-1iyof

sapegin:
Feel free to send a pull request with a fix 👾

mitsuruog:
In my case, there are 2 issues.

1. `react-docgen` does not parse the type of the component correctly.

The type of this component will not parse anything.

```tsx
export const Button: React.FC<ButtonProps> = props => {
  // ...
}
```

But this component can parse.

```tsx
export function Button(props: ButtonProps) {
  // ...
}
```

It might be an issue or limitation for `react-docgen`.

2. `react-styleguidist` does not render `tsType` even if `react-docgen` has parsed successfully.

`react-docgen` returns `tsType` including the type info, but the current version does not seem to handle this correctly.

mitsuruog:
@sapegin I am working on it. 😉

mitsuruog:
@sapegin I want to hear your opinion.

I am wondering how to display types in the UI for TypeScript.
It is a little different in the result of the UI from the way what the old version was. 
because for the TypeScript users, it used to use `react-docgen-typescript`, but now it uses `react-docgen`. 

ex)

If type is something like:

```
prop3: () => void;
```
 Old version:

```
Prop name: prop3
Type: () => void;
```
New version:

```
Prop name: prop3
Type: function
// NOTE: `() => void` goes to tooltip
```

![stylesheet-example-column](https://user-images.githubusercontent.com/1703219/76680739-81c63200-662e-11ea-8598-42d2d2133fe3.png)

[...]

PR #1563, opened 2020-03-16
Fix Property types don't show up

mitsuruog:
Resolve

- Property types don't show up #1551

I organize react-docgen prop type(Named `PropDescriptor`) to have `flowType` and `tsType` both and I remove "WithFlow" word from the types.

review me please.

### DIFF
--- a/src/client/rsg-components/Props/Props.spec.tsx
+++ b/src/client/rsg-components/Props/Props.spec.tsx
@@ -84,6 +84,37 @@ function renderFlow(propsType: string[], defaultProps: string[] = []) {
 	return render(<ColumnsRenderer props={propsToArray(props.props)} />);
 }
 
+function renderTypeScript(
+	propsType: string[],
+	defaultProps: string[] = [],
+	preparations: string[] = []
+) {
+	const props = parse(
+		`
+		// @TypeScript
+		import * as React from 'react';
+		${preparations.join(';')}
+		type Props = {
+			${propsType.join(';')}
+		};
+		export default class Cmpnt extends React.Component<Props> {
+			static defaultProps = {
+				${defaultProps.join(',')}
+			}
+			render() {
+			}
+		}
+	`,
+		undefined,
+		undefined,
+		{ filename: 'Component.tsx' }
+	);
+	if (Array.isArray(props)) {
+		return render(<div />);
+	}
+	return render(<ColumnsRenderer props={propsToArray(props.props)} />);
+}
+
 describe('PropsRenderer', () => {
 	test('should render a table', async () => {
 		const { findAllByRole } = render(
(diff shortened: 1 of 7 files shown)


src/client/rsg-components/Props/Props.spec.tsx
```diff
@@ -509,94 +540,208 @@ describe('props columns', () => {
...
-	test('should render custom class type', () => {
-		const { container } = renderFlow(['foo: React.ReactNode']);
-
-		expect(getText(container)).toMatchInlineSnapshot(`
-		"Prop name: foo 
-		Type: React.ReactNode 
-		Default: Required 
-		Description:"
-	`);
+	describe('flowType', () => {
```
mitsuruog: I moved the tests for the `flowType` by wrapping `describe`

src/client/rsg-components/Props/Props.spec.tsx
```diff
@@ -84,6 +84,37 @@ function renderFlow(propsType: string[], defaultProps: string[] = []) {
...
 }
 
+function renderTypeScript(
+	propsType: string[],
+	defaultProps: string[] = [],
+	preparations: string[] = []
+) {
+	const props = parse(
+		`
+		// @TypeScript
```
sapegin: This comment doesn't mean anything in TypeScript and will confuse readers.

src/client/rsg-components/Props/Props.spec.tsx
```diff
@@ -509,94 +540,219 @@ describe('props columns', () => {
...
+			expect(getText(container)).toMatchInlineSnapshot(`
+			"Prop name: foo 
+			Type: React.ReactNode 
+			Default: Required 
+			Description:"
+		`);
+		});
+	});
+
+	describe('TypeScript', () => {
```
sapegin: Is there any difference between Flow and TypeScript test cases?

src/client/rsg-components/Props/util.ts
```diff
@@ -7,8 +7,9 @@ export function unquote(string?: string): string | undefined {
 	return string && string.replace(/^['"]|['"]$/g, '');
 }
 
-export interface PropDescriptorWithFlow extends PropDescriptor {
-	flowType?: FlowTypeDescriptor;
+export interface PropDescriptor extends BasePropDescriptor {
```
sapegin: If `flowType` and `tsType` have the same shape, can we merge them? This will remove lots of conditions.

src/client/rsg-components/Props/Props.spec.tsx
mitsuruog: Yes, there is some different output. and I also added for the unknown type for the TypeScript.

sapegin: But the input is the same for all but unknown? Can we use `describe.each` here to merge them?

src/client/rsg-components/Props/util.ts
mitsuruog: hmm. here is my 2 cents though.
I would like to keep the current structure for now.
because of this structure is for react-docgen, this structure is out of the control from our project. now it may have the same structure, but I am not sure if it continues.
Ideally, I hope react-docgen has the type definition and we can use it.

src/client/rsg-components/Props/Props.spec.tsx
mitsuruog: Cool. I didn't have this idea. Let me try once.

src/client/rsg-components/Props/util.ts
sapegin: Okay, let's keep it for now then.

src/client/rsg-components/Props/Props.spec.tsx
mitsuruog: ah sorry, the input has a bit different.
I made to be able to inject preparatory code into test components for TypeScript.
I probably can merge them, do you still want me to?

sapegin: Could you show an example? I believe we want to test very similar things, so merging makes sense. But maybe the difference is too big? ;-)

mitsuruog: https://github.com/styleguidist/react-styleguidist/pull/1563/files#diff-ea72f02dad8b44122fd2deb5537e1b1bR703

Only here is different. the difference is not too big, so I probably can merge them.

src/client/rsg-components/Props/Props.spec.tsx
```diff
@@ -509,94 +540,107 @@ describe('props columns', () => {
...
-		"Prop name: foo 
-		Type: React.ReactNode 
-		Default: Required 
-		Description:"
-	`);
+	describe.each([
+		[
+			'flowType',
+			renderFlow,
+			{ enum: { declaration: "type MyEnum = 'One' | 'Two'", expect: { type: 'enum' } } },
```
mitsuruog: I use the same `renderFlowType` (Now the name has changed to `renderAdvancedType`) for each.
Regarding rendering result, only different is the part of "enum". other types have the same result.

mitsuruog:
@sapegin please review me again.

Issue #1559, opened 2020-03-14
Sections with href are always opening in new tabs


iSysProc:
![image](https://user-images.githubusercontent.com/17797341/76674313-ba97e400-65f9-11ea-8fa8-c3ecc4cc3071.png)

Section field 'external' don't affect on `shouldOpenInNewTab`

Maybe change this row on: `shouldOpenInNewTab: section.external && !!section.href,` ?

sapegin:
Feel free to send a pull request with a fix 👾

mitsuruog:
What is the correct and intended behavior?

sapegin:
`external` should define the behavior if present. We probably can remove the magic that opens all sections with `href` in the new tab by default. At least the docs and the examples would be correct in this case:

https://react-styleguidist.js.org/docs/components.html

PR #1564, opened 2020-03-17
Fix Sections with href are always opening in new tabs

mitsuruog:
Resolved.

- Sections with href are always opening in new tabs #1559

The section does not open in the new tab unless `href` and `external` has set both.

please review me.

### DIFF
--- a/src/client/rsg-components/TableOfContents/TableOfContents.tsx
+++ b/src/client/rsg-components/TableOfContents/TableOfContents.tsx
@@ -76,7 +76,7 @@ export default class TableOfContents extends Component<TableOfContentsProps> {
 				heading: !!section.name && children.length > 0,
 				content,
 				selected,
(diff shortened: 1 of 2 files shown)


src/client/rsg-components/TableOfContents/TableOfContents.tsx
```diff
@@ -76,7 +76,7 @@ export default class TableOfContents extends Component<TableOfContentsProps> {
 				heading: !!section.name && children.length > 0,
 				content,
 				selected,
-				shouldOpenInNewTab: !!section.href,
+				shouldOpenInNewTab: !!section.external && !!section.href,
```
mitsuruog: `external ` is an optional field. 
`undefined && true` gives `undefined`, so I make force boolean cast here.

sapegin:
Thanks!

PR #1565, opened 2020-03-19
Fix usage of CleanWebpackPlugin

Lazyuki:
Was still using an option that no longer exists. Fixed to mimic the original behavior. The option `root` was [removed in v2](https://github.com/johnagan/clean-webpack-plugin/issues/106) but [this commit](https://github.com/styleguidist/react-styleguidist/commit/98042591dc75cad8cdf88cc2a2f0817d69e5e66d) forgot to address it.

It's deleting everything in the `styleguidistDir` folder which isn't the way it was before.
With this PR, it only deletes files in `build/`.

### DIFF
--- a/src/scripts/make-webpack-config.ts
+++ b/src/scripts/make-webpack-config.ts
@@ -98,7 +98,7 @@ export default function(
 			},
 			plugins: [
 				new CleanWebpackPlugin({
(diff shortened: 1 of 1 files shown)


sapegin:
Nice, thanks!

Issue #1566, opened 2020-03-23
Crash with some rewritten imports 


dustinsoftware:
**Current behavior**

Some imports are not rewritten correctly and produce naming collisions.

![image](https://user-images.githubusercontent.com/942358/77339140-bce8f500-6d01-11ea-9327-7114f33c5778.png)

Looks like the logic in this file does not handle unique identifiers for all cases: https://github.com/styleguidist/react-styleguidist/blob/3469cbba172d9f81657b388dbffbb8fd5dd7fba4/src/client/utils/rewriteImports.ts

These imports:

```js
import Demo from './demo1/component.js';
import Demo2 from './demo2/component.js';

<Button><Demo /><Demo2 /></Button>
```

Get transformed into:
```js
const component_js$0 = require('./demo1/component.js');
const Demo = component_js$0.default || component_js$0;
const component_js$0 = require('./demo2/component.js');
const Demo2 = component_js$0.default || component_js$0;

React.createElement( Button, null, React.createElement( Demo, null ), React.createElement( Demo2, null ) )
```

**To reproduce**

Demo here: https://github.com/dustinsoftware/styleguidist-example/

**Expected behavior**

Unique identifier names are used.

sapegin:
Feel free to send a pull request with a fix.

This will also go away after #1553.

PR #1567, opened 2020-03-25
Fix Crash with some rewritten imports

mitsuruog:
Resolve:

- Crash with some rewritten imports #1566

After the change, the relative path will split by `/` then joins by `_`.
For example, if gives `./cat/capybara/hamster`, it will be `dot_cat_capybara_hamster`.

Please review me.


src/client/utils/rewriteImports.ts
```diff
@@ -16,10 +16,10 @@ function alias(key: string): {key: string; name: string} {
 let num: number;
 function generate(keys: string[], dep: string, base: string, fn: string): string {
 	const tmp =
-		(dep
+		dep
```
sapegin: I think it can be done simpler and even more bulletrproof:

```suggestion
		dep.replace(/\W/g, '_')
```

You don't need split, map and join here.

mitsuruog: Excellent...

mitsuruog:
Thank you @sapegin 
my code is the way simpler and cleaner than before. 
please review me again.

PR #1571, opened 2020-03-28
Fix missing FlowType enum values in prop description

TomasBarry:
In `ef4c109b`, the file `PropsRenderer.js` (located at
`src/client/rsg-components/Props/PropsRenderer.js`) was removed. In
`PropsRenderer.js`, the `renderExtra` method checked whether `getType`
for the argument to `renderExtra` was present:

```es6
function renderExtra(prop) {
  const type = getType(prop);
  if (!type) {
    return null;
  }
  ...
}
```

However, in `ef4c109b`, this method was replaced with `renderExtra.tsx`
and the condition was changed to:

```typescript
export default function renderExtra(prop: PropDescriptorWithFlow): React.ReactNode {
  const type = getType(prop);
  if (!prop.type || !type) {
    return null;
  }
```

Unfortunately, this extra condition has resulted in this method always
returning `null` for a Flow typed prop as `prop.type` is always `null`
as `prop.type` is never set.

This commit reverts the condition to what it was before the migration to
TypeScript.

### DIFF
--- a/src/client/rsg-components/Props/Props.spec.tsx
+++ b/src/client/rsg-components/Props/Props.spec.tsx
@@ -605,13 +605,22 @@ describe('props columns', () => {
 
 		test('should render enum type', () => {
 			const { container } = renderFn(['foo: MyEnum'], [], [options.enum.declaration]);
-
-			expect(getText(container)).toMatchInlineSnapshot(`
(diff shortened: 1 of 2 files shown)


sapegin:
Thanks!

PR #1573, opened 2020-04-06
fix zero example

b2whats:
Example with targetIndex = 0 was not shown

### DIFF
--- a/src/client/utils/getRouteData.ts
+++ b/src/client/utils/getRouteData.ts
@@ -110,9 +110,10 @@ export default function getRouteData(
 				sections = [
 					{
 						...filteredSections[0],
-						components: filteredComponents
-							? [filterComponentExamples(filteredComponents[0], targetIndex as number)]
-							: [],
+						components:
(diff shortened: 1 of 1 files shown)


b2whats:
@sapegin

src/client/utils/getRouteData.ts
```diff
@@ -110,10 +110,9 @@ export default function getRouteData(
 				sections = [
 					{
 						...filteredSections[0],
-						components:
-							filteredComponents && targetIndex
-								? [filterComponentExamples(filteredComponents[0], targetIndex)]
-								: [],
+						components: filteredComponents
+							? [filterComponentExamples(filteredComponents[0], targetIndex as number)]
+							: [],
```
sapegin, line 115: I think something like this would be better:

```suggestion
						components: filteredComponents && typeof targetIndex === number
							? [filterComponentExamples(filteredComponents[0], targetIndex)]
							: [],
```

Avoid `as` because it basically disables type check.

b2whats: Done

Issue #289, opened 2017-01-19
print out values of `PropTypes.oneOf`


chrisdrackett:
This is currently done with shape and its really nice. Right now if you use `oneOf` the only thing shown in type is `enum`, it would be great to be able to see the supported values.

sapegin:
Yeah, and I believe it’s not very difficult to make. It would be very cool if you have time and will to try it and send a pull request ;-)

chrisdrackett:
yeah, I have that on my list, figured I would post here in case someone gets to it before me :)

AoDev:
Hello, I was checking if I could help.

I found this issue and I am not sure what you mean. I already have the values in my styleguide, in the description column. (One of 'default', 'danger')

It looks like there is nothing to do here?

<img width="613" alt="screen shot 2017-01-31 at 21 13 21" src="https://cloud.githubusercontent.com/assets/1526150/22482659/3bf115b6-e7fa-11e6-8092-fbc8311785f8.png">

n1313:
@chrisdrackett the values of `oneOf` are already being printed, so not sure what issue you are describing here. Are you talking about `oneOfType`?

sapegin:
`objectOf` is supported in 5.x beta. Closing this for now. Feel free to reopen with more details.

PR #1575, opened 2020-04-10
Add function signature to description

malakada:
I didn't see any pre-existing format to use for PR's, so please let me know if there is one so I can update this text.

I also need help figuring out how the heck to get the test to work. The test file is quite large, and I had a hard time figuring out where my test should go or what helpers were available for it. Sadly, I ran out of timeboxed time for myself to figure this out on my own.

## Goal state

While using Styleguidist at work, I found that it would be helpful to have the function signature (ie: `funcName(): void` or so) display in the description for the Props & Methods table.

Screen Shot 2020-04-10 at 12 54 30 PM

#### How do I do this without this PR?

Add this to your config file:

```
  propsParser(filePath, source, resolver, handlers) {
    const docGen = require('react-docgen').parse(source, resolver, handlers);
    Object.keys(docGen[0].props).map((propName) => {
      const props = docGen[0].props[propName];
      if (!props || !props.flowType) { return; }
      if (props.flowType.name === "signature" && props.flowType.type === "function") {
        props.description += "<div>Function signature: `" + props.flowType.raw + "`</div>";
      }
    });
    return docGen;
  },
```

🤷 

## Current state

Screen Shot 2020-04-10 at 12 41 56 PM

### DIFF
--- a/src/client/rsg-components/Props/Props.spec.tsx
+++ b/src/client/rsg-components/Props/Props.spec.tsx
@@ -427,6 +427,17 @@ describe('props columns', () => {
 	`);
 	});
 
+	test('should render function signature in description', () => {
+		const { container } = renderFlow(['onSomething: (e) => console.log(e)']);
+
+		expect(getText(container)).toMatchInlineSnapshot(`
+    "Prop name: onSomething
+    Type: func
+    Description:
+     Function signature: onSomething(): void"
+    `);
+	});
+
 	test('should render arguments from JsDoc tags', () => {
 		const props: any = [
 			{
--- a/src/client/rsg-components/Props/renderExtra.tsx
+++ b/src/client/rsg-components/Props/renderExtra.tsx
@@ -59,6 +59,14 @@ function renderShape(props: Record<string, PropDescriptor>) {
 	});
 }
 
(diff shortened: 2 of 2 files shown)


sapegin:
If you hover on the `function` type, you'll see the signature. I'm not sure what exactly you're trying to fix with this pull request.

malakada:
> If you hover on the `function` type, you'll see the signature. I'm not sure what exactly you're trying to fix with this pull request.

Basically, I'm trying to enhance the UX to make it easier to understand what kind of functions a component is expecting.

It's the same thing that was fixed with the `enum` type which previously required a hover to expose what values were accepted. I saw that there was a PR that was accepted for listing out the `enum` values in the description (but I can't find it now, only this: https://github.com/styleguidist/react-styleguidist/issues/289 which isn't quite what I am _pretty sure_ I saw while researching this.) which led me to believe that a PR would be welcome.

Sorry for not explaining the use case/motivation better.

sapegin:
Having it in two places is much worse for the UX, and if a similar change was approved, it was a mistake.

By looking at the screenshot I don't think it's an improvement. Improving tooltips would be much better solution here.

Issue #1594, opened 2020-05-16
Sections with numbers in the name don't work


sapegin:
**Current behavior**

The URL changes to `http://localhost:6061/#/4.%20Design%20Tokens` but no navigation happens. There's an error in the browser console:

```
Uncaught TypeError: Cannot read property 'length' of undefined
    at getRouteData (getRouteData.js:119)
    at renderStyleguide (renderStyleguide.js:46)
    at render (index.js:46)
    at eval (index.js:60)
    at Module../node_modules/react-styleguidist/lib/client/index.js (main.bundle.js:3855)
    at __webpack_require__ (main.bundle.js:726)
    at fn (main.bundle.js:101)
    at eval (client:4)
    at Object.0 (main.bundle.js:13010)
    at __webpack_require__ (main.bundle.js:726)
```

**To reproduce**

1. Create a section with a number in the name (for example, `4. Design Tokens`).
2. Run Styleguidist
3. Try to navigate to this section

**Expected behavior**

The section is open.

mitsuruog:
@sapegin 
which version are you using?
I tried to reproduce it by the latest version `11.0.6`. but the URL format of the sections seems been changed, and I can not reproduce it. the navigation works for me.

**styleguide.config.js**

```
module.exports = {
  ...
  sections: [{
    name: '1.Documentation',
    content: 'src/docs/Document.md',
  }, {
    name: '2. Des ign',
    content: 'src/docs/Design.md',
  }, {
    name: "Components",
    components: "src/components/**/*.{js,jsx}",
  }]
};
```

**Result:**
<img width="890" alt="スクリーンショット 2020-05-17 11 02 17" src="https://user-images.githubusercontent.com/1703219/82134043-f3d51600-982d-11ea-80ba-72edc3164088.png">


**The URL for the section:**

`http://localhost:6060/#section-2-des-ign`

or am I missing something?

sapegin:
Interesting: I had some 10.x version. Upgraded to 11.0.6. The error is gone but the navigation doesn't work: the URL changes but not the page contents ;-/ I also have `pagePerSection: true`. Looks like some regexp ignores numbers.

PR #1595, opened 2020-05-17
Fix Sections with numbers in the name don't work

mitsuruog:
Resolve:

- Sections with numbers in the name don't work #1594

The main reason is that `parseInt` parses the first number string as a number.
Therefore, when `1.Doc` given, it was computed as `1`.

### DIFF
--- a/src/client/utils/__tests__/getInfoFromHash.spec.ts
+++ b/src/client/utils/__tests__/getInfoFromHash.spec.ts
@@ -35,4 +35,24 @@ describe('getInfoFromHash', () => {
 			targetIndex: undefined,
 		});
 	});
+
+	it('should return a proper parsed result even though the hash starts with a number', () => {
+		const result = getInfoFromHash('#/1.Documentation');
+		expect(result).toEqual({
+			isolate: false,
+			hashArray: ['1.Documentation'],
+			targetName: '1.Documentation',
+			targetIndex: undefined,
+		});
+	});
+
+	it('should return a proper parsed result even though the hash contains a space', () => {
+		const result = getInfoFromHash('#/1. Just do it');
+		expect(result).toEqual({
+			isolate: false,
+			hashArray: ['1. Just do it'],
+			targetName: '1. Just do it',
+			targetIndex: undefined,
+		});
+	});
 });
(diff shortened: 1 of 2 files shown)


src/client/utils/__tests__/getInfoFromHash.spec.ts
```diff
@@ -35,4 +35,24 @@ describe('getInfoFromHash', () => {
 			targetIndex: undefined,
 		});
 	});
+
+	it('should return a proper parsed result even though the hash starts with a number', () => {
```
mitsuruog, line 39: I'm a little concerned about if test cases are enough.
let me know if you know some URL that it should test it. I'll add it.

src/client/utils/getInfoFromHash.ts
```diff
@@ -1,8 +1,7 @@
-import isNaN from 'lodash/isNaN';
 import { hasInHash, getHashAsArray } from './handleHash';
 
-function filterNumbers(item: string): boolean {
-	return isNaN(parseInt(item, 10)) && item !== '';
+function consistsOnlyNumbers(item: string): boolean {
```
sapegin, line 3: Something like this would be more correct and clear:

```suggestion
function hasDigitsOnly(item: string): boolean {
```

src/client/utils/__tests__/getInfoFromHash.spec.ts
sapegin: I don't really see the difference between these two ;-/ Do we have anything with a number in `targetIndex`?

src/client/utils/getInfoFromHash.ts
mitsuruog: Thank you. Since I am not a native speaker so I am always welcoming this kind of suggestion.

src/client/utils/__tests__/getInfoFromHash.spec.ts
mitsuruog: You're right. At first I thought it was worth it, but now it doesn't seem like it's worth much.
I've added a case that extracts the target index.

mitsuruog:
I've updated the PR. please review me again.

sapegin:
Thanks, merging!

PR #1609, opened 2020-06-08
Support Promise API for webpack config

gzaripov:
Webpack supports promise in configuration:
https://webpack.js.org/configuration/configuration-types/#exporting-a-promise
Added Promise support to styleguidist. 
PR has breaking changes.


sapegin:
What problem you're trying to solve with this pull request?

gzaripov:
@sapegin we return promise in our webpack, we need it because we get tokens for development using developer ssh key. We have already solved it by making sync version of webpack config.

PR #1610, opened 2020-06-08
Treat exact props the same as shape

aleshh:
Let's treat `PropTypes.exact` the same way we treat `PropTypes.shape`. That is, show their constituent properties in Styleguidist. 

(Noob open-source contributor, sorry if I missed something. I tested this with the basic example, but didn't add any tests for it in code.)


sapegin:
Thanks for the pull request! Please add some tests.

sapegin:
Merging! And please write a comment next time when you're doing with the changes — there's a big chance I'll never see them otherwise.

PR #1623, opened 2020-07-03
Fix param of array

isaacbuckman:
Doctrine has a function to stringify the parameter types. If the type is a `string[]`, doctrine will stringify this into `Array.<string>`. I believe this is backwards compatible.

### DIFF
--- a/examples/sections/styleguide.config.js
+++ b/examples/sections/styleguide.config.js
@@ -93,8 +93,8 @@ module.exports = {
 			env === 'development'
 				? false
 				: {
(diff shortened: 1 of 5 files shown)


isaacbuckman:
@sapegin Any idea on why the doctrine import does not seem to be working? It is working locally.

sapegin:
It says `'Unknown type undefined'` — likely we need to add some setup in the tests that are failing.

sapegin:
Thanks, merging!

PR #1627, opened 2020-07-12
fix: closing isolated vue

### LINKED ISSUE #1479: Closing isolated view doesn't return to correct location when using sections
**Current behavior**

When I view a component in isolated view then close it returns to root not to the component (when using sections)

**To reproduce**

View isolated component then return

**Expected behavior**

Return to component I was viewing. eg. #/Forms?id=form but returns me to #/form

elevatebart:
fixes #1479

**Remark** For technical reasons, I had to remove the return from isolate button here: I could not access `href` without going through a context or a prop train. 

It looked logical to me: If there was only one action it should not be in two places. 
With this change, it would be only on the top right. Tell me if you need a fix for it.

Screen Shot 2020-07-11 at 11 20 36 PM


src/client/rsg-components/TableOfContents/__snapshots__/TableOfContents.spec.tsx.snap
```diff
@@ -26,49 +26,46 @@ exports[`should filter list when search field contains a query 1`] = `
   searchTerm=""
```
sapegin, line 26: Let's get rid of these snashots — they aren't really testing anything ;-/ We used to have way too many but there's no reason to keep them ;-)

https://blog.sapegin.me/all/snapshot-tests/

src/client/utils/processComponents.ts
```diff
@@ -22,6 +32,15 @@ export default function processComponents(components: Rsg.Component[]): Rsg.Comp
...
 						examples: [...(component.props.examples || []), ...(component.props.example || [])],
 					},
+					href:
+						component.href ||
+						getUrl({
+							name: component.props.displayName,
+							slug: component.slug,
+							anchor: !useRouterLinks,
+							hashPath: useRouterLinks ? hashPath : false,
+							id: useRouterLinks ? useHashId : false,
```
sapegin, line 42: I may be missing something but why are we assigning a boolean to an ID here?

sapegin:
>  I had to remove the return from isolate button here

I think it's fine ;-)

src/client/utils/processComponents.ts
elevatebart: I believe the name of this param could be a lot more explicit. 
This options says: `urlContainsAndId`

https://github.com/styleguidist/react-styleguidist/blob/921e8b1ea8ba0c62a2b7400facdcee06cde07fd2/src/client/utils/getUrl.ts#L122-L133

elevatebart: maybe the name `useSlug` would be more appropriate.

sapegin: Yeah, the API of this function is unfortunate ;-) `useSlug` is good name.

elevatebart:
Hello @sapegin, I hope you are well.
Should I merge this one?

sapegin:
The code is good now but I still see changes in a gigantic snapshot that I'd love to remove entirely ;-)

elevatebart:
> The code is good now but I still see changes in a gigantic snapshot that I'd love to remove entirely ;-)

Snapshots replaced with real RTL tests. I hope you like them.

src/client/rsg-components/TableOfContents/TableOfContents.spec.tsx
```diff
@@ -1,121 +1,95 @@
...
-		</TableOfContentsRenderer>
+	const { getByPlaceholderText, getAllByTestId, getByTestId } = render(
+		<TableOfContents sections={sections} />
 	);
-
-	expect(actual).toMatchSnapshot();
+	expect(getAllByTestId('rsg-toc-link').length).toBe(6);
+	fireEvent.change(getByPlaceholderText('Filter by name'), { target: { value: searchTerm } });
+	expect(getAllByTestId('rsg-toc-link').length).toBe(1);
+	expect(getByTestId('rsg-toc-link').innerHTML).toBe('Forms');
```
sapegin, line 92:
```suggestion
	expect(getByTestId('rsg-toc-link')).toHaveTextContent('Forms');
```

src/client/rsg-components/TableOfContents/TableOfContents.spec.tsx
```diff
@@ -1,121 +1,95 @@
...
-			<div>foo</div>
-		</TableOfContentsRenderer>
+	const { getByPlaceholderText, getAllByTestId, getByTestId } = render(
+		<TableOfContents sections={sections} />
 	);
-
-	expect(actual).toMatchSnapshot();
+	expect(getAllByTestId('rsg-toc-link').length).toBe(6);
+	fireEvent.change(getByPlaceholderText('Filter by name'), { target: { value: searchTerm } });
+	expect(getAllByTestId('rsg-toc-link').length).toBe(1);
```
sapegin, line 91:
```suggestion
	expect(getAllByTestId('rsg-toc-link')). toHaveLength(1);
```

src/client/rsg-components/TableOfContents/TableOfContents.spec.tsx
```diff
@@ -1,121 +1,95 @@
...
-	const actual = shallow(
-		<TableOfContentsRenderer classes={{}} searchTerm={searchTerm} onSearchTermChange={noop}>
-			<div>foo</div>
-		</TableOfContentsRenderer>
+	const { getByPlaceholderText, getAllByTestId, getByTestId } = render(
+		<TableOfContents sections={sections} />
 	);
-
-	expect(actual).toMatchSnapshot();
+	expect(getAllByTestId('rsg-toc-link').length).toBe(6);
```
sapegin, line 89:
```suggestion
	expect(getAllByTestId('rsg-toc-link')).toHaveLength(6);
```

elevatebart:
agreed

sapegin:
Yay, merging! 🦄

Issue #1621, opened 2020-06-26
Wrong union description with TypeScript


alextewpin:
**Current behavior**
When using TypeScript unions, react-styleguidist isn't providing any description. Moreover, is shows annoying empty div.

<img width="580" alt="Screenshot 2020-06-26 at 17 31 05" src="https://user-images.githubusercontent.com/1542068/85868602-2d0c8880-b7d3-11ea-9781-f32bc5cf3130.png">

**To reproduce**
Use TypeScript and add description to any union field. Demo: https://codesandbox.io/s/goofy-forest-zmeiy

**Expected behavior**
Unwrapped union description or at least no empty div.

**Version**
11.0.8

mitsuruog:
You can hover your mouse over the field for a few seconds to see the actual value, but I can understand the fact that more people don't realize it.

<img width="553" alt="スクリーンショット 2020-07-03 22 43 39" src="https://user-images.githubusercontent.com/1703219/86475298-5d2dcb80-bd7f-11ea-9043-67f48af491ea.png">

So... we can change this, but first, we need to decide what kind of UI we want. 
Here's my idea. Does anyone have any other good ideas?

1. Show the actual value instead
<img width="621" alt="スクリーンショット 2020-07-03 22 45 46" src="https://user-images.githubusercontent.com/1703219/86475721-2906da80-bd80-11ea-836b-2902688eda52.png">

2. Show the actual value below the description
<img width="554" alt="スクリーンショット 2020-07-03 22 47 45" src="https://user-images.githubusercontent.com/1703219/86475730-2e642500-bd80-11ea-9686-5411ff1ed172.png">

FYI @sapegin what do you think?

alextewpin:
Yeah, I know that I can hover on `union`, that's a cool feature. I'm actually bothered with ugly empty space between table rows. As far as I know, when parsing prop-types union there is actual description, like `One of: red, blue`. But not in this case.

Pic. 2 is fine I guess. Union description could be quite long, so putting it in type column may look sloppy.

sapegin:
I think unions can be very very long, and improving tooltips could be a better solution in general — showing a nice HTML tooltip where we can format code on multiple lines, instead of a `title`, which, I agree, isn't the best and easily discoverable UI pattern.

sapegin:
There are many places that could be improved with tooltips:

[...]

PR #1629, opened 2020-07-12
Fix Wrong union description with TypeScript #1621

mitsuruog:
Resolved:

- Wrong union description with TypeScript #1621

Main changes:
- Create a new component named `ComplexType` and `Tooltip` to show the complex type
- Introduce [tippyjs-react](https://github.com/atomiks/tippyjs-react) in the project
- ~~Add styleMock in the test assets to mock CSS import~~

Hover Actions:
Jul-12-2020 22-42-08

Keyboard Actions:
(use Tab key)
Jul-12-2020 22-42-58

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "node": ">=10"
   },
   "dependencies": {
+    "@tippyjs/react": "4.1.0",
     "@vxna/mini-html-webpack-template": "^1.0.0",
     "acorn": "^6.4.1",
     "acorn-jsx": "^5.1.0",
--- a/src/client/rsg-components/ComplexType/ComplexType.spec.tsx
+++ b/src/client/rsg-components/ComplexType/ComplexType.spec.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import ComplexType from './ComplexTypeRenderder';
+
+function renderComponent(name = 'color', raw = 'red | blue') {
+	return render(<ComplexType name={name} raw={raw} />);
+}
+
+describe('ComplexType', () => {
+	test('should render name', () => {
+		const { getByRole } = renderComponent();
+		expect(getByRole('button')).toHaveTextContent('color');
+	});
+
+	test('should render raw text in the tooltip', () => {
+		const { container, getByRole } = renderComponent();
+		fireEvent.focus(getByRole('button'));
+		expect(container.querySelector('[data-tippy-root]')).toHaveTextContent('red | blue');
+	});
+});
--- a/src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
+++ b/src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import Styled, { JssInjectedProps } from 'rsg-components/Styled';
+import { MdInfoOutline } from 'react-icons/md';
+import Text from 'rsg-components/Text';
+import Tooltip from 'rsg-components/Tooltip';
+import * as Rsg from '../../../typings';
+
+export const styles = ({ space }: Rsg.Theme) => ({
+	complexType: {
+		alignItems: 'center',
+		cursor: 'pointer',
+		display: 'inline-flex',
+		'& span': {
+			marginRight: space[0],
+			cursor: 'pointer',
+		},
+	},
+});
+
+export interface ComplexTypeProps extends JssInjectedProps {
+	name: string;
+	raw: string;
+}
+
+function ComplexTypeRenderer({ classes, name, raw }: ComplexTypeProps) {
+	return (
+		<Tooltip placement="right" content={raw}>
+			<span className={classes.complexType}>
+				<Text>{name}</Text>
+				<MdInfoOutline />
+			</span>
+		</Tooltip>
+	);
+}
+
+export default Styled<ComplexTypeProps>(styles)(ComplexTypeRenderer);
--- a/src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
+++ b/src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
@@ -8,12 +8,10 @@ import * as Rsg from '../../../typings';
 export const styles = ({ space }: Rsg.Theme) => ({
 	complexType: {
 		alignItems: 'center',
-		cursor: 'pointer',
 		display: 'inline-flex',
-		'& span': {
-			marginRight: space[0],
-			cursor: 'pointer',
-		},
+	},
+	icon: {
+		marginLeft: space[0],
 	},
 });
 
@@ -27,7 +25,7 @@ function ComplexTypeRenderer({ classes, name, raw }: ComplexTypeProps) {
 		<Tooltip placement="right" content={raw}>
 			<span className={classes.complexType}>
 				<Text>{name}</Text>
-				<MdInfoOutline />
+				<MdInfoOutline className={classes.icon} />
 			</span>
 		</Tooltip>
 	);
--- a/src/client/rsg-components/Props/renderType.tsx
+++ b/src/client/rsg-components/Props/renderType.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { PropTypeDescriptor } from 'react-docgen';
 import Type from 'rsg-components/Type';
-import Text from 'rsg-components/Text';
+import ComplexType from 'rsg-components/ComplexType';
 
 import { getType, PropDescriptor, TypeDescriptor } from './util';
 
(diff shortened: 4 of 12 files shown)


package.json
```diff
@@ -84,6 +85,7 @@
     "remark": "^11.0.1",
     "strip-html-comments": "^1.0.0",
     "terser-webpack-plugin": "^2.2.1",
+    "tippy.js": "6.2.5",
```
mitsuruog, line 88: This dependency is actually not needed tho. but I get a lint error during the test if I did not add it.\
Do you know of any better way to do it?

Given:
```jsx
import 'tippy.js/dist/tippy.css';
```

Then

```
'tippy.js' should be listed in the project's dependencies. Run 'npm i -S tippy.js' to add it  import/no-extraneous-dependencies
```

mitsuruog:
I can't pass the CI steps of `npm run build:sections` due to the next error. 

```
asset size limit: The following asset(s) exceed the recommended size limit (1.05 MiB).
```

@sapegin 
Can I remove the performance settings here?
https://github.com/styleguidist/react-styleguidist/blob/master/examples/sections/styleguide.config.js#L92-L99

src/client/rsg-components/ComplexType/ComplexTypeRenderer.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { MdInfoOutline } from 'react-icons/md';
+import Tippy from '@tippyjs/react';
+import 'tippy.js/dist/tippy.css';
```
sapegin, line 5: This import means css-loader is now required to use Styleguidst, I'd really like to avoid that. I think we should have our own styles written using JSS.

src/client/rsg-components/ComplexType/ComplexTypeRenderer.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
sapegin, line 1: Let's extract the tooltip to its own component — there are other use cases for it.

src/client/rsg-components/ComplexType/ComplexType.spec.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { ComplexTypeRenderer, styles } from './ComplexTypeRenderer';
+
+const props = {
+	classes: classes(styles),
+	name: 'color',
+	raw: '"blue" | "red"',
+};
+
+it('renderer should render complex type', () => {
+	const actual = shallow(<ComplexTypeRenderer {...props} />);
```
sapegin, line 12: Let's avoid shallow rendering and snapshot testing: both patters are very problematic. Unfortunately, we have a lot of tests using both but it's not a reason to create new ones ;-)

https://blog.sapegin.me/all/snapshot-tests/

mitsuruog: That is awesome posts! I'll change the test to less problematic one.

src/client/rsg-components/ComplexType/ComplexTypeRenderer.tsx
mitsuruog: I agree with you in terms of consistency. Since the tooltip will be an individual component, I'll consider an alternative implementation.

mitsuruog:
Actually the bundle size is increased `1.1MB => 1.5MB`. In my opinion, this increase is unacceptable.
However, it is almost close to the limit as it is now, and it may have the same problem again in the near future.

sapegin:
That's the idea: the limit should be very close to the actual size so any increase is visible. Let's check again with custom styles, I believe they'll be much smaller.

mitsuruog:
@sapegin Hi, I have updated my PR but I am still facing the CI step problem.
Actually the bundle size is increased `1.1MB => 1.11MB` not so bad.

sapegin:
Have you tried to update the limit?

mitsuruog:
not yet. I can update it, but what do you think the next limit should be?

sapegin:
As I've already explained: as close as possible, so 1.15 should be fine.

mitsuruog:
hmmm... I have no idea how to increase the coverage any further.

sapegin:
I believe covering this line is the way:

```jsx
return <Type>unknown</Type>;
```

mitsuruog:
Do you know what input to give to reach this case?

src/client/rsg-components/Props/renderType.tsx
```diff
@@ -28,31 +28,19 @@ export function renderType(type: ExtendedPropTypeDescriptor): string {
...
-function renderComplexType(name: string, title: string): React.ReactNode {
-	return (
-		<Text size="small" underlined title={title}>
-			{name}
-		</Text>
-	);
-}
-
 function renderAdvancedType(type: TypeDescriptor): React.ReactNode {
-	if (!type) {
```
mitsuruog, line 40: @sapegin 
I think the line above is usually unreachable code for now.

Because:
- The `type` passed to this function is based on `prop.flowType` or `prop.tsType`.
- This function does not enter the branch unless `prop.flowType` or `prop.tsType` is defined.

Therefore, what I can do is:

- Remove unreachable codes to increase coverage.
- Test only this function by itself to get the coverage

what do you think?

I'd like to keep this line and test it on its own, as there may be some changes in the future that might unintentionally go into this branch. in my opinion tho.

sapegin: I think removing makes more sense — according to code and types it should be defined. And it would be great to fix `type as any` on the call site.

mitsuruog: > And it would be great to fix type as any on the call site.

good point. haha.
but It may be difficult in places where the type definitions are a bit mixed up. 
I'll try it out for a bit, but if it's difficult, I think I'll ask for a review as it is.

sapegin: Thanks! Don't spend too much time on it though ;-)

src/typings/dependencies/react-docgen.ts
```diff
@@ -58,7 +58,6 @@ declare module 'react-docgen' {
 			| 'objectOf'
 			| 'shape'
 			| 'exact'
-			| 'union'
```
mitsuruog, line 61: duplicated definition for `TypeUnionDescriptor`

mitsuruog:
@sapegin please review me again when you have time!

src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Styled, { JssInjectedProps } from 'rsg-components/Styled';
+import { MdInfoOutline } from 'react-icons/md';
+import Text from 'rsg-components/Text';
+import Tooltip from 'rsg-components/Tooltip';
+import * as Rsg from '../../../typings';
+
+export const styles = ({ space }: Rsg.Theme) => ({
+	complexType: {
+		alignItems: 'center',
+		cursor: 'pointer',
```
sapegin, line 11: Shouldn't this be in the Tooltip component?

src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
```diff
@@ -0,0 +1,36 @@
...
+		display: 'inline-flex',
+		'& span': {
```
sapegin, line 13: What's this span? Looks like this selector is very fragile and relies on the implementation of another component.

src/client/rsg-components/Tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+	);
+}
+
+describe('Tooltip', () => {
+	test('should render child component as is', () => {
+		const { container, getByTestId } = renderComponent();
+		expect(container).toContainElement(getByTestId('child'));
+	});
+
+	test('should the child component be wrapped by "role=button" element', () => {
```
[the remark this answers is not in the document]
mitsuruog: it's fine to remove it as it's been tested in other cases.

src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
mitsuruog: Good point. this span is actually the icon. but I should have used a more safe way to avoid breaking.

src/client/rsg-components/ComplexType/ComplexTypeRenderder.tsx
mitsuruog: I moved it to the tooltip, but the cursor didn't change as I expected, so I added a style here.
However, I think this cursor should change depending on the component it's wrapped around, so I would like to remove it for now.

mitsuruog:
@sapegin please review me again.

mitsuruog:
@sapegin review me, please.

sapegin:
Thanks, merging! Sorry for the delay ;-/

PR #1641, opened 2020-07-22
Use fastest-levenshtein instead of leven

ka-weihe:
`fastest-levenshtein` [(link)](https://github.com/ka-weihe/fastest-levenshtein) is much faster than `leven`. 

### Benchmarks
N = length of strings
Numbers are in ops/sec
Higher is better.
| Test Target               | N=4   | N=8   | N=16  | N=32 | N=64  | N=128 | N=256 | N=512 | N=1024 |
|---------------------------|-------|-------|-------|------|-------|-------|-------|-------|--------|
| fastest-levenshtein       | 44423 | 23702 | 10764 | 4595 | 1049  | 291.5 | 86.64 | 22.24 | 5.473  |
| leven                     | 19688 | 6884  | 1606  | 436  | 117   | 30.34 | 7.604 | 1.929 | 0.478  |


src/scripts/utils/sanitizeConfig.ts
```diff
@@ -10,7 +10,7 @@ import isFinite from 'lodash/isFinite';
 import map from 'lodash/map';
 import listify from 'listify';
 import kleur from 'kleur';
-import leven from 'leven';
+import levenshtein from 'fastest-levenshtein';
```
sapegin, line 13: Why not?

```suggestion
import { distance } from 'fastest-levenshtein';
```

PR #1651, opened 2020-08-01
Fix: Fix site links for adblock users

Anne-Victoria:
When using e.g. uBlock Origin, navigation in the styleguidist documentation is broken. This is because in that case "window.goatcounter" is undefined. I added a check that assures, goatcounter.count will only be called if goatcounter is defined. This way clicking on links and navigating the documentation still works when goatcounter is blocked.


Anne-Victoria:
Thank you for merging! This is the first time I sent a PR to an open source project, so I'm super excited about this working out :star2:

Issue #1667, opened 2020-08-20
Warning in ArgumentRenderer


lyz810:
**Current behavior**

Version: ^11.0.8
Report warning when document has a public method in class
```
Warning: Failed prop type: The prop `name` is marked as required in `ArgumentRenderer`, but its value is `undefined`.
```

Since v10.3.0, it started to use tsx, and name props in ArgumentRenderer became a required prop:
https://github.com/styleguidist/react-styleguidist/blob/master/src/client/rsg-components/Argument/ArgumentRenderer.tsx#L18
However, the `@returns` param don't have a `name` props

**To reproduce**

```js
import React from 'react';
import PropTypes from 'prop-types';

export default class Issue extends React.Component {
    static propTypes = {
        myProps: PropTypes.bool,
    };

    /**
     * Public Method
     *
     * @public
     *
     * @returns {Boolean} return a Boolean Value
     */
    somePublicMethod = () => {
        // doSomething

        return this.props.myProps;
    }

    render() {
        return 'Open console, see warnings';
    }
}
```
https://github.com/lyz810/styleguidist-example/blob/master/src/components/Issue.js
please open console, and check the warning printed.
**Expected behavior**

Don't show warning when using `@returns` param, or don't consider `name` props of ArgumentRenderer as required props.

sapegin:
Feel free to send a pull request with a fix! 👾

PR #1671, opened 2020-08-27
Fix Warning in ArgumentRenderer #1667

mitsuruog:
Resolved

- Warning in ArgumentRenderer #1667

I checked the behavior of the component of the original issue after I changed the `name` to optional and found that the following error occurred.

```
DoctrineError {name: "DoctrineError", message: "Unknown type undefined"}
```

What I have found in my research is that in the above cases, `type.type` in the `returns` is lost in `getProps` in this case.
`returns` should have `NameExpression` in `type.type` in my opinion.

```js
// I saw this shape.
{ 
  description: 'return a Boolean Value', 
  type: { name: 'boolean' } 
}

// but it should be this shape.
{
  title: 'returns',
  description: 'return a Boolean Value',
  type: { type: 'NameExpression', name: 'boolean' }
}
```

### DIFF
--- a/src/loaders/utils/getProps.ts
+++ b/src/loaders/utils/getProps.ts
@@ -94,7 +94,15 @@ export default function getProps(doc: DocumentationObject, filepath?: string): R
 			allTags as TagProps,
 			JS_DOC_METHOD_RETURN_TAG_SYNONYMS
 		) as TagParamObject[];
-		const returns = (method.returns || returnTags[0]) && { ...method.returns, ...returnTags[0] };
+		const returns = method.returns
+			? {
+					...method.returns,
+					type: {
+						type: 'NameExpression',
+						...method.returns.type,
+					},
+			  }
+			: returnTags[0];
 
 		if (returns) {
 			method.returns = returns;
--- a/src/loaders/utils/getProps.ts
+++ b/src/loaders/utils/getProps.ts
@@ -94,7 +94,7 @@ export default function getProps(doc: DocumentationObject, filepath?: string): R
 			allTags as TagProps,
 			JS_DOC_METHOD_RETURN_TAG_SYNONYMS
 		) as TagParamObject[];
-		const returns = method.returns || returnTags[0];
+		const returns = (method.returns || returnTags[0]) && { ...method.returns, ...returnTags[0] };
 
 		if (returns) {
 			method.returns = returns;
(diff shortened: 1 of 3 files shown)


src/loaders/utils/getProps.ts
```diff
@@ -94,7 +94,7 @@ export default function getProps(doc: DocumentationObject, filepath?: string): R
 			allTags as TagProps,
 			JS_DOC_METHOD_RETURN_TAG_SYNONYMS
 		) as TagParamObject[];
-		const returns = method.returns || returnTags[0];
+		const returns = (method.returns || returnTags[0]) && { ...method.returns, ...returnTags[0] };
```
sapegin, line 97: Looks like this isn't different from the previous version, unless I'm missing something (in this case we should make it more clear).

We're returning `method.returns` if it's truthy, or `returnTags[0]`.

sapegin (changes requested):
Thanks for the pull request!

src/loaders/utils/getProps.ts
mitsuruog: hmm... we need to make it more clear. in my case, `method.returns` have this object:

```js
{ 
  description: 'return a Boolean Value', 
  type: { name: 'boolean' } 
}
```

and then `returnTags[0]` have this object:

```js
{
  title: 'returns',
  description: 'return a Boolean Value',
  type: { type: 'NameExpression', name: 'boolean' }
}
```

`method.returns` will have occurred `DoctrineError` when it does not have `{ type: 'NameExpression' }`.
so I figured that merging both results would result in the intended object shape. 
But I'm also sure I don't understand enough about this logic.

What should the correct object shape of `results` be?

sapegin: Now I'm even more confused. So now we're overwriting anything in `method.returns` with `returnTags[0]` but previously `method.returns`  had higher priority.

If `method.returns`  is missing the type, why not add it there? Something like this:

```
const getReturns = ({method, tags}) => {
                 if (method.returns) {
                       return {
                           ...method.returns,
                           type: { type: 'NameExpression', ...method.returns.type } 
                       }
                 }

		const returnTags = getMergedTag(
			tags as TagProps,
			JS_DOC_METHOD_RETURN_TAG_SYNONYMS
		) as TagParamObject[];

                 return returnTags[0];
}
```

mitsuruog: > previously `method.returns` had higher priority.

I got it. I’ll add a change on the `method.returns` side.

mitsuruog:
@sapegin please review me.

sapegin (approved):
🦄

PR #1675, opened 2020-09-02
Fix visibleName with dots

Lazyuki:
Fixes: https://github.com/styleguidist/react-docgen-typescript/issues/292

With react-docgen-typescript's version 1.20.4, components with `@visibleName` JSDoc tags are working. However, it changes the react component's `displayName` to the one specified in `@visibleName`, and when it contains a `.`, the code examples were not working. This was because Styleguidist was trying to import the component with the dot in its name in the code example which resulted in the syntax error.

Example:
```tsx
// ===== DialogTitle.tsx ======
/**
 * @visibleName Dialog.Title
 */
export const DialogTitle = props => (<h2 {...props} />);

// ===== Dialog.tsx ======
import DialogTitle from './DialogTitle.tsx';
const Dialog = ({ children }) => <div>{children}</div>;
Dialog.Title = DialogTitle;
export default Dialog;
```
This caused Styleguidist to try 
```jsx
const Dialog.Title$0 = require('./DialogTitle.tsx');
```
in the example code renderer, which resulted in `SyntaxError: Missing initializer in const declaration`.

### DIFF
--- a/src/loaders/__tests__/examples-loader.spec.ts
+++ b/src/loaders/__tests__/examples-loader.spec.ts
@@ -9,7 +9,17 @@ const query = {
 	shouldShowDefaultExample: false,
 };
 
+const subComponentQuery = {
+	file: '../fooSub.js',
+	displayName: 'FooComponent.SubComponent',
+	shouldShowDefaultExample: false,
+};
+
+
 const getQuery = (options = {}) => encode({ ...query, ...options }, '?');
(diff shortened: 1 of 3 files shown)


311chaos:
I had this issue come up, but with different characters. I had visible names with hyphens or commas that also stopped working.

Lazyuki:
Oh I didn't expect those characters to be in `@visibleName`, but I guess that's possible. I can modify this to exclude all non-alphabetical characters.

sapegin:
Thanks!

Issue #1674, opened 2020-09-01
initialOpen as an option in sections from the config


josephrexme:
**The problem**
It will be great if the `initialOpen` option [on this line](https://github.com/styleguidist/react-styleguidist/blob/9b6b6f60051f170dcb7f548fb114420a48a8e1e9/src/client/rsg-components/TableOfContents/TableOfContents.tsx#L69) is allowed to be passed from the styleguide.config with each section similar to `sectionDepth`.

As I was going to make a PR, I had thought simply removing it from that line will make the one I set in config available to use in the `ComponentListRenderer` but then I realized every custom options passed to our sections in the config aren't being sent all the way to `ComponentsList` [here](https://github.com/styleguidist/react-styleguidist/blob/9b6b6f60051f170dcb7f548fb114420a48a8e1e9/src/client/rsg-components/TableOfContents/TableOfContents.tsx#L74).

I understand if we don't want to make `initialOpen` available at the config level for some reasons but I would like to implement this effect myself of having specific sections in toc expanded and I can't do that if I don't get my custom parameters. Here's me trying to use `expand`:

```js
module.exports = {
  sections: [
    {
      name: 'Components',
      content: 'docs/ui.md',
      components: 'components/**/index.js',
      exampleMode: 'collapse',
      usageMode: 'expand',
      expanded: true, // need access to this
      sectionDepth: 1,
    }
  ]
}
```

**Proposed solution**

Make `initialOpen` set per section from the config and do not override but instead provide defaults for when user didn't set

**Alternative solutions**

Passing all user options at config and always making them available for developers writing custom styleguide render files

josephrexme:
Made a PR at #1689

PR #1689, opened 2020-09-07
Add expand and allow custom root section options

josephrexme:
This addresses #1674 which allows user set and `expand` setting for each root section. It also sends all custom options from the user through to be usable


sapegin (changes requested):
Thanks for the pull request! Please update the docs and tests.

josephrexme:
Thanks for taking a look @sapegin . I have just added tests. The [syntax highlighting test here](https://github.com/styleguidist/react-styleguidist/pull/1689/files#diff-d56e2054ed9db0792599da8bcf6ce89b) that I modified was failing even though I hadn't touched it so I made it look like the value it was expecting. I can revert that if it's not needed. I don't know if I also had to include an updated snapshot

sapegin:
> The syntax highlighting test here that I modified was failing even though I hadn't touched it so I made it look like the value it was expecting. I can revert that if it's not needed. I don't know if I also had to include an updated snapshot

This is because Prism was updated at some point, and the snapshot is already fixed in master.

docs/Configuration.md
```diff
@@ -136,6 +136,12 @@ Defines the initial state of the example code tab:
 - `hide`: hide the tab and it can´t be toggled in the UI.
 - `expand`: expand the tab by default.
 
+## `expand`
```
sapegin, line 139: I'm confused: this isn't a global configuration option, right? If so, we should document it here:

https://github.com/styleguidist/react-styleguidist/blob/master/docs/Components.md#sections

src/loaders/utils/getSections.ts
```diff
@@ -88,6 +88,7 @@ export function processSection(
 	}
 
 	return {
+		...section,
 		name: section.name,
```
sapegin, line 92: We don't need this line anymore, and also `description` and `external` below.

sapegin (changes requested):

docs/Configuration.md
josephrexme: Thank you!

josephrexme:
> > The syntax highlighting test here that I modified was failing even though I hadn't touched it so I made it look like the value it was expecting. I can revert that if it's not needed. I don't know if I also had to include an updated snapshot
> 
> This is because Prism was updated at some point, and the snapshot is already fixed in master.

To be clear, should I undo my change with the **highlightCode.spec.ts** file?

sapegin:
Try to merge master into your branch.

josephrexme:
Thanks for the reviews and accepting my PR @sapegin

PR #1690, opened 2020-09-08
Fix broken assetsDir due to updated copy-webpack-plugin

### LINKED ISSUE #1676: assetsDir broken in 11.0.9 with create-react-app
**Current behavior**

After updating to 11.0.9 it is no longer possible to build the styleguide if using assetsDir (which invokes the webpack copy plugin)

**To reproduce**

`styleguideist build`

```
Building style guide...
ValidationError: Invalid options object. Copy Plugin has been initialized using an options object that does not match the API schema.
 - options[0] misses the property 'patterns'. Should be:
   [non-empty string | object { from, to?, context?, globOptions?, filter?, toType?, force?, flatten?, transform?, cacheTransform?, transformPath?, noErrorOnMissing? }, ...] (should not have fewer than 1 item)
ValidationError: Invalid options object. Copy Plugin has been initialized using an options object that does not match the API schema.
 - options[0] misses the property 'patterns'. Should be:
   [non-empty string | object { from, to?, context?, globOptions?, filter?, toType?, force?, flatten?, transform?, cacheTransform?, transformPath?, noErrorOnMissing? }, ...] (should not have fewer than 1 item)
    at validate (.../node_modules/react-styleguidist/node_modules/schema-utils/dist/validate.js:98:11)
    at new CopyPlugin (.../node_modules/react-styleguidist/node_modules/copy-webpack-plugin/dist/index.js:32:30)
    at _default (.../node_modules/react-styleguidist/lib/scripts/make-webpack-config.js:114:34)
    at build (.../node_modules/react-styleguidist/lib/scripts/build.js:13:63)
    at commandBuild (.../node_modules/react-styleguidist/lib/bin/styleguidist.js:105:20)
    at Object.<anonymous> (.../node_modules/react-styleguidist/lib/bin/styleguidist.js:71:5)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:754:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)

```
If i remove
`assetsDir: 'src/assets/images/',`

Then all builds as normal. I noticed webpack copy plugin has been updated to 6.0.x, downgrading this alone to 5.x.x does not fix the issue. 

Downgrading styleguidist to 11.0.8 does allow the styleguide to build.

**Expected behavior**

Styleguide should build

swashata:
Because copy-webpack-pluign was updated, the assetsDir was broken. We need to update the config being passed to the plugin to make it work again. Update the plugin to latest version 6.1.0 and update the config and tests to make it work.

- [x] Update `copy-webpack-plugin` to `^6.1.0`.
- [x] Fix configuration passed to `copy-webpack-plugin` [`make-webpack-config.ts:111`].
- [x] Update tests to reflect new changes.
- [x] Fix a non-relevant failing tests in `highlightCode.spec.ts` so that build passes.

Fixes #1676


src/scripts/make-webpack-config.ts
```diff
@@ -108,8 +108,14 @@ export default function(
 			},
 		});
 		if (config.assetsDir && webpackConfig.plugins) {
+			const copyPatterns = {
+				patterns: castArray(config.assetsDir).map(dir => ({ from: dir })),
+			};
 			webpackConfig.plugins.push(
-				new CopyWebpackPlugin(castArray(config.assetsDir).map(dir => ({ from: dir })))
+				// since we don't have the type of copy-webpack-plugin@6.0
```
sapegin, line 115: To make it more visible:

```suggestion
				// FIXME: Since we don't have the type of copy-webpack-plugin@6.0
```

sapegin (changes requested):
Thanks! One tiny suggestion.

swashata:
@sapegin Done.

Issue #1525, opened 2020-01-22
Custom styles are not applied to minified build


Xiphe:
**Current behavior**

This has been reported earlier: https://github.com/styleguidist/react-styleguidist/issues/1431

When I run `styleguidist build` with a webpack config that minifies function names then my [custom styles](https://react-styleguidist.js.org/docs/configuration.html#styles) are not applied.

**To reproduce**

https://github.com/McManning/styleguidist-demo

**Expected behavior**

I expect the minified bundle to still apply my custom styles.

Xiphe:
The problem comes from the [`Styled` HOC](https://github.com/styleguidist/react-styleguidist/blob/master/src/client/rsg-components/Styled/Styled.tsx#L14) relying on the function name.

I would propose to refactor this to `const componentName = (WrappedComponent.displayName || WrappedComponent.name).replace(/Renderer$/, '');` and explicitly decorate all rsg components with their `displayNames` as string. Like `LogoRenderer.displayName = 'LogoRenderer';` This way the StyleHOC can identify the rsg components even after they are minified.

Let me know what you think, I'd happily provide a PR for this.

BboyAwey:
I've met this shit too. Is there any solution on it ?

Eddie-CooRo:
@BboyAwey Disabling the minification worked for me.

BboyAwey:
Create a new file named Styled.fixed.js which is copied from `react-styleguidist/lib/client/rsg-components/Styled/Styled.js`, and fixe this bug like below, then put it into our `src` dir, then import it in the custom styleguide components instead of importing this file directly from it's origin position.

```js
/* eslint-disable */
/**
 * this file will fix a styleguidist bug:
 * https://github.com/styleguidist/react-styleguidist/issues/1525
 * and if there's any fix solution from styleguidist
 * please remove this file and use the latest version of styleguidist
 * 
 * Awey 2020-04-30
 */
import "core-js/modules/es.function.name";
import "core-js/modules/es.object.assign";
import "core-js/modules/es.regexp.exec";
import "core-js/modules/es.string.replace";
import React, { Component } from 'react';
import Context from 'rsg-components/Context';
import createStyleSheet from 'react-styleguidist/lib/client/styles/createStyleSheet';

function _assertThisInitialized(self) { if (self === void 0) { throw new ReferenceError("this hasn't been initialised - super() hasn't been called"); } return self; }

[...]

PR #1696, opened 2020-09-21
Don't (only) rely on function names for custom styles

Xiphe:
Fix https://github.com/styleguidist/react-styleguidist/issues/1525

This makes the `Styled` HOC aware of a static `displayName` which will survive code-minifying and therefore will apply custom styles even when the component function names are mangled. 

---

Re: https://github.com/styleguidist/react-styleguidist/issues/1525#issuecomment-695938069 I see how [my comment](https://github.com/styleguidist/react-styleguidist/issues/1525#issuecomment-695775762) was offending and am sorry! I don't take your work here for granted and am grateful for your contribution to the react eco-system 🙏 

I was salty because I offered submitting a PR [8 months ago](https://github.com/styleguidist/react-styleguidist/issues/1525#issuecomment-577207560) and have not gotten any feedback from maintainers but a bot closing down my attempt to start a conversation. 

Still this does not excuse me "demanding" your involvement. So again: I'm sorry and thanks for giving me the chance to learn.

### DIFF
--- a/src/client/rsg-components/Argument/ArgumentRenderer.tsx
+++ b/src/client/rsg-components/Argument/ArgumentRenderer.tsx
@@ -77,5 +77,6 @@ ArgumentRenderer.propTypes = {
 	returns: PropTypes.bool,
 	block: PropTypes.bool,
 };
+ArgumentRenderer.displayName = 'ArgumentRenderer';
 
 export default Styled<ArgumentPropsWithClasses>(styles)(ArgumentRenderer);
--- a/src/client/rsg-components/Arguments/ArgumentsRenderer.tsx
+++ b/src/client/rsg-components/Arguments/ArgumentsRenderer.tsx
@@ -54,5 +54,6 @@ ArgumentsRenderer.propTypes = {
 	).isRequired,
 	heading: PropTypes.bool,
 };
+ArgumentsRenderer.displayName = 'ArgumentsRenderer';
 
 export default Styled<ArgumentsProps>(styles)(ArgumentsRenderer);
--- a/src/client/rsg-components/Code/CodeRenderer.tsx
+++ b/src/client/rsg-components/Code/CodeRenderer.tsx
@@ -24,5 +24,6 @@ CodeRenderer.propTypes = {
 	classes: PropTypes.objectOf(PropTypes.string.isRequired).isRequired,
 	children: PropTypes.node.isRequired,
 };
+CodeRenderer.displayName = 'CodeRenderer';
 
 export default Styled<CodeProps>(styles)(CodeRenderer);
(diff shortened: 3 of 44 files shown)


sapegin (commented):
Thanks for the pull request!

Function names aren't supposed to be minified:

https://github.com/styleguidist/react-styleguidist/blob/85123153e0816110d492cdb22b43f651b30f23d6/src/scripts/make-webpack-config.ts#L73-L90

I'm wondering why it's not enough, and whether assigning static `displayName` is a better solution. One downside I see is code reviewing will be more difficult because we'll have to make sure all renderer component have names. Maybe we can lint this somehow...

sapegin:
> I was salty because I offered submitting a PR 8 months ago and have not gotten any feedback from maintainers but a bot closing down my attempt to start a conversation.

Unfortunately, I don't have the capacity to check issues regularly. I'd recommend to either open a pull request right away if that's something simple or ping me in comments if it's not.

Xiphe:
> Function names aren't supposed to be minified

Why? I always minify fnames for production. I see that it does not work with this project but from my point of view thats snowflake behaviour.

> I'm wondering why it's not enough, and whether assigning static displayName is a better solution.

I agree that `keep_fnames` also solve the issue but in my case it took me half a day researching issues of this project and digging into the source until i found that as a solution.
So adding explicit `displayName` is better because it makes this project easier and more intuitive to use.

> Maybe we can lint this somehow...

I can install [eslint#react/displayName](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md) if you'd like.
I'd personally apply the [rule-of-three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) and let it slip a couple of times until I add additional linters for that.

sapegin:
> Why? I always minify fnames for production. I see that it does not work with this project but from my point of view thats snowflake behaviour.

Because that's the setup for all Styleguidist builds. The question is why it doesn't work anymore. Manually setting `displayName` isn't less “snowflake”.

> I'd personally apply the rule-of-three and let it slip a couple of times until I add additional linters for that.

This means someone will have to spend another half-a-day debugging why this doesn't work again. 

> I can install eslint#react/displayName if you'd like.

Looks like this rule with `ignoreTranspilerName` option is what we need for our use case? If that's true, I'm fine with having `displayName` property.

Xiphe:
> Because that's the setup for all Styleguidist builds. The question is why it doesn't work anymore. 

In January when I opened the issue it was because i've [used a custom webpack config](https://react-styleguidist.js.org/docs/webpack). Now when I read the docs it seems that the customisations I made are not possible anymore. I don't work on the project anymore so I can not test if it may just work currently.

If you say it should just work™ we can close this until the problem happens again.

> Manually setting displayName isn't less “snowflake”.

I disagree. Commonly in react when I want to identify a component by name I use `displayName` not the function name.

See [React docs on easy-debugging-hocs](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) or [@types/react having the prop on all component types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L551)

PR #1697, opened 2020-09-22
Update cookbook documentation on CSS animation usage

### LINKED ISSUE #1695: @keyframes fail to work with rsg-components/Styled
When I try to use a @keyframe animation following this [cookbook example](https://github.com/styleguidist/react-styleguidist/blob/e662394aea164c78f11408cddca4bcfba2998cc2/docs/Cookbook.md#how-to-change-styles-of-a-style-guide), I get:

```
TypeError: container.addRule(...).addRule is not a function
    in Styled(StyleGuide) (created by StyleGuide)
    in StyleGuide
```
At first I tried to do this in the component I was overriding and got that error, then I tried to do it from the style section of the configuration and got the same. Here's what the code looks like:

```js
content: {
      opacity: 0,
      animation: 'fade .3s ease .5s forwards',
      '@keyframes fade': {
        to: { opacity: 1 }
      },
}
```

josephrexme:
Fixes #1695

### DIFF
--- a/docs/Cookbook.md
+++ b/docs/Cookbook.md
@@ -217,7 +217,7 @@ module.exports = {
       // We're changing the LogoRenderer component
       logo: {
         // We're changing the rsg--logo-XX class name inside the component
-        animation: 'blink ease-in-out 300ms infinite'
+        animation: '$blink ease-in-out 300ms infinite'
       },
       '@keyframes blink': {
         to: { opacity: 0 }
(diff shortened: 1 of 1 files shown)


docs/Cookbook.md
```diff
@@ -299,6 +299,12 @@ module.exports = {
 }
 ```
 
+## How to use CSS animations in your style guide?
+As seen in the `@keyframes` animation examples above, the animation property in CSS rules do
+not directly use the name of their keyframe animations because of internal keyframe scoping.
```
sapegin, line 304:
```suggestion

As seen in the `@keyframes` animation examples above, the animation property in CSS rules do not directly use the name of their keyframe animations because of internal keyframe scoping.
```

sapegin (approved):
Thanks! Just one tiny tweak.

Issue #1661, opened 2020-08-13
Para components should use `theme.fontSize.base` or `theme.fontSize.text`


juanca:
**Current behavior**

When stylegudist is configured to use a font size for base and/or text to be anything other than 16px, the paragraph content generated from markdown does not use the specified font size.

Question: Is a markdown paragraph considered to be either `base` or `text`? I am assuming it should be `text` at the very least.

**To reproduce**

1. Config:

    ```
    module.exports = {
      theme: {
        fontSize: {
          base: 24,
          text: 36,
        },
      },
    };
    ```

1. Some markdown file:

    ```
    # I am an H1

    I am just a paragraph! My font size should be really big!
    ```

https://github.com/styleguidist/example/pull/8

**Expected behavior**

The generated paragraph content should use the theme provided. 

Or there should be some explanation of how the "theme" is defined: How is a paragraph not "text" or "base"?

juanca:
At the moment, my solution is to duplicate the theme as a direct style to the Para component. Something like:

```
styles: {
  Para: { para: {
    fontSize: 14,
  }},
}
```

sapegin:
I think it has `inherit` because it's used in `Props` where it should be smaller, and I think it shouldn't be used here:

https://github.com/styleguidist/react-styleguidist/blob/53bc4bb357133e2208ca28e5181b44306fdc4e3d/src/client/rsg-components/Props/PropsRenderer.tsx#L24

And the `text` would be the correct size.

Feel free to send a pull request with a fix! 👾

senadglojnaric:
Hello, I would like to be assigned to this issue.

sapegin:
@senadglojnaric sure, go for it! 👍

senadglojnaric:
I have a question about testing. I ran a test on my clone repo without changing anything in the code. Like an initial test.
And it failed.

Here's my debug log file:

0 info it worked if it ends with ok
1 verbose cli [
1 verbose cli   'C:\\Program Files\\nodejs\\node.exe',
1 verbose cli   'C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'run',
1 verbose cli   'lint'
1 verbose cli ]
2 info using npm@6.14.8
3 info using node@v12.18.4
4 verbose run-script [ 'prelint', 'lint', 'postlint' ]
[...]

PR #1698, opened 2020-10-02
Update ExamplesRenderer.tsx

senadglojnaric:
Fix issue #1661  with font size.
ExamplesRenderer.tsx is updated so that now paragraphs from Readme.md have the same font size as other text after the theme is set.

### DIFF
--- a/src/client/rsg-components/Examples/ExamplesRenderer.tsx
+++ b/src/client/rsg-components/Examples/ExamplesRenderer.tsx
@@ -1,10 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Styled, { JssInjectedProps } from 'rsg-components/Styled';
+import * as Rsg from '../../../typings';
 
-const styles = () => ({
-	// Just default jss-isolate rules
-	root: {},
+const styles = ({ fontSize }: Rsg.Theme) => ({
+	article: {
+		fontSize: fontSize.text,
+	},
 });
 
 interface ExamplesRendererProps extends JssInjectedProps {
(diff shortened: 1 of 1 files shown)


sapegin (commented):
Thanks for the pull request! 

Why did you choose this solution over [the proposed one](https://github.com/styleguidist/react-styleguidist/issues/1661#issuecomment-678952897)?

senadglojnaric:
Hello, first of all, I'm sorry for the late answer.

I tried the proposed solution first. But that piece of code is actually responsible for "props & methods"  table. 
Here is a screenshot:
image

And the text that remained small after the theme change was in the examples section. And that comes from _Readme.md_ file.

I found that ExamplesRenderer is responsible for rendering that text and that Para inherits fontsize from a parent container.

So I added a fontsize from a theme to a container where Para is. Para then can inherit fontsize from that container. 

Now when a theme is set, everything changes accordingly.

sapegin:
The idea was to set the correct size on the primitive level instead of having the need to set it on the container, and don't use the `Para` when we need a different size.

senadglojnaric:
Ok. If I understand You right, I could use another component that would be rendered instead of Para when different fontsize in theme is set?

mitsuruog:
@senadglojnaric @sapegin how is it going? I can try to fix it by making another PR If this PR has stopped.

senadglojnaric:
You can do it.

senadglojnaric:
yes i think you can do that.


uto, 17. stu 2020. 15:18 Mitsuru Ogawa <notifications@github.com> je
napisao:

PR #1700, opened 2020-10-08
feat: Add compiler option, replace Buble with Sucrase

### LINKED ISSUE #462: Create an option for a compiler
If someone has a certain babel configuration and wants to use the same code in the styleguide readme, we need to enable swaping out buble for babel.

sapegin:
1. Replace [Bublé](https://buble.surge.sh/) with [Sucrase](https://github.com/alangpierce/sucrase/): faster, supports more ECMAScript features, like imports (we used to transpile them ourselves), optional chaining, class fields, etc.

2. Allow custom compilers, like Babel or TypeScript using the new `compilerModule` and `compileExample` option.

Closes #462

##  New docs preview

https://deploy-preview-1700--styleguidist.netlify.app/docs/configuration#compilermodule

https://deploy-preview-1700--styleguidist.netlify.app/docs/configuration#compileexample


docs/Configuration.md
```diff
@@ -8,28 +8,64 @@ Type: `String` or `Array`, optional
 
 Your application static assets folder will be accessible as `/` in the style guide dev server.
 
+## `compiler`
+
+Type: `String`, default: `sucrase`
```
okonet, line 13: I'm wondering why you went with the `String` here instead of importing the compiler in the config file like

```javascript
import surcase from "surcase";

module.exports = {
  ...
  compiler: surcase
}
```

Specifically I think it's important to keep in mind that this option should allow plugging any compiler here (like TypeScript) so I'm wondering if we should also normalize and expect a function here?

docs/Configuration.md
```diff
@@ -8,28 +8,64 @@ Type: `String` or `Array`, optional
...
+
+You can change the compiler by using this option:
+
+```javascript
+module.exports = {
+  compiler: '@babel/standalone'
+}
+```
+
+The value of the `compiler` option is the name of the compiler npm package. The module should have a named export `transform` with the signature:
```
okonet, line 43: And that's exactly why I think it would be better to not rely on that and allow importing a function with a signature and make the parameter a function so even if the package doesn't have this export or uses a bit different signature, we could still easily plug it in.

docs/Configuration.md
```diff
@@ -8,28 +8,64 @@ Type: `String` or `Array`, optional
...
+;(code: string, compilerConfig: Record<string, any>) => {
+  code: string
+}
+```
+
+> **Warning:** You should change [the compilerConfig option too](#compilerconfig) to pass the options for the new compiler.
+
+> **Info:** See [the cookbook](Cookbook.md) to learn how to configure other compiler, like Babel or TypeScript.
+
 ## `compilerConfig`
```
okonet, line 55: What do you think about merging those too into one `complier` option? 

```javascript

compiler: (code, options) => transformFunction(code, options)
```

This should help with integrating with TypeScript which relies on `tsconfig.json`

docs/Configuration.md
sapegin: The thing is it's _not_ a function, it's a module name because we need to tell webpack to bundle this module instead of the default one. So your example won't work.

And of course, TypeScript is the only compiler that [has a different API](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#a-simple-transform-function). Buble, Sucrase, and Babel have exactly the same signature.

Do you have any ideas?

docs/Configuration.md
sapegin: How will it work for changing default compiler options?

docs/Configuration.md
sapegin: The only way I see now is having two options: `compilerModule` and `compileExample` (names could be different).

```js
{
  compilerModule: 'sucrase',
  compileExample: (compiler, code) => compiler.transform(code, { /* options */ })
}
```

I guess this also answers the question about changing default compiler options.

sapegin: This approach seems to work, I've updated the branch!

docs/Configuration.md
okonet: Same, you’d need to import the default compiler function from styleguidist package and call it here I’d say.

okonet (approved):

docs/Configuration.md
sapegin: This wouldn't work, that's why we have to pass the compiler function from the module that is bundled by webpack. Styleguidist isn't, so we can't use any dependencies in `compileExample` function. I probably should add a note about that.

jchenevarin:
Hello guys, I've been looking for these commits in one of the releases to try that out but couldn't find them. Could you confirm me that with @11.2.0 it's still not possible to have typescript example in markdown and in that case could you point me towards the version containing these commits? thanks a lot

sapegin:
v12 should have this: https://github.com/styleguidist/react-styleguidist/releases/tag/v12.0.0-alpha9.9

a-maggi:
Hey @sapegin, I'm getting this following error with that alpha version. Do I need to set something additional?

`export 'mdx' (imported as 'mdx') was not found in '@mdx-js/react' (possible exports: MDXContext, MDXProvider, useMDXComponents, withMDXComponents)`

Issue #1703, opened 2020-10-23
Support for Webpack v5


jasonwilliams:
Is there any plan to add support for Webpack v5?
I get this deprecation warning for `styleguidist server`

```
(node:2280) [DEP_WEBPACK_COMPILATION_NORMAL_MODULE_LOADER_HOOK] DeprecationWarning: Compilation.hooks.normalModuleLoader was moved to NormalModule.getCompilationHooks(compilation).loader
    at getNormalModuleLoader ([dir]/node_modules/webpack/lib/Compilation.js:297:39)
    at Object.get normalModuleLoader [as normalModuleLoader] ([dir]/node_modules/webpack/lib/Compilation.js:603:12)
    at StyleguidistOptionsPlugin.plugin ([dir]/node_modules/react-styleguidist/lib/scripts/utils/StyleguidistOptionsPlugin.js:30:18)
    at Hook.eval [as call] (eval at create ([dir]/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:100:1)
    at Hook.CALL_DELEGATE [as _call] ([dir]/node_modules/webpack/node_modules/tapable/lib/Hook.js:14:14)
    at Compiler.newCompilation ([dir]/node_modules/webpack/lib/Compiler.js:919:26)
    at [dir]/node_modules/webpack/lib/Compiler.js:960:29
    at Hook.eval [as callAsync] (eval at create ([dir]/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] ([dir]/node_modules/webpack/node_modules/tapable/lib/Hook.js:18:14)
    at Compiler.compile ([dir]/node_modules/webpack/lib/Compiler.js:955:28)
    at [dir]/node_modules/webpack/lib/Watching.js:113:19
    at Hook.eval [as callAsync] (eval at create ([dir]/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:15:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] ([dir]/node_modules/webpack/node_modules/tapable/lib/Hook.js:18:14)
    at run ([dir]/node_modules/webpack/lib/Watching.js:67:33)
    at Watching._go ([dir]/node_modules/webpack/lib/Watching.js:124:4)
[...]

PR #1707, opened 2020-10-30
support Webpack 5

jasonwilliams:
The changes below still support Webpack v4. Styleguidist itself still uses webpack 4 but this adds support for projects using Webpack 5. Fixes https://github.com/styleguidist/react-styleguidist/issues/1703

From my testing it does work fine for both versions, there are some upstream issues but i don't see those blocking this from going in.

- _Inline loaders and ! prefixes should not be used as they are non-standard. They may be use by loader generated code._
https://webpack.js.org/configuration/module/#ruleenforce
- Adds support for both Webpack 4 and 5 in [`StyleguidistOptionsPlugin`](https://github.com/styleguidist/react-styleguidist/pull/1707/commits/1311b1f63dd0caaa6f22ac19561dcf3622ea5d25#diff-67ee9cb393b4198c6d229d75b25f956005e9579c9a85f7ba2cb049de8e511746)
- Updates dependencies in Webpack example (easier to test against)
- Because automatic polyfilling is switched off in 5, `assert` was brought in as a dependency (it's used by Doctrine, see below)
- Some types needed changing

## Upstream issues
Both issues below are tied to https://github.com/facebook/create-react-app/pull/7929

- `Process is not defined`  - The page still builds but this error will show in the console. I can't seem to polyfill this if anyone else can that would be great, then I think we're done.
- `TypeError: message.split is not a function` - you may not get this error, but if you do its related to https://github.com/facebook/create-react-app/pull/7929. The quick fix is to go into `node_modules/react-dev-utils/formatWebpackMessages.js:19` and add the code from https://github.com/facebook/create-react-app/pull/7929#issuecomment-718261802

### Not needed for this but nice to have.
- Doctrine should be replaced with another JSDoc parser. Doctrine is end of life and no longer supported. It causes problems with Webpack 5 because it pulls in assert which WP5 does not polyfill. For now we can fix it by adding those polyfills (assert) but a more stable solution should be found. Issue raised: https://github.com/styleguidist/react-styleguidist/issues/1708 

## How to test

- Checkout this branch
- Change webpack to version 5 in the package.json (this is so that during runtime the correct version of webpack is used, otherwise because of yarn link it will end up using styleguidists webpack version and not the example's version)
- yarn link this project into `examples/webpack`
- staying at the top level run `rm -rf lib && npm install && npm run-script compile`
- Go into `examples/webpack`  and run `yarn styleguide`


examples/webpack/package.json
```diff
@@ -1,43 +1,44 @@
...
-    "styleguide:build": "styleguidist build"
-  },
-  "devDependencies": {
-    "@babel/core": "^7.3.3",
-    "@babel/plugin-proposal-class-properties": "^7.3.3",
-    "@babel/preset-env": "^7.3.1",
-    "@babel/preset-react": "^7.0.0-beta.56",
-    "babel-loader": "^8.0.0-beta.4"
-  }
+	"name": "react-styleguidist-example-webpack",
```
sapegin, line 2: Please revert whitespace changes in JSON files.

src/scripts/utils/StyleguidistOptionsPlugin.ts
```diff
@@ -14,15 +15,33 @@ export default class StyleguidistOptionsPlugin implements WebpackPluginInstance
...
 
-	private plugin = (compil: compilation.Compilation) => {
-		// FIXME: Find out how to type it properly
+	/**
+	 *
+	 * @param compil webpack 4 `compilation.Compilation`, webpack 5 `Compilation`
+	 */
+	private plugin = (compil: any) => {
+		// Webpack 5
+		if ('NormalModule' in webpack) {
```
sapegin, line 36: I think we can do something like this here (assuming webpack 5 is a dev dependency): 

```ts
const webpack5compil = compil as Compilation;
```

so we have at least some type checking.

sapegin (commented):
Thanks for the pull request!

Would it make sense to stop using create-react-app’s scripts? I doubt that supporting multiple versions of webpack is their goal, and we may end up with a dependency that only works with webpack 4 or 5.

src/scripts/utils/StyleguidistOptionsPlugin.ts
jasonwilliams: I don't think that's possible as `Compilation` won't exist because we're not using webpack 5 here.

jasonwilliams:
> Thanks for the pull request!
> 
> Would it make sense to stop using create-react-app’s scripts? I doubt that supporting multiple versions of webpack is their goal, and we may end up with a dependency that only works with webpack 4 or 5.

I don't know what half those scripts do so i can't really answer that, but `formatWebpackMessages` could be brought into this app as it's just one function. That would only unblock one of the issues though

src/scripts/utils/StyleguidistOptionsPlugin.ts
sapegin: That's why:

> assuming webpack 5 is a dev dependency

jasonwilliams: What i've done is added the type back in if we're on 4 (as that's the version thats installed)

sapegin:
> There is plan for react-dev-util to support webpack 5 here facebook/create-react-app#9994 i don't know if they will drop 4 or support both 4 and 5.

They bundle webpack so there's no reason for them to support multiple versions.

jasonwilliams:
@sapegin I’m not sure what you want me to do about that. Replacing the functionality of React-dev-utils is far out of scope for me as I don’t know what it does and I doubt I could rebuild it all in this repo.

How they add functionality to 5 I don’t know if that means 4 will no longer work, they bundle webpack but that doesn’t neccesarily mean it can’t integrate with 4, I suppose we just wait and see for that.

In terms of this PR, everything needed here is done, the only remaining problems are not here so this could go ahead in theory.

jasonwilliams:
@sapegin is there a reason you need to keep supporting 4, other plugins and deps are dropping 4 going forward. Webpack themselves no longer support 4. So my guess is any user of this will just lock to a version if they can’t upgrade yet.

sapegin:
There are three things we use from react-dev-utils: 

* [`webpackHotDevClient`](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/webpackHotDevClient.js)
* [`formatWebpackMessages`](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/formatWebpackMessages.js)
* [`WebpackDevServerUtils`](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/WebpackDevServerUtils.js) (from wich we only use `prepareUrls` function)
* [`openBrowser`](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openBrowser.js)

And we could probably use more. Maintaining them all inside Styleguidist isn't an option, there's a lot of tricky code, that will break, and someone will have to fix it. Better to rely on a popular project for that.

> I’m not sure what you want me to do about that

This is up to you, if this pull request isn't breaking anything for webpack 4 right now, we can merge it as is, and it will be a huge step forward in supporting webpack 5.

> is there a reason you need to keep supporting 4? Other plugins and deps are dropping 4 going forward. Webpack themselves no longer support 4. So my guess is any user of this will just lock to a version if they can’t upgrade yet.

The experience of working with real projects is my reason. It will take years to migrate to webpack 5. And migrations, when you have to upgrade multiple tools at the same time because they are compatible only with specific versions, are extremely painful.

jasonwilliams:
> This is up to you, if this pull request isn't breaking anything for webpack 4 right now, we can merge it as is, and it will be a huge step forward in supporting webpack 5.

It's good to merge, all we can do at this point is see what react-dev-utils does.

sapegin:
Cool, then let's merge as soon as it's green. I guess we'll need to ignore coverage on one of the webpack version branches.

https://codewithhugo.com/jest-exclude-coverage/

jasonwilliams:
@sapegin its a long build, is it hanging?

sapegin:
It is but it's very very slow now 😢 

Merging, thanks for the help!

PR #1710, opened 2020-11-04
[CVE-2020-7753] Update remark to 13.0.0

### LINKED ISSUE #1706: [CVE-2020-7753] Update remark to 13.0.0
I was tasked with resolving [a vulnerability coming from `trim` package](https://snyk.io/vuln/SNYK-JS-TRIM-1017038). Dependency chain is as follows: `trim` <- `remark-parse` <- `remark` <- `react-styleguidist`<- our repo.

`remark-parse` in version 9.0.0 [removes the `trim` completely](https://github.com/remarkjs/remark/commit/6b42465526bb15f70d98a0ea0daccea01ffb8004#diff-0828111e0233dec7891d8f0d6a9f496957215446a0935a1c71e22cc6c6c1760c), and to have `remark-parse` in 9.0.0, we need to bump the `remark` to 13.0.0.

Unfortunately, `react-styleguidist` fixes the `remark` to ^11.0.1.

Because of the fact that `trim`'s vulnerable code won't be used anyway in modern browsers, I've managed to reduce the vulnerability from **high** to **low**, but it would be better if we could get rid of `trim` altogether - hence I'm opening an issue on your side.

poteirard:
Fixes #1706


sapegin:
Thanks for the pull request! Please have a look at the Travis CI results.

poteirard:
Fixed the emphasis tags. I asked in the remark repo how to have the auto links https://github.com/remarkjs/remark/discussions/560

src/loaders/utils/chunkify.ts
```diff
@@ -64,13 +64,14 @@ export default function chunkify(
 	}
 
 	const rendered = remark()
+		.use({ settings: { emphasis: '_' } })
```
sapegin, line 67: Let's keep the default value, it doesn't really matter for us.

sapegin (commented):
The autolinking I think is actually undesirable and the new behavior is better. I found that even remove this behavior manually:

https://github.com/styleguidist/react-styleguidist/blob/master/src/loaders/utils/noAutoLinkRemarkPlugin.ts

So I think we probably can remove this file now.

src/loaders/utils/chunkify.ts
poteirard: Removed and snapshots updated

sapegin:
Could you try to remove `noAutoLinkRemarkPlugin.ts`?

poteirard:
Removed noAutoLinkRemarkPlugin

sapegin:
Thanks!

PR #1711, opened 2020-11-10
WIP:  Replace doctrine with another JSDoc parser

### LINKED ISSUE #1708: Replace doctrine with another JSDoc parser
As part of the [Webpack 5 investigation](https://github.com/styleguidist/react-styleguidist/pull/1707) it was discovered doctrine uses some node modules which WP5 no longer polyfills by default. [Doctrine](https://github.com/eslint/doctrine) itself is end of life and no longer supported. It would be good for styleguidist to move off it permanently and replace it with another solution.

poteirard:
Replace doctrine as it's no longer maintained.

- [comment-parser](https://github.com/syavorsky/comment-parser/issues) for searching tags
- [jsdoctypeparser](https://github.com/jsdoctypeparser/jsdoctypeparser) for parsing the types (if needed)

TODO:
- Fix merge with description
- The react-docgen types are incompatible with the new implementation. So I'm not sure if <ArgumentRenderer/> (line 43) will work as expected if I use the stringifying from the new library. 

Fixes #1708


src/loaders/utils/parseComment.ts
```diff
@@ -0,0 +1,55 @@
+import commentParse, { Comment } from 'comment-parser';
```
sapegin, line 1: I'm not really sure what this module does: looks like it's changing one format to another but I don't see why it's useful. The idea was to encapsulate all the comment parsing complexity (including normalization) into a single module and return a useful, simple, and typed object. Right now it creates even more complexity.

Please have a look at the example in [my comment](https://github.com/styleguidist/react-styleguidist/issues/1708#issuecomment-723224686).

poteirard: when you say `encapsulate all the comment parsing complexity` you mean also remove the react-docgen part in getProps? Sorry, I think I misunderstood the code.

sapegin: _Maybe_, let's solve problems one by one.

poteirard:
Closing as I don't fully understand the code

Issue #1702, opened 2020-10-19
Links don't work anymore for `pagePerSection`


alexcanessa:
**Current behavior**

Since `11.0.9` (I suspect [because of this PR](https://github.com/styleguidist/react-styleguidist/pull/1627)), when clicking on any component (with `pagePerSection: true`) the routing goes to `#/-?id=ComponentName`.

**To reproduce**

1. Configure styleguidist to have `pagePerSection: true`
2. Click on any component.
3. The link will resolve to a `Page Not Found`.

**Expected behavior**

Routing to the actual component page, showing the component.

mitsuruog:
@alexcanessa Would you create a code sandbox?

https://codesandbox.io/s/github/mitsuruog/react-styleguidist-issue-template

alexcanessa:
@mitsuruog sure:

https://codesandbox.io/s/small-hill-fg8yr?file=/styleguide.config.js

Dagniele:
I confirm to have the same problem

mitsuruog:
This seems to be due to the fact that `pagePerSection: true` is set but `sections` is not in the config. 
A simple workaround is to move your `components` into the `sections` for now.

```js
module.exports = {
  webpackConfig: require("./webpack.config"),
  // This is what breaks the URL:
  pagePerSection: true,
  sections: [
    {
      name: "<section name>",
      components: "src/components/**/*.{js,jsx}"
    }
  ]
};
```

Dagniele:
It works with the solution proposed by @mitsuruog 
Thanks

mitsuruog:
@sapegin 
If `pagePerSection: true` is set, I think we need to make an error while checking the configuration file scheme when `sections` are required.
What do you think?

sapegin:
Not sure why it should be an error, seems like a valid configuration. What am I missing?

(The problem is that this feature wasn't well thought and too brittle, that why we constantly have bugs, and when someone fixes one bug they usually create several others.)

mitsuruog:
From what I've seen in the implementation, it doesn't seem to work properly without `sections`.
if the `name` in the `sections` isn't passed, the `-` is used as a section name.
that's why the URL looks like`#/-?id=ComponentName`. The part marked with `-` is where the section name is originally set.

- https://github.com/styleguidist/react-styleguidist/blob/master/src/client/utils/processSections.ts#L19

Since the implementation looked like that, I thought that it was designed from the beginning to always have sections.

[...]

PR #1715, opened 2020-11-16
Fix Links don't work anymore for `pagePerSection`

mitsuruog:
Resolved:

- Links don't work anymore for `pagePerSection` #1702

### DIFF
--- a/src/client/utils/processSections.ts
+++ b/src/client/utils/processSections.ts
@@ -23,9 +23,9 @@ export default function processSections(
 			getUrl({
 				name: section.name,
 				slug: section.slug,
-				anchor: !options.useRouterLinks,
-				hashPath: options.useRouterLinks ? hashPath : false,
-				useSlugAsIdParam: options.useRouterLinks ? useHashId : false,
+				anchor: !useRouterLinks,
+				hashPath: useRouterLinks ? hashPath : false,
+				useSlugAsIdParam: useRouterLinks ? useHashId : false,
 			});
 
 		return {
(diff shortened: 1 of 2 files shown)


src/client/utils/processSections.ts
```diff
@@ -23,9 +23,9 @@ export default function processSections(
 			getUrl({
 				name: section.name,
 				slug: section.slug,
-				anchor: !useRouterLinks,
-				hashPath: useRouterLinks ? hashPath : false,
-				useSlugAsIdParam: useRouterLinks ? useHashId : false,
+				anchor: !options.useRouterLinks,
```
sapegin, line 26: What's the difference here?

sapegin (approved):
Thanks! I only have one question here ;-)

src/client/utils/processSections.ts
mitsuruog: ah... sorry, I didn't put anything details on the PR comment yet.

If you don't have a `sections` in your config, the `getUrl()` doesn't need to create a URL for the section format, so I tweaked it to work as the same as to create a regular URL format.

the main difference is the part of here:

```diff
const options = {
-  useRouterLinks,
+  useRouterLinks: Boolean(useRouterLinks && section.name),
}
```

sapegin: This part I see, I'm asking about this particular charge from `options.useRouterLinks` to `useRouterLinks` because it looks confusing to me.

mitsuruog: hmm... what should I do...?
I'm not confused because it was written by myself, but I understand the opinion that it is confusing to others.

sapegin: How should I know? ;-/ I have no idea why this change was made.

mitsuruog: lol. I've asked you a weird thing. sorry about that.
and then I've checked my code and noticed that I don't need to change this part anything.
so I'll revert this part. 

Thanks a lot! you have always led me to the right place.

PR #1717, opened 2020-11-18
Fix Para components should use `theme.fontSize.text`

mitsuruog:
Resolved:

- #1661 

This image shows that I've changed the `fontSize.text` in the config.

React-Styleguidist-Style-Guide

config

```
module.exports = {
  theme: {
    fontSize: {
      text: 36,
    },
  },
};
```


sapegin:
Thanks!

sapegin:
Looks like it's breaking the props table, we shouldn't use this large font there:

image


It's even visible on your own screenshot but I've missed it ;-/

mitsuruog:
> Looks like it's breaking the props table, we shouldn't use this large font there:

@sapegin I'll work on that later.

Issue #1705, opened 2020-10-27
Dev server crashes with react-scripts v4


varoot:
**Current behavior**

When running `styleguidist server` with `react-scripts` 4.0.0 and open in the browser, the server crashes with "Error: Already listening" message.

**To reproduce**
1. Create a new project with latest CRA: `yarn create react-app test`
2. Install Styleguidist `yarn add -D react-styleguidist`
3. Run dev server and open the browser `yarn styleguidist server --open`

Demo repository:
https://github.com/varoot/styleguidist-example

**Expected behavior**

The dev server should run without crashing. User should be able to see the styleguide.

rpeach:
Seems to be caused by the fast refresh changes in `react-scripts/config/webpack.config.js`

You can temporarily work around the error by setting an environment variable `FAST_REFRESH=false`

Something like this
```
cross-env FAST_REFRESH=false styleguidist server
```

krsnaa:
Alternatively, just include it in your package.json:

`"scripts": {
   "styleguide": "FAST_REFRESH=false styleguidist server"
}
`

aniketkolekar-ms:
**FAST_REFRESH=false** : Does this mean the app will not reload on changes?

aniketkolekar-ms:
Temporary Fix:

1. Install cross-env : `npm install --save-dev cross-env`
2. Update script in package.json : `"scripts": { "styleguide": "cross-env FAST_REFRESH=false styleguidist server" }`
3. Run : `npm run styleguide`

This worked for me. Auto-refresh working fine.

poteirard:
Maybe this is related https://github.com/facebook/create-react-app/pull/9872 ?

lyz810:
I have the same problem,when upgrade to 11.1.2,and 11.1.1 has the same problem

I find react-dev-utils in package.json upgrade from ^9.1.0 to ^11.0.0


react-dev-utils@9.1.0 use sockjs-client, an react-dev-utils@11.0.0 use browser WebSocket directly
react-dev-utils@9.1.0:
https://github.com/facebook/create-react-app/blob/v3.2.0/packages/react-dev-utils/webpackHotDevClient.js#L61
react-dev-utils@11.0.0:
https://github.com/facebook/create-react-app/blob/v4.0.0/packages/react-dev-utils/webpackHotDevClient.js#L60

In Chrome network panel, the requests are different .
react-styleguidist@11.1.0(react-dev-utils@9.1.0):
[...]

PR #1718, opened 2020-11-22
fix devServer crash

### LINKED ISSUE #1714: Dev Server crash when upgrade to 11.1.2
**Current behavior**

run script `styleguidist server`, after reload page, the dev server is crash

**To reproduce**
1.upgrade to 11.1.2 or 11.1.1, run styleguidist server and open styleguidist in chrome
2.Open chrome console panel
3. type `new WebSocket('ws://localhost:6060/sockjs-node')` and press enter
4. repeat step3 twice 
5. server crash

I found the reason: 
https://github.com/styleguidist/react-styleguidist/issues/1705#issuecomment-726536041

I think this should be config in https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/create-server.ts#L12, no in user webpack.config.js

**Expected behavior**
server don't crash

malcolm-kee:
resolves #1714 #1705 

- Fix based on advice from @lyz810
- Tested locally the issue is fixed (able to reproduce the bug by reverting the fix)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3835,9 +3835,9 @@
       }
     },
     "@types/body-parser": {
-      "version": "1.17.1",
-      "resolved": "https://registry.npmjs.org/@types/body-parser/-/body-parser-1.17.1.tgz",
-      "integrity": "sha512-RoX2EZjMiFMjZh9lmYrwgoP9RTpAjSHiJxdp4oidAQVO02T7HER3xj9UKue5534ULWeqVEkujhWcyvUce+d68w==",
+      "version": "1.19.0",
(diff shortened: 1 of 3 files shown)


malcolm-kee:
@sapegin would you mind to take a look?

sapegin:
Thanks!

sapegin:
Looks like it's breaking the props table, we shouldn't use this large font there:

image

sapegin:
Looks like it's now printing a warning on start but the dev server isn't crashing anymore.

```
⚠ ｢wds｣: transportMode is an experimental option, meaning its usage could potentially change without warning
```

PR #1719, opened 2020-11-26
chore: Add GitHub workflow files

### LINKED ISSUE #1713: Migrate from Travis CI to GitHub Actions
Travis CI takes hours to complete each build now, which makes it difficult for maintainers and contributors to get quick feedback on their pull requests and release new versions.

We have quite complex config but I guess we can migrate each stage into its own GitHub Workflow:

https://github.com/styleguidist/react-styleguidist/blob/master/.travis.yml

Consider starting with [`npx mrm ci`](https://mrm.js.org/docs/mrm-task-ci) and [`npx mrm semantic-release`](https://mrm.js.org/docs/mrm-task-semantic-release), since they already create GitHub Actions.

balazsmatepetro:
Closes #1713


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,35 @@
...
+      - master
+
+jobs:
+  check:
+    name: Check
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout project
```
sapegin, line 18: Let's remove all these names: they are mostly redundant and don't add any useful information. The only useful ones are with parameters, like `Use Node.js ${{ matrix.node-version }}`.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,35 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  check:
+    name: Check
```
sapegin, line 13: Let's remove all these names: if there's one job they will repeat the workflow name, which isn't useful, or "Check" which is even less helpful.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: install dependencies
+        run: npm ci
+
+      - name: coverage
+        run: npm run test:coverage -- --runInBand
+
+      - name: Publish result
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
sapegin, line 33: This is only needed for private repositories, and we don't have this value anyway.

```suggestion
```

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,35 @@
...
+          fail_ci_if_error: true
```
sapegin, line 34: I'd probably keep the default `false` here. What do you think?

```suggestion
```

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: install dependencies
+        run: npm ci
+
+      - name: lint
```
sapegin, line 35: No need to run lint, type check, and especially integrations (they take a lot of time), in every Node.js version: in the current setup, we run them only once. Looks like we need a separate workflow for them.

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Checkout project
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+
+      - run: npm ci
```
sapegin, line 23:
```suggestion
      - run: npm ci

```

Readme.md
```diff
@@ -3,8 +3,7 @@
 <img src="https://d3vv6lp55qjaqc.cloudfront.net/items/061f0A2n1B0H3p0T1p1f/react-styleguidist-logo.png" alt="React Styleguidist" width="400">
 
 **Isolated React component development environment with a living style guide**
```
sapegin, line 5:
```suggestion
**Isolated React component development environment with a living style guide**

```

tsconfig.json
```diff
@@ -15,6 +15,7 @@
 			"rsg-components/*": ["src/client/rsg-components/*"]
 		}
 	},
-	"include": ["src"],
+  "files": ["dangerfile.ts"],
```
sapegin, line 18: Let's fix the indentation in this file: it should be two spaces, not sure why we have tabs here.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,58 @@
+name: CI
```
sapegin, line 1: I don't see where we run tests here.

https://github.com/styleguidist/react-styleguidist/blob/9c2ca5b33a2d6b5008154521ed7a41f52616bbd1/.travis.yml#L12-L13

sapegin (changes requested):
Thanks! 🦄 

We'll also need to remove the [Travis config file](https://github.com/styleguidist/react-styleguidist/blob/master/.travis.yml) once we're sure this thing is working.

.github/workflows/coverage.yml
balazsmatepetro: 75ea2be4b5f5b9994de9eaa3e26167be85925d21

.github/workflows/coverage.yml
balazsmatepetro: 75ea2be4b5f5b9994de9eaa3e26167be85925d21

.github/workflows/coverage.yml
balazsmatepetro: 6f1ea5a99d236e49ffb989c62795538fa6f88ba4

.github/workflows/coverage.yml
balazsmatepetro: 7aa379c1924110995ba31c92e78de8d6a7c5bbe8

Readme.md
balazsmatepetro: c712f0e020329ca052679cfcaee6866fa883326b

.github/workflows/node.js.yml
balazsmatepetro: a6070739a11af9b7d1a563181e52ce8a764f82eb

.github/workflows/node.js.yml
balazsmatepetro: 1ecbc86cb99f0f809d2fbba5a39e3f2054b12fea

tsconfig.json
balazsmatepetro: 20df7421c97184a3d5fec6ccb6b5c8afda9449ec

.github/workflows/release.yml
balazsmatepetro: Not really related to this one, but the new line was added when I did this change:

2883cbd85a84c6c30273817b3d02f287b87ced12

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - run: npm ci
+      - run: npm run test:jest -- --runInBand
+      - run: |
+          # Compile JS
+          npm run compile
+          # Build all examples
+          npm run build:basic
+          npm run build:customised
+          npm run build:sections
```
sapegin, line 74: We don't need this here.

```suggestion
```

.github/workflows/release.yml
sapegin: I believe accepting changes would make the whole process faster for both of us ;-)

sapegin (approved):
Approving with one little change.

tsconfig.json
okonet: This can and should be handled by lint-staged and prettier. Just saying ;)

okonet (approved):

sapegin:
Looks like it works!

Issue #1723, opened 2020-12-02
Type description tooltip isn't spaced properly in Firefox


alextewpin:
**Current behavior**

Type description tooltip isn't spaced properly in Firefox.
<img width="978" alt="Screenshot 2020-12-02 at 17 03 22" src="https://user-images.githubusercontent.com/1542068/100883093-fb55c700-34c0-11eb-9a12-1add23cac246.png">

**Expected behavior**

Same table in Chrome:
<img width="975" alt="Screenshot 2020-12-02 at 17 07 12" src="https://user-images.githubusercontent.com/1542068/100883036-e8db8d80-34c0-11eb-8c19-5fc4257e5d8c.png">

mitsuruog:
Which version of firefox actually are you using?
I could not reproduce it. I might have missed something though.

<img width="808" alt="スクリーンショット 2020-12-03 20 11 20" src="https://user-images.githubusercontent.com/1703219/101004400-e3a33f00-35a3-11eb-8f90-00c30ff09b2f.png">

react-styleguidist: 11.1.4
Firefox: 83.0 (Mac: Big Sur)

alextewpin:
@mitsuruog same as you, but on Catalina. It seems like not all tables are affected. Here is live example: https://whale.cargoonline.ru/#!/DateRangePicker

mitsuruog:
Thanks. I've seen it now.
The only firefox seems to have a problem with inline-flex, but I don't find the reason so far.

mitsuruog:
I've sent a PR.

PR #1724, opened 2020-12-03
Fix Type description tooltip isn't spaced properly in Firefox

mitsuruog:
Resolved:

- #1723 

The steps to reproduce:
1. Use Firefox and TypeScript project
2. Run the styleguide and open the props & methods table
3. Open the dev-tool and modify the type name to something very long. (for example: `function` => `functionsdfakdshjaksdjfka`)
4. Reproduce 🎉 

Before the change:
スクリーンショット 2020-12-03 21 50 56

After the change:
スクリーンショット 2020-12-03 21 51 25


sapegin:
Thanks!

Issue #1721, opened 2020-11-30
env is hardcoded to "production" which causes React v17 to fail building properly 


jasonwilliams:
**Current behavior**

I think there's a mismatch of environment variables being read by babel (JSX config) and Styleguidist, causing some issues deeper down.

(I've noticed this was raised before but was closed)
- build.js sets the env to be `production` here: https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/build.ts#L9
- This check is redundant because it will always be true, as env is hardcoded from the call above https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/make-webpack-config.ts#L31
- Again, env is hardcoded to production and then passed through here https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/make-webpack-config.ts#L131

So the issue is. My webpack config has `mode` set to "production" (because mode comes from `env`) and anything else that uses NODE_ENV will see `developement`.

I ended up with this error:
`ReferenceError: _jsx is not defined`.

I don't know why its hardcoded to production regardless of what you set, but it seems to be the cause of this bug.

**Fix**
https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/make-webpack-config.ts#L29 should be 

```js
env = process.env.NODE_ENV = process.env.NODE_ENV || env;
```

sapegin:
I guess the idea was to prevent the user environment to affect Styleguidist build. But I guess we shouldn't overwrite the `process.env.NODE_ENV` because it might lead to a discrepancy with other parts of the project, which is the issue you have.

I guess your solution would be good enough, so feel free to send a pull request.

Edu93Jer:
Hi @sapegin is anyone still working on the case, I don't know if @jasonwilliams send the PR? If no, could you please explain to me a little more about the project, I'm a newbie and would like to contribute...

jasonwilliams:
hi @Edu93Jer i have not done a PR for this, feel free to have a go. Please see potential fix in my comment https://github.com/styleguidist/react-styleguidist/issues/1721#issue-753531032 for what needs to change

Edu93Jer:
@jasonwilliams Ok, I saw at your comment, but I didn't understand how to reproduce the error, could you perhaps explain me a little more?
Do I have to create a file .env? and set the NODE_ENV variable to production?

joelchan24:
Hi Im new in this project, how can I see that error??, Im sorry Im newbie in this

[...]

PR #1727, opened 2020-12-08
Env is hardcoded to "production" which causes React v17 to fail building properly

Edu93Jer:
closes issue #1721

Fix issue #1721 now the env variable is not hardcoded to 'production'

### DIFF
--- a/src/scripts/build.ts
+++ b/src/scripts/build.ts
@@ -6,7 +6,7 @@ export default function build(
 	config: Rsg.SanitizedStyleguidistConfig,
 	callback: (err: Error, stats: webpack.Stats) => void
 ) {
-	return webpack(makeWebpackConfig(config, 'production'), (err, stats) => {
+	return webpack(makeWebpackConfig(config, 'development'), (err, stats) => {
 		// require('fs').writeFileSync('stats.json', JSON.stringify(stats.toJson()));
 		callback(err, stats);
 	});
--- a/src/scripts/make-webpack-config.ts
+++ b/src/scripts/make-webpack-config.ts
@@ -24,8 +24,8 @@ interface AliasedConfiguration extends Configuration {
 
 type Mode = Configuration['mode'];
 
-export default function (config: Rsg.SanitizedStyleguidistConfig, env: Mode): Configuration {
-	process.env.NODE_ENV = (process.env.NODE_ENV as Mode) || env;
+export default function (config: Rsg.SanitizedStyleguidistConfig, envRaw: Mode): Configuration {
+	const env = (process.env.NODE_ENV as Mode) || envRaw;
 
 	const isProd = env === 'production';
 
--- a/src/scripts/make-webpack-config.ts
+++ b/src/scripts/make-webpack-config.ts
@@ -58,7 +58,7 @@ export default function (config: Rsg.SanitizedStyleguidistConfig, env: EnvType):
 			new StyleguidistOptionsPlugin(config),
 			new MiniHtmlWebpackPlugin(htmlPluginOptions),
 			new webpack.DefinePlugin({
-				'process.env.NODE_ENV': JSON.stringify(env),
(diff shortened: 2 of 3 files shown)


src/scripts/make-webpack-config.ts
```diff
@@ -22,11 +22,10 @@ interface AliasedConfiguration extends Configuration {
 	resolve: Resolve & { alias: Record<string, string> };
 }
 
-export default function (
-	config: Rsg.SanitizedStyleguidistConfig,
-	env: 'development' | 'production' | 'none'
-): Configuration {
-	process.env.NODE_ENV = process.env.NODE_ENV || env;
+type EnvType = 'development' | 'production' | 'none';
```
sapegin, line 25: This type is available in webpack:

```suggestion
type Mode = Configuration['mode'];
```

https://github.com/styleguidist/react-styleguidist/blob/04d900d7271761c257eacf57bb5e7279ded3525a/src/scripts/make-webpack-config.ts#L16

src/scripts/make-webpack-config.ts
```diff
@@ -22,11 +22,10 @@ interface AliasedConfiguration extends Configuration {
...
+
+export default function (config: Rsg.SanitizedStyleguidistConfig, env: EnvType): Configuration {
+	env = (process.env.NODE_ENV as EnvType) || env;
```
sapegin, line 28: Please don't overwrite function parameters.

sapegin (changes requested):
Thanks for the pull request! Please have a look at the failing CI.

Edu93Jer:
Hi @sapegin how are you, first of all merry x-mas, just asking how is the review of the PR going, I made the changes and now the check passes correctly.

src/scripts/build.ts
```diff
@@ -6,7 +6,7 @@ export default function build(
 	config: Rsg.SanitizedStyleguidistConfig,
 	callback: (err: Error, stats: webpack.Stats) => void
 ) {
-	return webpack(makeWebpackConfig(config, 'production'), (err, stats) => {
+	return webpack(makeWebpackConfig(config, 'development'), (err, stats) => {
```
sapegin, line 9: I'm not sure about this change now: this will increase the bundle size for every generated style guide, which seems unnecessary. Would it make more sense to for `NODE_ENV` for override `env` in `make-webpack-config.ts` when it's provided?

Edu93Jer: Sorry for the late replay @sapegin I was a little busy today, so isn't that the proposal @jasonwilliams did `env = process.env.NODE_ENV || env` of course now would be something like this `env = (process.env.NODE_ENV as Mode) || env` or what do you mean with override `env`?

src/scripts/make-webpack-config.ts
```diff
@@ -22,11 +22,10 @@ interface AliasedConfiguration extends Configuration {
...
 
-export default function (
-	config: Rsg.SanitizedStyleguidistConfig,
-	env: 'development' | 'production' | 'none'
-): Configuration {
-	process.env.NODE_ENV = process.env.NODE_ENV || env;
+type Mode = Configuration['mode'];
+
+export default function (config: Rsg.SanitizedStyleguidistConfig, env: Mode): Configuration {
+	process.env.NODE_ENV = (process.env.NODE_ENV as Mode) || env;
```
sapegin, line 28: I think it should be something like this:

```suggestion
export default function (config: Rsg.SanitizedStyleguidistConfig, envRaw: Mode): Configuration {
	const env = (process.env.NODE_ENV as Mode) || envRaw;
```

src/scripts/build.ts
jasonwilliams: It should be ok to stick with “production” here, so long as it passes through and it’s overridable. So I think you can leave this bit as it is

src/scripts/make-webpack-config.ts
Edu93Jer: We can try, I will do the commit from here

PR #1743, opened 2021-02-09
Support number in moduleAlias

### LINKED ISSUE #1742: No longer supports module aliases that start with a number
**Current behavior**

If any of your webpack aliases start with a number, the example fails to render. This is specifically when using es6 imports

**To reproduce**
Demo repository
https://github.com/laytong/example

Styleguide will build successfully.
Open style guide and see that example for button shows syntax error instead of the example.
Screenshot 2021-02-08 at 18 15 18

This does not fail in Styleguide v10.

Looks like evalInContext now starts variable declaration with the name of the module alias and variables cannot start with a number.

If you change the import to use require, or not start with a number it works.

**Expected behavior**

The example should render

laytong:
Detect if the proposed variable is going to start with a number and pad the front with a safe prefix

Fixes #1742

### DIFF
--- a/src/client/utils/rewriteImports.ts
+++ b/src/client/utils/rewriteImports.ts
@@ -24,12 +24,12 @@ function generate(keys: string[], dep: string, base: string, fn: string): string
 	let out = `const _${name} = ${dep};`;
 
 	if (base) {
-		out += `\nconst _${base} = _${tmp}.default || _${tmp};`;
+		out += `\nconst ${base} = _${tmp}.default || _${tmp};`;
 	}
 
 	keys.forEach((key) => {
 		obj = alias(key);
-		out += `\nconst _${obj.name} = _${tmp}.${obj.key};`;
+		out += `\nconst ${obj.name} = _${tmp}.${obj.key};`;
 	});
 
 	return out;
--- a/src/client/utils/rewriteImports.ts
+++ b/src/client/utils/rewriteImports.ts
@@ -17,29 +17,30 @@ let num: number;
 function generate(keys: string[], dep: string, base: string, fn: string): string {
 	const tmp = dep.replace(/\W/g, '_') + '$' + num++; // uniqueness
 	const name = alias(tmp).name;
+	const safePrefix = /[0-9]/.test(name.split('')[0]) ? 'safeVarPrefix' : '';
 
 	dep = `${fn}('${dep}')`;
 
 	let obj;
-	let out = `const ${name} = ${dep};`;
+	let out = `const ${safePrefix}${name} = ${dep};`;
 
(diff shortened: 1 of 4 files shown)


laytong:
I hate to be a pest, but can I get some insight into when someone might be able to take a look at this?
I'm hoping we can release this as a patch upgrade in due course as this is a blocker for something I am working on.

If there's anything else I can do to help/make this easier please let me know.

src/client/utils/rewriteImports.ts
```diff
@@ -17,29 +17,30 @@ let num: number;
 function generate(keys: string[], dep: string, base: string, fn: string): string {
 	const tmp = dep.replace(/\W/g, '_') + '$' + num++; // uniqueness
 	const name = alias(tmp).name;
+	const safePrefix = /[0-9]/.test(name.split('')[0]) ? 'safeVarPrefix' : '';
```
sapegin, line 20: Let's just prefix all imports with `_`, there's no need for a condition.

sapegin (commented):
Thanks! One minor comment.

src/client/utils/rewriteImports.ts
laytong: that works too :)

laytong:
Looks like this neither of these approaches are quite going to work. `React.createElement` still refers back to the non-underscored variable name. Do you have a recommendation for where we should handle this?

sapegin:
Could you show an example of the generated code that fails?

laytong:
> Could you show an example of the generated code that fails?

The test failure shows it pretty well.
const _Button = _button$0.default || _button$0;
React.createElement( Button, null )

You can see the transpiled imports rename with the _ prefix, but the subsequent call to react.createElement still expects the non prefixed variable.

sapegin:
Yup, it should be:

```js
const Button = _button$0.default || _button$0;
React.createElement( Button, null )
```

laytong:
Right, but that leads us back to the original issue. The pre-transpiled value might start with a number, which will fail

laytong:
I think I realised where I was going wrong. We don't need to prefix the final assignment, just the initial transpiled name of the import

laytong:
Not sure what's causing the CI failure here

laytong:
Looks like the CI failures can be attributed to https://github.com/cypress-io/cypress/issues/6184

laytong:
Looks like this is good to go 😄

laytong:
Just a gentle reminder to see if we can move forward with this update?

Thanks

sapegin (approved):
Looks good! 

There are some conflicts now, could you please have a look?

PR #1756, opened 2021-03-06
fix: generate typings directly in lib instead of lib/src

elevatebart:
in version 11.1.5 when we moved to Github Actions we added a dangerfile.ts to the tsconfig.json
This file, since it is outside the src folder, changed the reference rootFolder for typings to calculate.
This quick adjustment should fix the typings.

I also updated CI to have clearer action steps (where we can measure progress from a high level).
Cypress 3.0.2 does not seem to work well with the Linux version on the github action docker images.

### DIFF
--- a/tsconfig.types.json
+++ b/tsconfig.types.json
@@ -4,6 +4,8 @@
 		// only generate typings when using this tsconfig
 		"declaration": true
 	},
+	// ignore the dangerfile in calculation of the root folder
+	"files": [],
 	// avoid generating typings for tests and mocks
 	"exclude": [
 		"dangerfile.ts",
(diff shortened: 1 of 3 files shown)


tsconfig.types.json
```diff
@@ -1,16 +1,18 @@
...
-		"**/*.spec.tsx",
-		"**/*.spec.ts"
-	]
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    // only generate typings when using this tsconfig
+    "declaration": true
+  },
+  // ignore the dangerfile in calculation of the root folder
+  "files": [],
```
elevatebart, line 8: This is the only actual change the rest is white spaces

elevatebart: I will remove the white space changes to limit pollution of the git log

elevatebart:
@sapegin I would like your opinion on the CI changes
When you have a minute, please advise.

PR #1767, opened 2021-03-17
Update TableRenderer.tsx

FDiskas:
On small props the table looks not optimized. And for big property lists does not effect

Original:
image
image

Updated:
image
image


FDiskas:
I would like to have at least opinion on this :)

FDiskas:
Ping pong

FDiskas:
Ding dong :D

FDiskas:
As alternate solution is provide custom style config
```js
export default {
  Table: {
    cell: {
      '&:last-child': {
        width: 'auto'
      }
    },
  }
};
```

Issue #1378, opened 2019-06-03
how to highlighting  java、ruby  language in *.md files


xudihui:
first of all,react-styleguidist is amazing! thanks very much!

this problem:
Warning: Syntax highlighting for “java” isn’t supported

dysticus:
@xudihui The warning message says it all: 
`Warning: Syntax highlighting for “java” isn’t supported. Supported languages are: clike, markup, xml, html, mathml, svg, markdown, md, css, scss, javascript, js, flow, typescript, ts, jsx, tsx, graphql, json, bash, shell, diff.
`
A workaround would be to edit this file https://github.com/styleguidist/react-styleguidist/blob/master/src/loaders/utils/highlightCode.js#L3 to also include Ruby/Java from Prism, but it would be overwritten on a fresh `npm install`

@sapegin I understand why languages other then the most common used in component development are not part of the default. However, perhaps the API can change a bit so a user can decide which Prism highlighting to add to their bundle, instead of the current hardcoded list. What do you think?

sapegin:
Sure, feel free to send a pull request.

johanbook:
I am also interested in this. Is there some solution without editing `node_modules` or forking it?

PR #1784, opened 2021-04-20
feat:Added the language highlighter option to allow users to freely expand the list of hard-coded languages.

xudihui:
what a excellent Styleguidist, and it is very convenient for component testing. I love it!

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] TypeScript definition update
- [ ] Bundle size optimization
- [ ] Performance optimization
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

[issues#1378](https://github.com/styleguidist/react-styleguidist/issues/1378)
[pull#1520](https://github.com/styleguidist/react-styleguidist/pull/1520)

### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [X] Doc is updated/provided or not needed
- [X] Demo is updated/provided or not needed
- [X] TypeScript definition is updated/provided or not needed
- [X] Changelog is provided or not needed

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -78,6 +78,22 @@ module.exports = {
 }
 ```
 
+## `highlightCodeLanguages`
+
+Type: `String[]`, default: `[]`
+
+Array of languages that allow you to decide which syntax highlighter to add.
+
+By default the following languages are supported.
+
+> atom, bash, clike, css, diff, flow, graphql, html, javascript, js, json, jsx, less, markdown, markup, mathml, md, rss, scss, shell, ssml, svg, ts, tsx, typescript, webmanifest, xml
+
+```javascript
+module.exports = {
+  highlightCodeLanguages: ['java', 'ruby']
+}
+```
+
 ## `configureServer`
 
 Type: `Function`, optional
--- a/src/loaders/styleguide-loader.ts
+++ b/src/loaders/styleguide-loader.ts
@@ -13,6 +13,7 @@ import getSections from './utils/getSections';
 import filterComponentsWithExample from './utils/filterComponentsWithExample';
 import slugger from './utils/slugger';
 import resolveESModule from './utils/resolveESModule';
+import { getHighlightCodeLanguages } from './utils/highlightCode';
 import * as Rsg from '../typings';
 
 const logger = createLogger('rsg');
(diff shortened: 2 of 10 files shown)


docs/Configuration.md
```diff
@@ -78,6 +78,22 @@ module.exports = {
 }
 ```
 
+## `highlightCodeLanguages`
```
sapegin, line 81: The docs aren't very clear:

1. Unclear whether the value of this option overwrites or extends the default list. 
2. Unclear what "by default" means: are these languages _enabled_ by defaults or could be used in the config option. In any case, we need two lists here: enabled by default and available.
3. We only use one “syntax highlighter” which is Prism, the option adds highlighting of more languages for static code examples in the style guide.

Also, this name would sound better I think:

```suggestion
## `codeHighlightLanguages`
```

src/loaders/styleguide-loader.ts
```diff
@@ -45,6 +46,9 @@ export function pitch(this: Rsg.StyleguidistLoaderContext) {
 
 	const config = this._styleguidist;
 
+	// get highlight code languages. Array of languages that allow you to decide which syntax highlighter to add
+	getHighlightCodeLanguages(config.highlightCodeLanguages);
```
sapegin, line 50: The name of the function suggests that it returns a list of languages which isn't used.

Maybe:

```suggestion
	loadCodeHighlightLanguages(config.highlightCodeLanguages);
```

src/loaders/styleguide-loader.ts
```diff
@@ -45,6 +46,9 @@ export function pitch(this: Rsg.StyleguidistLoaderContext) {
 
 	const config = this._styleguidist;
 
+	// get highlight code languages. Array of languages that allow you to decide which syntax highlighter to add
```
sapegin, line 49: The comment doesn't add anything that's not in the code already.

src/typings/dependencies/prismjs-components-index.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'prismjs/components/index' {
```
sapegin, line 1: Can we use `@types/prismjs` instead?

sapegin (changes requested):
Thanks for the pull request!

PR #1844, opened 2021-06-21
Fix Component Driven Domain

vinzid:
Seems that Component-Driven have change the domain name from `.io` to `.dev`, and the original domain name is not accessible now.


sapegin:
Thanks!

PR #1845, opened 2021-06-21
Add Elementz in Showcase

el1s7:
Thanks for making styleguidist <3 !
 I've used it for documenting my latest project https://github.com/elementz-ui/elementz and would like to include it in the showcase examples if possible.

### DIFF
--- a/site/src/pages/index.js
+++ b/site/src/pages/index.js
@@ -70,12 +70,7 @@ const examples = [
 	{
 		title: 'Elementz UI',
 		href: 'https://elementz.style',
-		image: 'https://i.imgur.com/JruKLG1.png'
-	},
-	{
-		title: 'Dialog components',
-		href: 'https://dialogs.github.io/dialog-web-components/',
-		image: '/img/example2.png',
+		image: 'https://elementz.style/showcase.PNG'
 	},
 	{
 		title: 'Everydayhero Constructicon',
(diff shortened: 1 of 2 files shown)


site/src/pages/index.js
```diff
@@ -67,16 +67,16 @@ const examples = [
 		href: 'https://react-styleguidist.js.org/examples/basic/',
 		image: '/img/example1.png',
 	},
+	{
+		title: 'Elementz UI',
+		href: 'https://elementz.style',
+		image: 'https://elementz.style/showcase.PNG'
```
sapegin, line 73: Please update an existing image and make it the same size and style as the other images (it should be an actual screenshot). Don't forget to optimize an image too.

Also, no need to change the order.

sapegin (changes requested):
Thanks!

site/src/pages/index.js
el1s7: I have resized it to 590x473. I'm not sure what you mean by style and order, it is an actual screenshot, if you mean to use the Light theme then you can add this instead https://elementz.style/showcase_white.png

sapegin: Yup, light theme would better match the other screenshots. And it's also the default.

By order I mean the order or examples on the page.

el1s7: Feel free to edit it

el1s7: @sapegin

PR #1862, opened 2021-07-31
Supports extra language highlignting

### LINKED ISSUE #1857: Extra syntax highlighting
Hi,

**The problem**

I would like to highlight other languages than the subset enforced by `styleguidist`, in instance `yaml`.

I am new to js world, but I did check `src/loaders/utils/highlightCode.ts` and it sounds like `styleguidist` only provides a subset of the available `prismjs` highlighters.

I guess you do it to reduce the size of the client side code? or the loading time? I am not an expert.

**Proposed solution**

Could it be possible to provide a kind of configuration field to manually allow to request to include other languages?
So what, per project people could tune the expected list.

**Alternative solutions**

Or maybe you are anyway interested in including `yaml` by default?

What do you think?

vallsv:
This PR allows to define extra languages which can be highlighted as Markdown code block.

- It exposes a `extraLanguages` to the `styleguidist` configuration
- This config is used to import the required prismjs module
- As prismjs cache such import, the remaining code of `styleguidist` renders properly this extra language

Closes #1857

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -136,6 +136,22 @@ Defines the initial state of the example code tab:
 - `hide`: hide the tab and it can´t be toggled in the UI.
 - `expand`: expand the tab by default.
 
+## `extraLanguages`
+
+Type: `String[]`, default: `[]`
+
+Extra alias languages which can be rendered using syntax highlighter. By default, `styleguidist` only highlight web related languages.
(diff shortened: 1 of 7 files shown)


vallsv:
Soulds like it was already proposed here #1784 and nobody answer anything.

Issue #1880, opened 2021-09-02
posibility of a 11.1.8 release with less strict webpack-dev-server version dependency 


bestrivens001:
Hi, @sapegin,
### Issue Description

   I noticed that a vulnerability is introduced in **_react-styleguidist@11.1.7_**:
     Vulnerability **CVE-2020-28469** affects package **_glob-parent_** (versions:<5.1.2):  [https://snyk.io/vuln/SNYK-JS-GLOBPARENT-1016905](https://snyk.io/vuln/SNYK-JS-GLOBPARENT-1016905)
  The above vulnerable package is referenced by **_react-styleguidist@11.1.7_** via:
`react-styleguidist@11.1.7 ➔ webpack-dev-server@3.11.2 ➔ chokidar@2.1.8 ➔ glob-parent@3.1.0`

Since **_react-styleguidist@11.1.7_** ([19,010 downloads per week](https://www.npmjs.com/package/react-styleguidist/v/11.1.7?activeTab=versions)) is referenced by **27** downstream projects (e.g., [vue-styleguidist 4.41.0](https://www.npmjs.com/package/vue-styleguidist) (latest version), [vue-cli-plugin-styleguidist 4.41.0](https://www.npmjs.com/package/vue-cli-plugin-styleguidist) (latest version), [@irvingjs/styleguide 6.1.0](https://www.npmjs.com/package/@irvingjs/styleguide) (latest version), [@comicrelief/component-library 5.3.0](https://www.npmjs.com/package/@comicrelief/component-library) (latest version), [create-react-styleguide 7.0.1](https://www.npmjs.com/package/create-react-styleguide) (latest version)), the vulnerability [**CVE-2020-28469**](https://snyk.io/vuln/SNYK-JS-GLOBPARENT-1016905) can be propagated into these downstream projects and expose security threats to them via the following package dependency paths:
 (1)`@comicrelief/component-library@5.3.0 ➔ react-styleguidist@11.1.7 ➔ webpack-dev-server@3.11.2 ➔ chokidar@2.1.8 ➔ glob-parent@3.1.0`
(2)`@irvingjs/styleguide@6.1.0 ➔ react-styleguidist@11.1.7 ➔ webpack-dev-server@3.11.2 ➔ chokidar@2.1.8 ➔ glob-parent@3.1.0`
  **......**

If **_react-styleguidist@11.1.\*_** removes the vulnerable package from the above version, then its fixed version can help downstream users decrease their pain.

[...]

PR #1881, opened 2021-09-02
Fix the vulnerabiltiy issue discussed in #1880

bestrivens001:
Hi @sapegin, this PR can fix the vulnerability  discussed in https://github.com/styleguidist/react-styleguidist/issues/1880.
It can update `webpack-dev-server ^3.11.0 ➔ ^4.0.0` to remove the  vulnerability [**CVE-2020-28469**](https://snyk.io/vuln/SNYK-JS-GLOBPARENT-1016905) introduced by **glob-parent**, since **webpack-dev-server@4.0.0**(>=4.0.0-beta.0) transitively depends on **glob-parent@5.1.2** which has fixed the vulnerability [**CVE-2020-28469**](https://snyk.io/vuln/SNYK-JS-GLOBPARENT-1016905).
It would be better if **react-styleguidist** can fix this issue in its **11.1.8** release.Then this vulnerbility patch version can be automatically propagated into a large amount of downstream projects :)
Please check it.
Thanks again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "to-ast": "^1.0.0",
     "type-detect": "^4.0.8",
     "unist-util-visit": "^2.0.0",
-    "webpack-dev-server": "^3.11.0",
+    "webpack-dev-server": "^4.0.0",
     "webpack-merge": "^4.2.2"
   },
   "peerDependencies": {


sapegin (changes requested):
Changing a version number **isn't** an upgrade, there are a bunch of other things that have to be done for an actual upgrade: https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md, plus the lock file needs to be updated.

PR #1895, opened 2021-10-04
Add tests: scroll to element when clicked in sidebar, trigger filter …

milafreckled:
Add tests: scroll to element when clicked in the sidebar, trigger filter …in the table of contents, modified toggling the isolated mode tests.

Hi,

just added some integration tests with cypress (#776 ). The only test that is unpassed currently is  'changes the render after code change'. I have added 'scrolls to the component clicked in the sidebar' and 'typing in the sidebar should filter components in the table of content' tests, also corrected 'toggles isolated component mode correctly' for core and component. Feel free to review it and send me feedback, I really appreciate it as a beginner.
Have a nice day and happy coding!

### DIFF
--- a/test/cypress/integration/core_spec.js
+++ b/test/cypress/integration/core_spec.js
@@ -1,5 +1,5 @@
 describe('Styleguidist core', () => {
-	before(() => cy.visit('/'));
+	beforeEach(() => cy.visit('/'));
 
 	it('loads the page', () => {
 		cy.title().should('include', 'React Styleguidist');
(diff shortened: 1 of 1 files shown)


sapegin (commented):
Thanks for taking this!

The existing end-to-end tests are very very old and it doesn't make sense to continue the same way, this approach is already introducing practices we should avoid like querying elements by class names and HTML elements instead of semantic queries (like input labels or ARIA roles).

I'd recommend first add Cypress Testing Library to the project and update the existing tests to use semantic queries.

https://blog.sapegin.me/all/react-testing-4-cypress/

PR #1897, opened 2021-10-12
Remove `is-directory`, use `fs` module directly

visualfanatic:
This PR removes `is-directory` that is no longer maintained and replaces it with direct calls using `fs` module. It does exactly the same thing as `is-directory` with a bonus of having types built-in.


sapegin:
Thanks!

Issue #1932, opened 2022-01-27
Hot Reload cannot work


lyz810:
**Current behavior**
Hot Reload cannot work

**To reproduce**
Upgrade react-styleguidist from 11.1.8 to 11.2.0

11.1.8 works well

**Expected behavior**

juanca:
Which Webpack version are you using?

juanca:
Just for another data point. I tested with Webpack 5 and it works. It should work with Webpack 4.

lyz810:
> Which Webpack version are you using?
version of webpack: 5.67.0
I can't find any kind of websocket request in Chrome Network Panel
![image](https://user-images.githubusercontent.com/6788391/151477056-8a2c61d9-6ac8-496d-a5ae-189af4404b07.png)

I tried to revert part of new changes in 11.2.0(change the file in node_modules, and restart)
https://github.com/styleguidist/react-styleguidist/commit/92518df856bbd6ba0003b4f5ea7ed7565bdfea65#diff-5c873f8bf346d5bd54eaebda2215d65647f4f53a4b3e84987d520091d0d913f9L122
It didn't work, but something differerent in Chrome Network Panel
![image](https://user-images.githubusercontent.com/6788391/151477486-96ad7a6c-4339-49c7-b337-993c7a084c3d.png)

Then, I downgrade to 11.1.8，and restart the project, everything is ok
![image](https://user-images.githubusercontent.com/6788391/151477742-94e51c93-c567-46bc-a282-1329ab867dbb.png)
 

├─┬ babel-loader@8.2.3
│ └── webpack@5.67.0 deduped
├─┬ copy-webpack-plugin@10.2.1
│ └── webpack@5.67.0 deduped
├─┬ css-loader@6.5.1
│ └── webpack@5.67.0 deduped
├─┬ less-loader@10.2.0
│ └── webpack@5.67.0 deduped
├─┬ postcss-loader@6.2.1
│ └── webpack@5.67.0 deduped
├─┬ react-styleguidist@11.2.0
│ ├─┬ clean-webpack-plugin@3.0.0
│ │ └── webpack@5.67.0 deduped
│ ├─┬ copy-webpack-plugin@6.4.1
│ │ └── webpack@5.67.0 deduped
│ ├─┬ mini-html-webpack-plugin@3.1.3
│ │ └── webpack@5.67.0 deduped
│ ├─┬ react-dev-utils@12.0.0
[...]

PR #1996, opened 2022-04-19
WIP: Webpack 5 upgrade

ThomasRoest:
This is a start for upgrading to webpack 5 and webpack-dev-server ^4. With these changes, the basic examples are working  ( incl. hot module reloading ). But there are more todo's left

Todo

- [x] update dev-server config  
- [x] make StyleguidistOptionsPlugin.ts work with webpack 5 and fix test
- [x] fix make-webpack-config.spec.ts tests 
- [x] upgrade other webpack loader dependencies to latests versions (optional?)
- [x] fix type errors
- [x] fix deprecation warnings

This fixes some issues like #1932 
This requires a new major release


ThomasRoest:
@sapegin What is your vision concerning the future of this project? I'm willing to help out with some maintenance ( like this pull request ). Do you have time to review / accept these updates?

sapegin (approved):
Hey, thanks for the pull request! I'm happy to give you more rights here if you have time and the desire to work on the project. Unfortunately, I don't have either right now.

https://blog.sapegin.me/all/going-offline/

ThomasRoest:
> Hey, thanks for the pull request! I'm happy to give you more rights here if you have time and the desire to work on the project. Unfortunately, I don't have either right now.
> 
> https://blog.sapegin.me/all/going-offline/

sure!

sapegin:
I gave you write access to the repository ;-)

sapegin:
Thanks! I can have a look quickly, if it's ready, and then we merge it, and pray that semantic-release still works.

https://react-styleguidist.js.org/docs/maintenance

ThomasRoest:
@sapegin Did you have time to take a look? They only thing that is failing right now is codecov/patch. I'm not really familiar with codecov so not sure yet how to fix this..

src/client/rsg-components/Preview/Preview.tsx
```diff
@@ -88,7 +88,7 @@ export default class Preview extends Component<PreviewProps, PreviewState> {
 			try {
 				ReactDOM.render(wrappedComponent, this.mountNode);
 			} catch (err) {
-				this.handleError(err);
+				this.handleError(err as Error);
```
sapegin, line 91: This would be better than silencing type checker:

```suggestion
        if (err instanceof Error) {
				  this.handleError(err as Error);
				}
```

Same in other places.

src/loaders/props-loader.ts
```diff
@@ -47,7 +47,7 @@ export default function(this: Rsg.StyleguidistLoaderContext, source: string) {
 			}
 			docs = docs[0];
 		}
-	} catch (err) {
+	} catch (err: any) {
```
sapegin, line 50: Same here, we should remove all `any`s.

src/scripts/__tests__/make-webpack-config.spec.ts
```diff
@@ -15,15 +12,17 @@ const styleguideConfig = ({
...
+
 const getClassNames = (plugins: Tapable.Plugin[] = []): string[] =>
 	plugins.map((x) => x.constructor.name);
 
 const process$env$nodeEnv = process.env.NODE_ENV;
 
 beforeEach(() => {
-	(CopyWebpackPlugin as jest.Mock).mockClear();
+	const mockedPlugin = CopyWebpackPlugin as unknown;
+	(mockedPlugin as jest.Mock).mockClear();
```
sapegin, line 25: `as unknown as something` is a common pattern (though the one that should be avoided), and better not to split it into multiple lines.

```suggestion
	((CopyWebpackPlugin as unknown) as jest.Mock).mockClear();
```

src/scripts/__tests__/make-webpack-config.spec.ts
```diff
@@ -34,9 +33,7 @@ it('should return a development config', () => {
 	const env = 'development';
 	const config = makeWebpackConfig(styleguideConfig, env);
 
-	const errors = validate(config);
-	expect(errors).toHaveLength(0);
-
+	validate(config);
```
sapegin, line 36: How do we know there are no errors now?

src/scripts/__tests__/make-webpack-config.spec.ts
```diff
@@ -15,15 +12,17 @@ const styleguideConfig = ({
 	title: 'Style Guide',
 } as unknown) as Rsg.SanitizedStyleguidistConfig;
 
-const getClasses = (plugins: Tapable.Plugin[] = [], name: string): Tapable.Plugin[] =>
+const getClasses = (plugins: any[] = [], name: string): Tapable.Plugin[] =>
```
sapegin, line 15: Could we get rid of `any` here?

src/scripts/utils/StyleguidistOptionsPlugin.ts
```diff
@@ -24,21 +27,10 @@ export default class StyleguidistOptionsPlugin implements WebpackPluginInstance
 
 	/**
 	 *
-	 * @param compil webpack 4 `compilation.Compilation`, webpack 5 `Compilation`
+	 * @param compil Compilation
```
sapegin, line 30: We probably don't need this comment anymore, it repeats the type we already have below.

src/scripts/utils/__tests__/StyleguidistOptionsPlugin2.spec.ts
```diff
@@ -0,0 +1,39 @@
+import StyleguidistOptionsPlugin from '../StyleguidistOptionsPlugin';
```
sapegin, line 1: This is confusing: why do we have two tests for one module?

src/scripts/utils/StyleguidistOptionsPlugin.ts
ThomasRoest: this was removed in newer commits

src/scripts/__tests__/make-webpack-config.spec.ts
ThomasRoest: fixed

src/loaders/props-loader.ts
ThomasRoest: fixed

src/client/rsg-components/Preview/Preview.tsx
ThomasRoest: fixed

src/scripts/__tests__/make-webpack-config.spec.ts
ThomasRoest: the validate function still throws errors if the config is invalid, just doesn't return anything.

sapegin: I believe it does but the test isn't testing this. Why not [use `toThrow()` here](https://github.com/sapegin/jest-cheat-sheet/blob/master/Readme.md#exceptions)?

src/scripts/utils/__tests__/StyleguidistOptionsPlugin2.spec.ts
ThomasRoest: yes had same trouble mocking webpack in these tests, I'm not sure how to get this working with both mocks / tests in the same file

sapegin: As a workaround, `resource` could be a global variable that is modified inside each test.

src/scripts/__tests__/make-webpack-config.spec.ts
ThomasRoest: fixed

src/scripts/utils/__tests__/StyleguidistOptionsPlugin2.spec.ts
ThomasRoest: I don't think that works because the mock is set once, it doesn't matter if you modify the resource variable, the mocked value will stay the same. I've tried multiple ways to clear, or reset the webpack mock between tests, but separate files seems to be the easiest.

ThomasRoest: @sapegin nevermind, I fixed this as well. So what do you think? Ready to merge?

sapegin:
There are still a few unresolved comments, `any`s and `as`es, otherwise it looks good.

src/scripts/__tests__/make-webpack-config.spec.ts
ThomasRoest: I don' think I understand, It's still testing if the config make in makeWebpackConfig is valid, it will throw a readable error if not

```
ValidationError: Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
     - configuration.output has an unknown property 'testInvalid'. These properties are valid:
 ```

Not sure how using toThrow() will make it better.

ThomasRoest: added
https://github.com/styleguidist/react-styleguidist/pull/1996/commits/20497265e9aae42d815f6ce56a1d022f47d98ba5

ThomasRoest:
@sapegin should be good now right?

sapegin:
There are still open comments and _a lot_ of new `any`s and `as`es.

ThomasRoest:
sorry but I don't know which comments you are talking about, the ones from your review are resolved. 
Concerning the new any's and as assertions, again which ones?

sapegin:
I see at least one suggestion above, for the any and as — there should be none of them added to the codebase. 

This kind of code reviews take too much time from me, I'm tired of repeating everything 20 times, and I'm going to ignore any activity in this project in the future. Open source became too damaging to my mental health.

ThomasRoest:
> I see at least one suggestion above, for the any and as 
I fixed those suggestions. 

> there should be none of them added to the codebase.
This is not realistic, there are always situations in which you need an escape hatch, I can't replace all the existing any's and `as` assertions that where already in the codebase. Also I think all these outdated dependencies/lack of maintenance are a bigger problem for this project than a few any or 'as' assertions.

> This kind of code reviews take too much time from me, I'm tired of repeating everything 20 times, and I'm going to ignore any activity in this project in the future. Open source became too damaging to my mental health.

I understand! But we've already put in alot of work for this pull request, so I would really like to complete this. Thank you!

ThomasRoest:
@sapegin i fixed the remaining new 'any's that I've missed.

Issue #2033, opened 2022-09-16
Webpack "devServer.*" options unable to be customized


dannobytes:
**Current behavior**

After upgrading to v12, webpack@5 now by default enables the `devServer.client.overlay` option, which shows a blocking modal whenever a compilation warning or error occurs. This is very obstructive, and we should be able to easily modify this by overriding the option. In addition, `progress` is disabled, and we should be able to turn this on.

```
// styleguide.config.js
{
  ...,
  webpackConfig: {
    devServer: {
      // Turn overlays off. Enable progress indicator.
      client: { overlay: false, progress: true },
    },
  }
}
```

However, these configs are not making their way to the webpack dev server. I tracked it down to the sequence of how the internal webpack dev config was being shallowly merged:

https://github.com/styleguidist/react-styleguidist/blob/a460fcc3d6a674146cd74b53362058e17e9aea7e/src/scripts/create-server.ts#L36-L39

Because the `baseConfig` contains a `client` field, it's always overwriting what's being passed through from `...webpackConfig.devServer`.

**Solution**

A simple fix would be reorder this to ensure the baseConfig is applied first, while merging onto it the `webpackConfig.devServer`, e.g.

```js
	const webpackDevServerConfig: Configuration = {
		...baseConfig,
		...webpackConfig.devServer,
	};
```

This would ensure our options from the styleguide.config file were being passed through.

**Expected behavior**

When overriding `webpack.devServer.client` [options](https://webpack.js.org/configuration/dev-server/#devserverclient), they should be passed through so I can enable progress and/or turn off overlays.

ThomasRoest:
can you add a commit that only switches these lines, and leaves out other changes? https://github.com/styleguidist/react-styleguidist/blob/master/src/scripts/create-server.ts#L38

PR #2034, opened 2022-09-16
fix: ensure devserver client options are configurable

dannobytes:
When a custom `webpackConfig.devServer.client` option is set, they are
currently being overridden by the `baseConfig.client` field.

Ensure that field is extendible by consumers while also ensuring the
base config values stay fixed.

https://github.com/styleguidist/react-styleguidist/issues/2033

### DIFF
--- a/src/scripts/create-server.ts
+++ b/src/scripts/create-server.ts
@@ -36,6 +36,10 @@ export default function createServer(
 	const webpackDevServerConfig: Configuration = {
 		...webpackConfig.devServer,
 		...baseConfig,
+		client: {
+			...webpackConfig.devServer.client,
+			...baseConfig.client,
+		},
 	};
 
 	const compiler = webpack(webpackConfig);


dannobytes:
@sapegin FYI, this fix is ready for review whenever you have a moment

src/scripts/create-server.ts
```diff
@@ -33,17 +33,14 @@ export default function createServer(
...
-	const webpackDevServerConfig: Configuration = {
-		...webpackConfig.devServer,
+	// Allow custom devServer options to override base config.
+	webpackConfig.devServer = {
 		...baseConfig,
-		client: {
-			...webpackConfig.devServer.client,
-			...baseConfig.client,
-		},
+		...webpackConfig.devServer,
```
ThomasRoest, line 39: can you maybe only switch these lines? And leave out the other changes here?

dannobytes: Technically, yes. Swapping these two lines is all that's required to fix the issue. However, in order to properly test these changes, we need the `webpackConfig` object passed to the compiler on L42 to have the same dev server options that are passed into `WebpackDevServer` on L43. Without this, we have no way to reference dev server options in our tests.

This should be an improvement as it ensures the same config options are passed to both `WebpackDevServer` and `webpack`. It clarifies the source of truth, which is otherwise confusing.

PR #2038, opened 2022-10-18
Migrated enzyme tests to react-testing-render or @testing-library/react.

FritsvanCampen:
I've tried to keep the impact on the tests as small as possible.
Shallow snapshot tests use react-test-renderer/shallow.createRenderer.
Deep snapshot tests use react-test-renderer.create.
Interaction tests use @testing-library/react.render.

In snapshots 'class' changed to 'className'.
The 'key' property is no longer emitted.
Injected HTML is formatted differently (quoted and no newlines).

Some tests navigated into the render output to pick out a smaller piece to snapshot. That was too difficult to replicate so some separation has been lost.

There is a problem with the deabsdeep/serializer. It breaks serialization of React Elements, because the map 'loses' the $$typeof marker used by jest's native React serializer.
I've copied the serializer code into this project and fixed the problem. In the future we could look at using this serializer selectively for only tests that need it.
deabsdeep library is still used in places, just not the serializer.

NB. The docs mentioned enzyme. I've updated it to something sensible.

PS. I had some problems with the path tests not working on Windows; because Windows uses a different path separator and the tests don't sanitize. So I might have missed some things 🤞

### DIFF
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -101,7 +101,7 @@ Because of isolation and theming you need to explicitly declare `fontFamily`, `f
 
 ## Testing
 
-We’re using [Jest with Enzyme](http://blog.sapegin.me/all/react-jest) for testing. Put your component tests into `Component.spec.js` file in the same folder and all other tests into `__tests__/filename.spec.js`.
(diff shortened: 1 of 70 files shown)


ThomasRoest:
@sapegin Any opinion on this? This migration is required for a future upgrade to React 18.

https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl

PR #2087, opened 2022-12-13
Upgrade to React 18

ThomasRoest:
Upgrading to React 18
- replacing deprecated api's incl. ReactDOM.render()


premsai2030:
May i know what is the node js version you are using ?

PR #2101, opened 2023-01-07
replace ReactDom.render with createRoot

joewab:
I don't think this is everything that needs to be done to update for React 18 but it's a start and the tests are passing.


ThomasRoest:
https://github.com/styleguidist/react-styleguidist/releases

PR #2125, opened 2023-02-13
fix: update peer deps

ThomasRoest:


alexsurelee:
As far as I can tell, this should have been a breaking change? Dependency tree now fails for any projects below React 18, where this was previously accepted.

PR #2148, opened 2023-07-25
Ensuring example re-renders always work

stevematney:
Because the `mountNode` never changes in the previous version of the code, `this.callbackRef` would not get called after an error was thrown. After a second re-render with new code, the Example would simply be blank on the screen (not rendering either an Example or an error message).

https://github.com/styleguidist/react-styleguidist/blob/0802ffbe716fef22086f87f5c6a9caec2f9126d1/src/client/rsg-components/Preview/Preview.tsx#L131-L136

This code ensures that we always get an Example re-render when new code is present.


stevematney:
Resolved by #2158

